PR #7, opened 2023-08-02
Update minimum temperature logic to defer to slicer during printing

kieraneglin:
**Issue:**

There are two main things this PR addresses: 
1. When printing with two different-temperature filaments (eg: PLA and PETG for easy support removal), HH waits until the higher of the two temperatures is reached before forming the tip or unloading the filament. This could result in a bad tip forming and wastes time during a swap
2. If your `min_temp_extruder` is set to something higher than your printing temp, `min_temp_extruder` will be set as your new printing temp during the swap. You could just set `min_temp_extruder` lower, but I have mine set at 230 to work for PLA, PETG, TPU, ABS, and nylon since this allows me to remove filament without knowing/caring what kind it is. But during a print we should defer to the slicer's temps instead

**Fix Description:**

See code comments, but at a high level:

1. Updates `_set_above_min_temp` to defer to the slicer when printing and renames to better reflect new behaviour
    - NOTE: if a target temp is explicitly specified, we do still respect that. This is useful for when you pause, set the temp manually, and then resume.
    - NOTE: this change does place more trust in the slicer, but I think that's reasonable. Just in case, we do still use Klipper's `can_extrude` as a safeguard
2. Updates `_load_extruder` to ensure the target temp has been reached before attempting a load. This should be unneeded in practice, but better safe than sorry!

**GitHub Issues Resolved:**

None


extras/mmu.py
```diff
@@ -2006,26 +2006,52 @@ def _get_print_status(self):
...
+        #   1. When printing with different temperature filaments, this prevents waiting for the higher
+        #      temperature before unloading the lower-temperature filament
+        #   2. Prevents setting the wrong temp if your `min_temp_extruder` is higher than the temp of the filament
+        #
+        # This does mean that we're trusting the slicer to have set the correct temperature, but that's a reasonable
+        # assumption and we do maintain the safeguard of Klipper's `can_extrude` flag
+        if (self._is_in_print() or self._is_in_pause()) and extruder_heater.can_extrude:
+            if target_temp_override == -1:
+                self._log_info("Deferring to slicer for extruder temperature (%.1f)" % extruder_heater.target_temp)
+            else:
```
kieraneglin, line 2034: I've just realized this won't catch the case where a heater turns off during a pause. Even if `target_temp_override` is set, it'll never get to that point because `can_extrude` is false. Will update

extras/mmu.py
```diff
@@ -2006,26 +2006,52 @@ def _get_print_status(self):
...
+        target_temp = self.printer.lookup_object(self.extruder_name).heater.target_temp
+        current_temp = self.printer.lookup_object(self.extruder_name).get_status(0)['temperature']
+
+        if abs(target_temp - current_temp) < 1:
+            self._log_debug("Waiting for extruder to reach target temperature (%.1f)" % target_temp)
+            self.gcode.run_script_from_command("TEMPERATURE_WAIT SENSOR=extruder MINIMUM=%.1f MAXIMUM=%.1f" % (target_temp - 1, target_temp + 1))
+
     # Ensure we are above desired or min temperature and that target is set
-    def _set_above_min_temp(self, target_temp=-1):
+    def _set_above_min_temp(self, target_temp_override=-1):
```
kieraneglin, line 2018: I think I'm going to rename this to better capture the new logic. It's no longer about the min temp, it's about waiting for a safe temp

moggieuk:
Thank you!  This seems to make sense and looks correct but I'm going to need to wait until I'm reunited with my printer in about 10 days to test / experiment...

kieraneglin:
@moggieuk updating to say that I've had time to test this with _many_ prints and it's working as I'd expect! Loads from PLA (210º) to PETG (255º) and back are working well with no downtime to wait for heating

moggieuk:
This looks good.  Can you help me with two things so I can merge easily:
1. Resolve the conflict -- I added a restoration of sync state a little differently than other of your suggestions.
2. Did you solve the "a heater turns off during a pause" case?

kieraneglin:
@moggieuk 

1. Resolved!
2. I have - `def _pause()` stores `self.paused_extruder_temp` which is passed to `_ensure_safe_extruder_temperature` in the `MMU_RESUME` macro. Within `_ensure_safe_extruder_temperature`, there's logic that will run if an override is set and the printer is currently printing or paused. That logic will respect the temp override and wait for the target temperature before continuing

bakatrouble:
https://github.com/moggieuk/Happy-Hare/issues/13
Seems to be related

PR #10, opened 2023-08-07
Update MMU_SYNC_GEAR_MOTOR to allow specifying print state

kieraneglin:
**Issue:**

When calling `MMU_SYNC_GEAR_MOTOR` during a print, the gear current is set to 100%

**Fix Description:**

`cmd_MMU_SYNC_GEAR_MOTOR` and accompanying docs have been updated to allow manually specifying whether it should treat `_sync_gear_to_extruder` like you're in a print. The default is automatically determined based on current print state

**GitHub Issues Resolved:**

None


moggieuk:
I don't have any problem with this PR but want to understand the use case... would it be:
you have gear sync current < 100% but in mid print you want to restart with it set to some other %.

If this is the case, then wouldn't:
`MMU_TEST_CONFIG sync_gear_current=xxx` be an easier way to specify?

The difference would be that using 'MMU_SYNC_GEAR_MOTOR' would be a one off change, whilst using `MMU_TEST_CONFIG` would present until explicitly changed or next restart.

kieraneglin:
@moggieuk the case I was trying to solve was when I didn't have sync enabled for a single-colour print (that used the gate rather than the bypass), but due to some feeding issues I decided to turn it on mid-print. I did so by calling `MMU_SYNC_GEAR_MOTOR` which worked, but in doing so the sync current was set to 100% instead of the 60% it should be during in-print syncing.

This PR is meant to resolve that, but also to give people the option in case they don't like that behaviour

moggieuk:
Got it. That makes sense.  I'll go ahead and merge this. Thanks!

PR #11, opened 2023-08-09
[WIP] Adds new file processing module for automatically determining tool usage

kieraneglin:
**Issue:**

As outlined in the [README](https://github.com/moggieuk/Happy-Hare#11-useful-pre-print-functionality), there is a use case for knowing which tools are used in a print ahead of time. PRs have been created for various slicers, but who knows when/if they'll be merged.

**Fix Description:**

A new module, `MmuServer` has been created for Moonraker. It uses a regular expression to search for `T<tool>` and `MMU_CHANGE_TOOL(_STANDALONE)? TOOL=<tool>` to enumerate which tools are used for a print. These stats are then stored alongside other metadata at the end of the gcode to speed up future reads. NOTE: I've been asking around the Moonraker/Klipper discord about any way to incorporate this into Moonraker's metadata scan/database and it doesn't look like that's possible at this moment. In fact, they used to have an event that fired on metadata scans which has now been removed. I'll keep looking into this because more generally storing freeform metadata would kick ass, but for now that's a dead end. The module is named `MmuServer` because that's a generic enough name to work for other moonraker addons, but for now it's just the file preprocessor

Currently, the reading/writing is done on file upload. There is a PR open in Moonraker for the metadata scan event (as discussed above), but that doesn't appear to be going anywhere in a hurry.

There are some potential limitations if the user utilizes some custom macro for tool changes, but I consider this iteration to be more about viability than perfection. If there is a valid usecase here, I'm sure we can also start looking for a magic comment and put that in the docs.

**Usage:**

In your slicer, update your `PRINT_START` (or similar) macro to use the magic string `!mmu_inject_referenced_tools!` like so: `PRINT_START MMU_REFERENCED_TOOLS=!mmu_inject_referenced_tools!`. Keep in mind that the only special thing here is the magic string - the `MMU_REFERENCED_TOOLS=` can actually be whatever you want. 

- NOTE: I used the string `!mmu_inject_referenced_tools!` because I _think_ all slicers are cool with that. I wanted to use square or curly braces instead of `!`, but Prusa Slicer (and its derivatives) interpret that to mean it's a variable that needs to be replaced. Those slicers would need to instead do something like `PRINT_START MMU_REFERENCED_TOOLS={'{mmu_inject_referenced_tools}'}`to get the string `{mmu_inject_referenced_tools}` to appear in the gcode which I thought was too much of a pain
- NOTE: I went with the magic comment approach to maximize flexibility in what the user can do since it's ultimately their choice how to use it - we aren't doing anything magic. For example, I'd use it in such a way that it does its checks while it's waiting for the bed to heat.

Update your `PRINT_START` klipper macro to use this new argument like `MMU_CHECK_GATES TOOLS={params.MMU_REFERENCED_TOOLS}` or whatever is appropriate for your setup

In `moonraker.conf`:

```
[mmu_server]
enable_file_preprocessor: True
```

**Before initial merge:**
- [x] Code review
- [ ] Extended testing with real prints - both single and multicolour, with and without tool changes, etc. There almost certainly will be errors in testing, but I want to know where and why
- [x] Validate that `MMU_CHECK_GATES` works as expected. For example, should it unload the current filament? Should it home?
- [x] Update installer to symlink file to the `~/moonraker/moonraker/components` directory
- [x] Update docs

**Before feature completion:**
- [ ] Hook up file preprocessing on upload

**Future goals:**
- [ ] Set up callback to run on file metadata scan

**GitHub Issues Resolved:**

None


extras/mmu.py
```diff
@@ -4419,10 +4425,15 @@ def cmd_MMU_CHECK_GATES(self, gcmd):
             gates_tools = []
             if tools != "!":
                 # Tools used in print (may be empty list)
+                if tools == 'AUTOMATIC':
```
kieraneglin, line 4428: I want to be really clear that I'm not dead-set on anything here. I wanted to get it running enough to test, but I have no strong feelings about how this macro should interface with `_get_toolchanges_from_file`

tests/mmu_file_processor_test.py
```diff
@@ -0,0 +1,62 @@
+import unittest
+import os
+import shutil
+
+from extras.mmu_file_processor import MmuFileProcessor
+
+class TestMmuFileProcessor(unittest.TestCase):
```
kieraneglin, line 7: I added some unit tests because testing on various real-world files was getting tedious

moggieuk: I really like this because I will use it!!   I haven't had a chance to try yet but hopefully will tomorrow.
To be clear this will only work if the entire file is immediately available so it should probably be noted that the few (only encountered 2) folks that use Octoprint and stream the gcode will be out of luck.  We just need to make sure that logic protects against this case and does nothing...

moggieuk:
You can probably see I'm pushing for your file metadata update event!!

moggieuk:
Just started playing... Couple of quick comments before diving in:
1) Should should add your (c) to the files!  Maybe use the same format as I'm using elsewhere but with credit to you.
2) Rexexp probably should also look for `MMU_CHANGE_TOOL_STANDALONE` since I still have this alias define and many are using it.
3) Could we change "!mmu_inject_tools_used!" to "!mmu_inject_referenced_tools!"  This is a nit by ties it nicely with my prusa slicer PR if that is ever picked up.
4) Could you create a test running script in `tests` directory so very easy to run.

More substantial feedback coming :-)

kieraneglin:
@moggieuk currently, the script just removes the placeholder if there are no toolchanges which could result in something like this: `PRINT_START MMU_REFERENCED_TOOLS= OTHER_ARG=other_val`. I haven't tested if this causes issues in the real world, but I'm wondering if it should add in some non-whitespace character if no toolchanges are found. My gut says `""` would be best, but maybe I could insert `!` to mirror what `MMU_CHECK_GATES` uses as a fallback for the `TOOLS` arg. Any thoughts?

moggieuk:
I believe that having empty param to PRINT_START is fine.   MMU_CHECK_GATES can handle an empty TOOLS= (at least it did when I first wrote it)

If MMU_CHECK_GATES does eject then it should probably record what was loaded and put it back afterwards.  A single color print has no tool changes and if a user (like me) doesn't bother with the initial MMU_CHAGE_TOOL STANDALONE=1 TOOL=0 but rather relies on the loaded tool...

moggieuk:
Ok. I've pulled this into my local development branch and I'm creating the install support for it... I want to run a couple of tests tomorrow after getting it all set up in my workflow, but it all looks good to me...!

moggieuk:
Ok. Pushed into my "development" branch and added installer support, doc, etc.  The current version uses "!referenced_tools!" as the placeholder because it is as close to "{referenced_tools}" (the Prusa PR) as we can get.  The "development" branch has updates to MMU_CHECK_GATES to unload/reload existing filaments.

I've run some big models though it without issue.

PR #21, opened 2023-09-01
Allow tracking of extrusion multiplier per tool

jorender:
**Issue:**
When running a multi-color print I had to adjust the extrusion multiplier mid-print for a single filament.  The remainder of the print needed to either be watched to reset the value for each tool change or allow all tools to use the modified value.

**Fix Description:**
This request adds the ability for HappyHare to track changes to the extrusion multipliers per tool and apply them on tool change.  This optional feature is enabled by setting the `enable_tool_extrusion_multiplier` value in `mmu_parameters.cfg`.   The default setting is disabled.  Once set after a tool is unloaded the current extrusion multiplier is retrieved from the `gcode_move` module and stored in a list.   On tool load the last value is retrieved from the list and applied by sending an `M221` command.   This list is reset to 100% for all tools on either a `FIRMWARE_RESTART` or Klipper restart.

A new entry has also been added to the API `get_status` method
Configuration example has been added to `mmu_parameters.cfg` in the optional section.

**Testing Done**
This has been tested with a live print as well as the manual steps listed below.
1. Home printer and MMU
2. Load a tool and set the extrusion multiplier
3. Load a second tool and verify
  3a. The extrusion multiplier reset to 100%
  3b. `get_status` returned the value set in step 2 for the initial tool
4. Load the tool loaded in step 2 and verify:
  4a. The extrusion multiplier was set to the value set in step 2
  4b. `get_status` returned the value set in step 2

**GitHub Issues Resolved**
None


moggieuk:
Cool.  Let me look at this tomorrow with a clear head but it sounds reasonable..  I guess the key here is that it is a dynamic change during a print or sequence of prints rather than persisted.  I ask because this could be added as an attribute of the filament much like material and color (perhaps temp is another important setting or even pressure advance).

My prior thinking has been that "material" and "color" would be enough to add logic in the toolchange macro to tweak settings.

jorender:
Correct being dynamic is the key here with the intent that once you land on a good multiplier mid-print then the slicer should be updated accordingly and Klipper reset to 100% for that tool.  The Klipper update would have to be a manual operation.  Perhaps a command could be added to manage the individual multipliers for each tool; something along the lines of `MMU_MATERIAL TOOL=0 EXTRUSION_MULTIPLIER=100`, with a matching display command.  Since color and material type would be constants per filament we could easily add those as well.  Note that I'm classifying extrusion multiplier as a constant as it is used by the slicer to generate E values and changing it mid-print doesn't actually change any gcode.

For the variable items (pressure advance/temperature) I'd have to put some thought into how that can be made to work since the slicer can be configured to set these after a tool change in the Filament Custom G-code section.  My concern is that if you change either one mid-print and have custom g-code then we would dynamically apply the mid-print value, but that would then be overwritten with the default values from the slicer. (Attaching screenshots for reference)

Filament Custom G-code
Generated G-code

moggieuk:
Ok, that makes sense.  I do think this should be added to the "list" of filament/tools loaded just like color and material.  It would then be cleared (set to 100%) if either of those are modified in addition to explicitly being set.  Currently I'm using `MMU_GATE_MAP` command to manipulate everything that is stored per gate (or tool if mapping is 1:1) and I'd prefer not to increase the command set if I can help it.  A user created macro can easily wrap into something easier to remember...

However... you may be right that there needs to be an easy way to reset this, obviously klipper/firmware restart is one so long as value is not persisted in mmu_vars.cfg, but do you think ideally this should be reset after a single print?  (also on cancel?)

I ask, because as you said, once you have this value nailed you would update your slicer and thus could have gcode files that use different extrusion multipliers.

Side issue: detecting print end is actually more problematic than it sounds!

moggieuk:
One last though (I am thinking hard about this :-). To be complete, both speed override (M220), extruder override (M221) should be treated in the same way?

moggieuk:
After noodling on this for a while, how about:
- Feature always enabled
- Records M220 and M221 overrides at start of tool change against the current TOOL (not gate)
- Restores M220 and M221 when tool is next selected
- Stores in parallel array not conceptually connected with "gate_map" (but is num_gates wide)
- New command `MMU_TOOL_OVERRIDES TOOL=x M220=% M221=% CLEAR=1` combinations allow for setting (nice for testing), selectively clearing, clearing all, displaying current settings
- Not persisted so cleared on klipper restart
? Ideally it could be cleared on print end but that it not easy to automate but command could be put in print_end macro by user..

jorender:
I like this path.  I'll get to work on the modifications you suggested after I've gotten some sleep, shouldn't take too much.

moggieuk:
Thanks. That would be helpful - I started to do it but always easier just to review ;-)

moggieuk:
Just checking in.  How is the PR going?

jorender:
Got the code change done/tested, still messing with formatting of the display message that I hope to have completed this today.

moggieuk:
Thanks for you help with this!  I've merged the PR.  I hope I didn't offend but I changed the code in three ways:
1. Reduced the quality by combining speed / extrude
2. Work only in % and the user level and store as a float
3. Reduced the volume of logging (I'm a little sensitive to volume when try to debug)

I'm going make an edit to the doc but will mention in next "Happy Hare update message"

jorender:
Yes, the white space stripping was very annoying :)

No offense on the changes at all, these requests are a great way to learn more about our craft while making sure things fit into the overall project seamlessly.

moggieuk:
Just added support for other possible cases (not just change_tool).  Adding doc update -- did change "CLEAR=1" to "RESET=1" to match other commands.

PR #24, opened 2023-09-02
Fix doc links

EiNSTeiN-:
I noticed this link was broken when viewing the readme from https://github.com/moggieuk/Happy-Hare, this PR fixes that for all README links: 

image


moggieuk:
Thanks! The previous links worked on my system but you are right they aren't right.

PR #25, opened 2023-09-02
fix: fix Moonraker update section

meteyou:
Add primary_branch to the config section.


moggieuk:
Ah, thank you!!  Wasn't aware of that parameters

Issue #33, opened 2023-09-06
date format for mmu.log


Benoitone:
https://github.com/moggieuk/Happy-Hare/blob/6416e02201701c6b8e63a18a32d9f046ce2e16a5/extras/mmu.py#L480C13-L480C112
can be change to;
`self.queue_listener.setFormatter(MultiLineFormatter('%(asctime)s %(message)s', datefmt='%H:%M:%S'))`
for correct 24h format:
![HH](https://github.com/moggieuk/Happy-Hare/assets/63300355/641cd53b-fc83-41cd-a5a6-4a4b6d0d8813)

moggieuk:
Fixed in merge of PR #34

PR #34, opened 2023-09-06
Fix time format for mmu.lof

Benoitone:
fix for issue https://github.com/moggieuk/Happy-Hare/issues/33#issue-1884725929


moggieuk (approved):
Thanks. No idea how it lasted in 12h format for so long!

Issue #36, opened 2023-09-09
Resuming after Pause during print changes temperature to MMU extruder default temp


matthewj301:
Running ERCF 1.11sb.

 When my printer resumes after either a manual PAUSE, or resuming after the printer is paused in order to initiate a tool change, ERCF sets my extruder temp to its self.default_extruder_temp (200).

My Klipper extruder min temp is set to 185. 

These are the logs showing the behavior.

extruder: target=265 temp=265.0 pwm=0.396
Stats 1468100.8: gcodein=0  mcu: mcu_awake=0.002 mcu_task_avg=0.000006 mcu_task_stddev=0.000004 bytes_write=240652 bytes_read=522690 bytes_retransmit=9 bytes_invalid=0 send_seq=13771 receive_seq=13771 retransmit_seq=2 srtt=0.000 rttvar=0.000 rto=0.025 ready_bytes=0 upcoming_bytes=0 freq=180000110 mmu: mcu_awake=0.002 mcu_task_avg=0.000007 mcu_task_stddev=0.000008 bytes_write=33021 bytes_read=471028 bytes_retransmit=9 bytes_invalid=0 send_seq=3493 receive_seq=3493 retransmit_seq=2 srtt=0.000 rttvar=0.000 rto=0.025 ready_bytes=0 upcoming_bytes=0 freq=12000158 adj=12000151 toolboard: mcu_awake=0.004 mcu_task_avg=0.000016 mcu_task_stddev=0.000014 bytes_write=76187 bytes_read=415414 bytes_retransmit=9 bytes_invalid=0 send_seq=7624 receive_seq=7624 retransmit_seq=2 srtt=0.000 rttvar=0.000 rto=0.025 ready_bytes=0 upcoming_bytes=0 freq=63999239 adj=63999241 Octopus: temp=26.8 raspberry_pi: temp=38.0 sd_pos=4166 heater_bed: target=108 temp=108.2 pwm=0.294 toolboard: temp=41.3 sysload=0.40 cputime=4967.628 memavail=3355488 print_time=2940.391 buffer_time=0.853 print_stall=0 extruder: target=265 temp=264.9 pwm=0.383
- - TRACE: Synchronously loading filament to nozzle. Stepper: 'gear+extruder' moved 85.0mm, encoder measured 83.6mm (delta 1.4mm)
- DEBUG: Total measured movement: 83.6mm, total delta: 1.4mm
[...]

PR #38, opened 2023-09-09
store the extruder temperature in _handle_runout

garyd9:
Should resolve Issue #37 .  Please see the comments in that issue for a long winding report, triage, and thought process.

The short version is that _handle_runout() should set self.paused_extruder_temp so that when the runout code eventually calls cmd_MMU_RESUME(), the extruder temperature isn't changed to self.default_extruder_temp (which might be safe for tips, but not for printing.)

PR #39, opened 2023-09-09
check that target_temp_override is >=  klipper_minimum_temp

matthewj301:
Fix for issue #36

target_temp_override is passed in as 0 from self.paused_extruder_temp (separate issue), evaluating to the first if statement and causing new_target_temp to be assigned 0.

This ultimately leads to the new extruder temp being set to the MMU's default temp. I don't see a reason in any scenario where checking if target_temp_override >= klipper_minimum_temp could cause issues, as we shouldn't be printing below that temp anyway.

Tested in my setup (RPi4, RatOS, RatRig 3.1 300mm) and fixed my issue.

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -2029,8 +2029,13 @@ class Mmu:
         current_target_temp = extruder.heater.target_temp
         klipper_minimum_temp = extruder.get_heater().min_extrude_temp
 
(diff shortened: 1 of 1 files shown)


matthewj301:
Related to https://github.com/moggieuk/Happy-Hare/pull/38

PR #41, opened 2023-09-09
Confirm self.paused_extruder_temp is set on cmd_MMU_RESUME

matthewj301:
Fix for issue #36 - When a manual pause is issued, we don't currently intercept the macro, and so do not set self.paused_extruder_temp - this causes the target_temp to be set to the default extruder temp when RESUME is called and intercepted by Happy Hare.

This ensures that no matter how the PAUSE macro is called, self.paused_extruder_temp will be set.

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -3874,7 +3874,9 @@ class Mmu:
             else:
                 self._log_always("State does not indicate flament is LOADED.  Please run `MMU_RECOVER LOADED=1` first")
                 return
-
+        # In case a manual pause was called where we do not intercept and set the extruder temp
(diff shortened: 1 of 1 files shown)


matthewj301:
Tested on RatOS, with a RatRig V-Core 3.1 300mm. 

Tested by issuing multiple manual pause and resumes.

moggieuk (approved):

moggieuk:
Thanks. This may work a little differently in an (larger) update that I'm planning because I do now wrap a regular "PAUSE" and "CANCEL_PAUSE" just for this little corner cases!

PR #42, opened 2023-09-11
allow custom value of park_pos in tip_forming macro

thisiscam:
Allows tip forming macro to set a custom value of park_pos. This is useful for tip forming routines like filament cutters, where the park pos should be managed by the cut macro.

### DIFF
--- a/config/base/mmu_software.cfg
+++ b/config/base/mmu_software.cfg
@@ -104,6 +104,13 @@ variable_final_eject: 0                # Default 0, enable during standalone tun
 # Park filament ready to eject
 variable_parking_distance: 0           # Final filament parking position after final cooling move, 0 will leave filament where it naturally ends up
 
+# The parking position 
+variable_output_park_pos: -1       # The parking position of the filament relative to the nozzle. This is an output variable decided by this macro. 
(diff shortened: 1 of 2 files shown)


thisiscam:
Some weird things is happening in the install script when you add a new `variable_` to `mmu_software.cfg`. Basically, a new field with empty value is copied to the new config file if the field does not exist in the old config.

moggieuk:
Awesome, thanks!   I'll check the installer but the intended behavior is to pull any old variable values from the old file (perform and name changes caused by updates -- none yet), copy the new mmu_software and update any variables with previous value..

moggieuk:
Does this actually work?  Any value set in the macro itself is not passed back to python for me...?

thisiscam:
> Does this actually work? Any value set in the macro itself is not passed back to python for me...?

Yeah, it works for me. What have you tried?

thisiscam:
> Does this actually work? Any value set in the macro itself is not passed back to python for me...?

Hmm, actually, I've only tried setting static values. I think what happened in your case is that I am missing a `toolhead.wait_moves`

moggieuk:
No worries. Fixed and merged.

PR #75, opened 2023-10-27
Retract gear while selector move

thisiscam:
This code tries to move the gear backwards while the selector is traveling.

The goal is to reduce the chance of filament protruding out of the gate for ERCF 1.1. 
To elaborate, to my observation, the only times with my ERCF 1.1 when filament protrudes is when the unloaded filament has pressure coming from the buffer. When the selector travels, it brings the gate washers up and so the pressurized filament just takes its chance coming out of the gate.
 
The hope here is that by moving the gear backwards (retracting), we use the friction to combat that pressure, so to get the washer a good enough time window til it drops back.  

Currently this adds a hidden parameter `retract_gear_speed_while_moving_selector`, which when set to a positive value, will retract gear by that speed while the selector travels (including homing).

The code has a bit of hacky flavor to it: I manually calculated the speed by which to move both the gear and selector so that klipper will do the vector decomposition to recover the individual movement speeds of the gear and the selector. 
That is, the speed of homing now becomes: 
```
sqrt(retract_gear_speed_while_moving_selector **2 + selector_speed **2)
```
And the gear move distance is calculated accordingly.


thisiscam:
This is still WIP. The operation works, but I will need to test it more to see if it actually helps with the protruding problem.

moggieuk:
Got to look at your code.  Cute :-). Actually I don't think this is better way to "vector" selector and retraction move. Maybe the exact speed doesn't actually matter and just the back that it is retracting at all is enough..?

thisiscam:
> Got to look at your code. Cute :-). Actually I don't think this is better way to "vector" selector and retraction move. Maybe the exact speed doesn't actually matter and just the back that it is retracting at all is enough..?

Yeah. That speed probably doesn't matter. But I don't see a good way to program this without making the gear stepper go burrrrr --- gate traveling from 0 -> 18 is 17 times longer than 0 -> 1. So if we just set a gear travel constant distance, the gear will spin 17 times faster in the later case.

thisiscam:
BTW, I have ran this for a while now. I have only got 1 protrude since. And that was caused by me moving the ERCF around when debugging other hardware issues.

moggieuk:
Thanks for the update. I'm going to hold off on this PR until after the ERCFv2 lauch...

PR #78, opened 2023-11-03
potential fix to a bug where tip forming is skipped due to misplaced filament state after manual intervention

thisiscam:
The original condition is basically dead code that gets blocked by the outer `_is_mmu_paused`! 
If I understand correctly, `is_in_print` should be the correct condition and it was basically a typo? 

Either ways, I think this fixes a problem where during a MMU_PAUSE, I manually intervened but the MMU lost track of the filament state (since I extruded manually)

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -4417,7 +4417,7 @@ class Mmu:
 
         if self._is_mmu_paused():
             # Sanity check we are ready to go
-            if self._is_printing() and self.filament_pos != self.FILAMENT_POS_LOADED:
+            if self._is_in_print() and self.filament_pos != self.FILAMENT_POS_LOADED:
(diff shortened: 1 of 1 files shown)


thisiscam:
Actually, I don't understand why this condition exists at all.. Seems like we should always check for filament state consistency when resuming, regardless of in print or not.

thisiscam:
@moggieuk pinging since I'm 90% sure here's a bug

moggieuk:
Thanks. You might be correct. This is on the list for review this coming weekend..

moggieuk:
I'm on the road but review the PR and yes, you are correct it was intended to be checking for "is_in_print()".  But why I test for in print at all is a mystery..  if I can't remember the reason I'll just remove this check...

Issue #74, opened 2023-10-25
Possibly incorrect calculation of unload filament parking position


dflemstr:
With the latest Happy Hare code (v2.1.0-26-gae84c581) I'm getting this error 80% of the times that I try unloading filament:

```
After tip formation, extruder moved (park_pos): 40.00, encoder measured 72.00
MMU [T0] <<< [En] <<<<<<< [Ex] << [Ts] <. [Nz] (@40.0 mm)
An issue with the MMU has been detected whilst out of a print
Reason: Unload sequence failed: Filament seems to be stuck in the extruder.
Occured when changing tool: T0 > T1
```

My encoder is very well calibrated so it is surprising that the difference between the actually moved value and the encoder measured value is so large.

I suspect that the issue is that the encoder value is snapshotted here:

https://github.com/moggieuk/Happy-Hare/blob/ea39e0604878aacf0c49fe97cd9f970516f35783/extras/mmu.py#L3833

...but because the `_MMU_FORM_TIP_STANDALONE` moves the filament back and forth so much, the encoder value will likely be widely inaccurate because of the constant change of direction leading to sample aliasing. Or am I interpreting the code wrong?

This error gets propagated here, causing the final unload move to be too short and for the error to eventually be reported:
 
https://github.com/moggieuk/Happy-Hare/blob/ea39e0604878aacf0c49fe97cd9f970516f35783/extras/mmu.py#L3507C57-L3507C57

I suppose that this won't happen if the `output_park_pos` variable is set, which might be the reason why everyone isn't seeing this error?

dflemstr:
I should add that setting:

```
variable_output_park_pos: 20           # The park position (output variable passed back to Happy Hare)
```

...in the `[gcode_macro _MMU_FORM_TIP_STANDALONE]` seems to fix the issue for me. However it seems like this isn't intended as a viable workaround because the variable is no longer there in the latest version of the `mmu_software.cfg` file in this repo?

moggieuk:
The encoder reading is likely correct - remember the encoder only counts in one direction but a tip forming move moves backwards and forwards.  The info is presented in the debug message but NOT used to determine park position.  The park position is based on the aggregate actual stepper movement.

The variable_output_park_pos should be `-1` or not set at all which tells Happy Hare to believe the stepper movement.  If you set it that it will be used instead of stepper movement.

[...]

PR #96, opened 2023-11-26
Add support for specifying the distance between the gate endstop and encoder

dflemstr:
In my mission to make loading/unloading more reliable for my setup (see also #74), I realized that if both a gate endstop and encoder is defined, there seems to be an implicit assumption in the code that once the filament has reached the gate endstop, that it will engage with the encoder. 

This is usually not the case and I figured it would be helpful to make the distance between the encoder and gate endstop customizable.  This hopefully removes the error during for example bowden length calibration where the initial loading distance before the filament reaches the encoder seems to (wrongly) count towards the "slippage" of the filament.

This change is by no means complete; I might have missed places in the code and docs that need updating.  Feel free to leave comments and I will try to address them!

### DIFF
--- a/config/base/mmu_parameters.cfg
+++ b/config/base/mmu_parameters.cfg
@@ -108,6 +108,7 @@ gate_homing_max: 70			# Maximum move distance to home to the gate (actual move d
 gate_unload_buffer: 50			# Amount to reduce the fast unload so that filament doesn't overshoot when parking
 gate_load_retries: 2			# Number of times MMU will attempt to grab the filament on initial load (max 5)
 gate_parking_distance: {gate_parking_distance}		# Advanced: Specifies parking postion in the gate (distance from gate endstop/encoder)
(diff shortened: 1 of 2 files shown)


dflemstr:
Makes sense; I think I would prefer to implement this change and keep using both encoder + gate endstop if possible, since as I mentioned in the beginning I'm on the hunt towards improving reliability and accuracy as much as possible.

For reference, I'm using this mod for the Tradrack: https://www.printables.com/model/559345-binkyrack-a-tradrack-encoder-mod
...but instead of using the `Replacement_TR_Parts_to_Print/TradRack Head MGN9-C.stl` file I simply kept using the original Tradrack part that has the gate endstop, hence I get both a gate endstop as well as an encoder!

moggieuk:
Got it.  Yes what you are doing is perfectly fine and homing to a switch is faster and less "jerky" than using the encoder, although I've never had issues gate parking with just the encoder.   Any looking into fix for the next release.

PR #115, opened 2023-12-17
MMU gate endstop fixes

### LINKED ISSUE #114: Impossible to define a gate endstop due to names being misaligned
The filament sensor for the MMU gate is sometimes in the code called `"mmu_gate"` and sometimes just `"gate"`. Parts of the documentation before the addition of `[mmu_sensors]` made it sound like the endstop could be fully customized, but now it seems like it absolutely needs to be called exactly `"mmu_gate"`.

I think that either enforcing the convention of calling it `"mmu_gate"` or else consistently allowing the name to be customized would fix this bug.  I'll create a PR corresponding to the first approach.

dflemstr:
Attempts to address MMU gate endstop naming inconsistencies by using shared variables/constants to hold the endstop names.

Fixes #114.

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -140,8 +140,8 @@ class Mmu:
     ENDSTOP_EXTRUDER_TOUCH     = "mmu_ext_touch"
     ENDSTOP_GEAR_TOUCH         = "mmu_gear_touch"
     ENDSTOP_GATE               = "mmu_gate"
-    ENDSTOP_EXTRUDER           = "mmu_extruder"
-    ENDSTOP_TOOLHEAD           = "mmu_toolhead"
+    ENDSTOP_EXTRUDER           = "extruder"
+    ENDSTOP_TOOLHEAD           = "toolhead"
     ENDSTOP_SELECTOR_TOUCH     = "mmu_sel_touch"
     ENDSTOP_SELECTOR_HOME      = "mmu_sel_home"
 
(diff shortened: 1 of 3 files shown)


extras/mmu.py
```diff
@@ -140,8 +140,8 @@ class Mmu:
     ENDSTOP_EXTRUDER_TOUCH     = "mmu_ext_touch"
     ENDSTOP_GEAR_TOUCH         = "mmu_gear_touch"
     ENDSTOP_GATE               = "mmu_gate"
-    ENDSTOP_EXTRUDER           = "mmu_extruder"
-    ENDSTOP_TOOLHEAD           = "mmu_toolhead"
+    ENDSTOP_EXTRUDER           = "extruder"
+    ENDSTOP_TOOLHEAD           = "toolhead"
```
dflemstr, line 144: Maybe it isn't desirable to change these names; please comment if I should change it back to the `mmu_*` format instead.  I switched to these names because that seems to be what e.g. the `[mmu_sensors]` seems to generate, and also the names that the old config format used to use.

moggieuk: Not quite sure what isn't working, but the convention I used is that all endstops for Happy Hare are:
`mmu_XXX`

The filament sensor rendering is really for cosmetics. They are named `mmu_XXX` for sensors that are part of the MMU, and just `XXX` if they are part of the printer.    The "extruder" and "toolhead" have always been named without the mmu because they started life as standalone sensors.

PR #120, opened 2023-12-19
Addresses issue where if your start_up macros changes extruder temp m…

mojotaker:
Addresses issue where if your start_up macros changes extruder temp multiple times before printing; _mmu_pause saves the first target_temp, and ignores the others. This creates a problem for resume/unlock/pause where for e.g. the wrong extruder target_temp value is stored in "paused_extruder_temp" causing undesired behavior.

  -- I think anytime the _mmu_pause function is called, it should use the last/current extruder target_temp.

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -2273,8 +2273,7 @@ class Mmu:
 
     def _mmu_pause(self, reason, force_in_print=False):
         run_pause_macro = False
-        if not self.paused_extruder_temp: # Only save the initial pause temp
-            self.paused_extruder_temp = self.printer.lookup_object(self.extruder_name).heater.target_temp
+        self.paused_extruder_temp = self.printer.lookup_object(self.extruder_name).heater.target_temp
         self.resume_to_state = "printing" if self._is_in_print() else "ready"
 
         if self._is_printing(force_in_print) and not self._is_mmu_paused():


moggieuk:
It don't think this is correct.  You will notice that the other side of this is under _mmu_unlock() and _mmu_resume() which resets the saved paused temperature...
The logic is simply:
on pause, remember the current target temp 
on unlock, restore the saved temp
on resume, clear the stored temp

moggieuk:
I'm going to close this out unless I hear from you and let me know if I've missed something.

mojotaker:
Sorry for the late reply. I havent been able to go back and review. But from my findings and experience,  the flow does not work like you outlined (it used to work very well until about a month or 2 ago.).  I make use of 2 macro commands to begin my print.
1. for heatsoak.
2. and the other is the actual print_start.

The heatsoak macro sets my extruder temp to 200c. and pauses the printer for soaking  -> after HeatSoak, print_start now sets the extruder temp to 240c.
Any event that interupts printing, and causes a mmu pause; either unlocking or resuming tries to resume at 200c.

After I made the changes in the pull request, it fixed this issue for me.

mojotaker:
Ok figured it out. (still could be wrong) If printer gets paused the first time, outside of "MMu control" and does not resume using resume, then `self.paused_extruder_temp = None` never gets called. So the initial extruder_temp (heatsoak temp) never gets cleared. So when an mmu_pause occurs, 200c is whats saved.

mojotaker:
Hi @moggieuk  been using Happy-Hare with these changes for a while now, and im not having any problems.

moggieuk:
Thanks for looking at this again. I've been a bit buried with work the last couple of weeks but will follow your observations closely and see.  I'm not doubting there might be a problem, just that your solution didn't seem right (in that it didn't cover all cases), but I'll look again..

mojotaker:
Thank you too. I can spend more time doing a deep dive, and post all my
observations

moggieuk:
With the v2.4 release there was some subtle changes to the temp setting logic.  It should behave the same way, but the code is very slightly simpler and thus less error prone.  Let me know how it goes and if you do spot something unexpected can you attach the `mmu.log` .. that will help be pinpoint.

omnidynmc:
2.4  still does not fix this issue. Happy Hare MUST NOT store the extruder temp on first pause only. HEAT_SOAK macros in the beginning of the start up macros will use pause while it spins to detect once the chamber has reached temp. The problem is Happy Hare (ERCF did this too and I had to ERCF_DISABLE/ERCF_ENABLE before/after the heat_soak call) takes the first extruder temp and stores that, which in the HEAT_SOAK macro is extruder_temp * 0.75. Then if there is an error and you correct the error and resume, HH will change the extruder back down to that initial temp which is NOT the printing temp and it'll fail.

You can't use MMU ENABLE=0 and MMU ENABLE=1 to wrap it either because HH doesn't correctly detect the print start to set everything up and it doesn't expose on_print_start(pre_start_only=True) so you can't manually set HH back to initial print start phase. Either HH needs to stop only accepting the initial print temp, or there needs to be a way to disable HH until after the HEAT_SOAK line runs and reenable it and force it to believe we're in started state.

mojotaker:
> 2.4 still does not fix this issue. Happy Hare MUST NOT store the extruder temp on first pause only and the key here is that the HEAT_SOAK doesn't call resume so HH never detects it to clear. HEAT_SOAK macros in the beginning of the start up macros will use pause while it spins to detect once the chamber has reached temp. The problem is Happy Hare (ERCF did this too and I had to ERCF_DISABLE/ERCF_ENABLE before/after the heat_soak call) takes the first extruder temp and stores that, which in the HEAT_SOAK macro is extruder_temp * 0.75. Then if there is an error and you correct the error and resume, HH will change the extruder back down to that initial temp which is NOT the printing temp and it'll fail.
> 
> You can't use MMU ENABLE=0 and MMU ENABLE=1 to wrap it either because HH doesn't correctly detect the print start (which leaves the encoder and extruder sync off) to set everything up and it doesn't expose on_print_start(pre_start_only=True) so you can't manually set HH back to initial print start phase. Either HH needs to stop only accepting the initial print temp, or there needs to be a way to disable HH until after the HEAT_SOAK line runs and reenable it and force it to believe we're in started state post heat soak, or we need a method to tell HH to forget that initial pause temp so the next error/pause sets it correctly without resetting anything else.

Nailed it. More eloquently explained than i could.

moggieuk:
Thank you @omnidynmc that is very clear.  The example was what I needed.  So if I'm thinking clearly there are two ways to address this:
1) I can change so that the pause temp is recorded on every call to pause (that wasn't already in a pause_resume paused state)
2) I can clear the paused temp in `CANCEL_PAUSE` hook.  Which the chamber warming must call to reset, right?

Thoughts?

omnidynmc:
Option (2) isn't horrible, assuming once the heat soak is done we can just call CLEAR_PAUSE to get it to clear that temp state. It seems like it'd be more preferable to just have HH ignore everything happening during the heat_soak macro, however, wrapping it in MMU ENABLE=0/1 doesn't work without hacking a call to _on_print_start(pre_start_only=True) to fool HH into thinking we just started again. The other issue is I had to add G4 waits to give HH time to "settle" and get it initialize callback to happen.

moggieuk:
@omnidynmc @mojotaker 

I've pushed what I believe is the best solution (and simply too) into my "development" branch.  It would be great if you can validate (don't do any enable/disable workarounds):
```
cd ~/Happy-Hare
git checkout development
git pull
./install.sh
```
The `install.sh` is just a precaution and if you are on v2.4.1 is unnecessary.
Don't stay on "development" branch after I push to main though -- no idea what I might do..

mojotaker:
Thanks a lot, I will give it a try as soon as I can

moggieuk:
Would like to hear back from you.  It's working fine as far as I can tell and I've really tried to break it.

BTW Don't update development again now because I'm in the middle of something else which may not be 100% stable. It will be back in shape in about 24h.  I.e. ping me before you update again.

omnidynmc:
I missed the window to pull it unfortunately. If you can branch it in a safe spot I can pull the branch.

omnidynmc:
Looks like the commit to test is 05d7dc7 ?

moggieuk:
Sorry, for the delay.   All changes are in `main` branch now.  Let me know if you still run into issues but I'm pretty sure the corner case you found has been addressed.   If you verify I can close this PR because it isn't needed anymore.

moggieuk:
Closing.  Please reopen I you experience the problem again.

PR #136, opened 2024-01-04
Fix fetching spoolman spool data

vertexbz:
Fixes way spools are fetched from spoolman, former filament.id was looking up filaments instead of spools - according to spoolman api fetching them one by one seems to be only way


moggieuk:
Thanks!  I must admit I did charge through this implementation.  Let me look at this and remind myself of what was what doing vs what you are proposing...

vertexbz:
Thank you very much! This comes from case I hit - wasn't adding spools in the same order 😅

PR #137, opened 2024-01-04
Make sensors and servo definition order independent

vertexbz:
If sensors, servo or encoders are defined after emu section lookup_object will fallback to none None, o avoid this it can be changed to load_object

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -663,7 +663,7 @@ class Mmu:
 
         # Setup filament homing sensors ------
         for name in [self.ENDSTOP_TOOLHEAD, self.ENDSTOP_GATE, self.ENDSTOP_EXTRUDER]:
-            sensor = self.printer.lookup_object("filament_switch_sensor %s_sensor" % name, None)
+            sensor = self.printer.load_object(self.config, "filament_switch_sensor %s_sensor" % name, None)
             if sensor is not None:
                 self.sensors[name] = sensor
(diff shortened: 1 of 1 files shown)


moggieuk:
I see your point, but isn't there a danger that if every module loads rather than looks up then we could get into an inter dependency issue.  Maybe that isn't pertinent to simple sensors...

Can you describe the order problem you ran into to help be understand

vertexbz:
That's the workaround I ultimately used  - have empty mmu_sensors defined before mmu and add pins later. 

Was thinking about making it more custom config proof - ensure sensors are loaded before mmu, definitely missed mmu_sensors here, maybe it is worth to ensure its loaded before mmu config too - added what is on my mind to the PR

moggieuk:
I can definitely add the load_object() to mmu.py to ensure mmu_sensors is loaded, but then I'd probably also want to ensure other hardware used by Happy Hare (and main mmu.py module) is also loaded...

I'll have to go through the code to be sure. Maybe it is just sensors that matter...

To help be understand your setup, you are individually `[including]` HH config files rather than the suggested:
```
[include mmu/base/*.cfg]
```
(where I'm careful to ensure the order expands correctly)

moggieuk:
Closing this PR because of inactivity.  The load order is important for HH so I don't want to make this change without carefull discussion.  If you feel it is still an issue, please reopen the PR and comment

PR #142, opened 2024-01-08
Add/subtract encoder dead space for bowden moves

dflemstr:
The new support for "encoder dead space" is fantastic, and I'm happy that the logic has been encapsulated well.

Throughout the code, though, it is not only enough to tolerate "dead space"-worth of errors during bowden moves, but to also actually try to move the filament across the dead space.  To be clear, this means that currently most bowden loads/unloads are failing for me, despite having manually increased the calibrated bowden length to try to compensate.

I created this PR mostly just to start a discussion, to illustrate the start of a possible solution.  There are however a few different options, for example:

 * Add/remove the dead space length during bowden moves automatically.  Might become tricky to track if someone for some reason wants to perform multiple moves within the bowden tube; in that case, the first move needs to have the correction applied, but subsequent moves in the bowden tube should not have the correction be applied.
 * Add another `FILAMENT_POS_*`  state to indicate that the filament is in the gate but hasn't reached the encoder yet.  This might also be useful for ERCF or other non-selector-endstop use cases?  This to me sounds like the "cleanest" solution but might require a lot of refactoring.
 * For the cases when there's a gate and extruder endstop defined, stop using precise moves all together, and only do homing moves, using the encoder to detect severe problems (eg. if no filament movement at all is happening, or filament is moving in the wrong direction, etc).  In other words, bowden load would always be a homing move to the extruder endstop with a huge max length, and bowden unload would always be a homing move to the selector gate endstop with a huge max length.

I can start implementing/testing one of these options if I get an indication of which approach is preferred (@moggieuk might have an opinion?)

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -3120,6 +3120,7 @@ class Mmu:
         if self.calibrated_bowden_length > 0 and not self.calibrating:
             length = min(length, self.calibrated_bowden_length)
         full = length == self.calibrated_bowden_length
+        length += self._get_encoder_dead_space()
         length -= self.mmu_toolhead.get_position()[1]
 
         self._log_debug("Loading bowden tube")
@@ -3186,6 +3187,7 @@ class Mmu:
             length = min(length, self.calibrated_bowden_length)
         full = length == self.calibrated_bowden_length
         if full:
+            length += self._get_encoder_dead_space()
             length -= self.toolhead_unload_safety_margin # Extra precaution against sync unload and small unload buffer
         length -= self.gate_unload_buffer
 


moggieuk:
Thanks.  Just acknowledging your PR.  I need to sit down with a clear head to think through this.  It would be great if you could verify if my "development" branch still suffers from this problem .. I did make a few changes in this area.  This brach will be released shortly as v2.4.

dflemstr:
Thanks for the detailed reply!

> Choose switch as endstop
I would of course prefer this as it is the most precise, and not susceptible to encoder issues; even with Binky encoder I have occasionally had to clean it due to ABS/ASA dust building up from the filament/encoder wheel rubbing against the encoder casing.  The switch sensor is not susceptible to dust to the same degree.

So what I'm hearing you say in this case is that if we say "let's not add another filament pos state machine thing" and "let's not do homing moves," but we still want to use the gate switch and not the encoder, there are still a few different options.  Each option might correspond closely to what HH is currently doing/what you plan to have it do/what I'm thinking/what you're thinking, but let's just lay them out:

 1. Bridging the encoder gap is only a calibration concern, and after calibration the bowden move is simply performed according to `calibrated_bowden_length` from the gate endstop to the extruder endstop.  In this case, one would have to ensure that:
     * The dead zone is not counted towards "slippage" during calibration so that it gets correctly added to `calibrated_bowden_length`, in other words that one either trusts the gear rotation distance or else adds the dead zone value to the read encoder value (or both, and compare them)
     * The dead zone is subtracted from the encoder measurement during actual bowden loading/unloading.
     * If one does any partial bowden moves, e.g. `MMU_TEST_MOVE`, `MMU_RECOVER` or `MMU_EJECT`, that the dead zone is taken into account (this to me sounds like the trickiest part?)
 2. Bridging the encoder gap is done both during calibration and during "real" loading, ie the filament is moved across the dead zone distance from the gate endstop switch to the encoder both during calibration and during loading.  This means that:
      * Essentially the calibration behaves as if the bowden length starts at the encoder, while still using the (more reliable) switch for gate homing.
      * Most of the MMU logic could stay the same, oblivious of the encoder dead zone, as long as any load/unload move makes sure to bridge the dead zone gap.
      * For example, one assumption that would have to change is that the unload sequence expects the gate endstop to trigger once the `calibrated_bowden_length` has been moved; instead, HH would have to unload `calibrated_bowden_length` and then move to the gate switch (maybe this could be a homing move then?).   This is *sort of* what's already happening in the code but I'm not sure if the behavior is intentional/explicit right now?
      * ...is there a drawback to this approach otherwise?
 3. Fully dynamic bowden: I'd be curious to learn more about your thoughts here, and e.g. if an Annex Belay could be helpful or not in adjusting variables... Sounds like a lot of things could be adjusted on the fly but that might also be tricky (I'm picturing having to PID tune the encoder or gear rotation distance, heh).

moggieuk:
To be clear:
1. I don't think we need an addition state to achieve what you want but I'm still thinking about your suggestions.
2. Homing moves may be possible in the future but I want to see if I can solve the deceleration issue first.

You make some good points about the ancillary moves and the corner cases they introduce.   I'm gong to think through all the points you bring up before proposing how to move forward...

I do appreciate the thoughtfulness and understanding of the issue...

dflemstr:
Yep I'm 100% on board, and happy to let this simmer for a while.  In the meantime let me know if I can provide any additional data points, since I have a fully featured TR with all the switches set up already, so I can definitely help in that regard!

dflemstr:
tl;dr: everything seems to work quite well in the new version. I still seem to have issues during the extruder unload step but that might be unrelated to the topic of this PR.  Keeping my testing log down below for reference.

Observations/log from trying v2.4:

>I am using the default values from running `./install.sh` in this new version; however that means that my params for gate, gear, and selector calibration remain the same as per the pre-2.4 version.
>
>For this test, I re-ran the bowden calibration process (running `MMU_CALIBRATE_BOWDEN BOWDEN_LENGTH=1100 REPEATS=7`... yes, my printer is tall and has a super long bowden as a result :) ).  I'm using `REPEATS=7` to get a good baseline for a soak test later; normally I would use the default value for `REPEATS`.
>
>During the load path, it seems that the load is first homed to the gate endstop and then a long move is done which is great.  The loading pass seems to work well and the calibration successfully homes to the extruder switch.  The resulting distance seems to be measuring the encoder value only, not how far the filament actually moves (ie ignoring the dead zone).
>
>I see that the values trend/drift downward during calibration which I would attribute to the filament deforming during each repeat.  Values vary between 1159mm-1164mm which is only a 0.4% spread in values over that long distance, I think meaning that my encoder is working fine.  Result is `Recommended calibration bowden length is 1161.4mm. Clog detection length: 24.2mm`
>
>Again to get a reliable soak test, I re-run the bowden calibration again.  This is to eliminate the influence of the deformed filament a bit more.  I do realize that the resulting value will be inaccurate, since it measures the already-deformed filament while most filament loads would be with "pristine" filament, but again, I want to reduce variance for this test :smile: The spread is now 1154mm-1157mm which is a slight improvement.  I get `Recommended calibration bowden length is 1154.4mm. Clog detection length: 23.1mm`.
>
>To run a lot of load sequences, I set `mmu_num_gates: 1` (because I don't have all gates loaded at the moment), manually modify `mmu_selector_offsets` to only have a single element in it, and then run `MMU_SOAKTEST_LOAD_SEQUENCE LOOP=10 FULL=1` .  Would be lovely if this command supported a single gate/gate range!
>
>The soak test worked 6 times, but then got stuck during unload due to `Warning: Filament was not detected by extruder (entry) sensor at start of extruder unload` and then `Unload sequence failed: Unloading gate failed`.  What actually ended up happening was that the filament was not ejected enough to leave the extruder gears.  Time to find out why this is happening...

moggieuk:
Extruder unload is almost certainly because the tip creation is pushing the filament out of the extruder or it is configured incorrectly... the end of filament needs to be reported by cutting macro or set with `slicer_park_pos` if slicer is doing it (HH will monitor the standalone macro it run so that case is automatic).  HH expect the filament to still be in the extruder after this move, else if will (correctly) determine that the filament is stuck if no toolhead sensor and no encoder movement.

On to this PR and things to be done:
1) bowden calibration:
This is a new "manual" process that just measures gear movement while retracting but that aside, the auto calibration need to be redone I think:
a) If the starting position is the `mmu_gate` then all the encoder should be used for is to aid collision detection if no other way (i.e. no 'extruder' entry sensor or mmu_gear_touch not being used)
b) the measured move would probably be more accurate to use the actual movement of the gear stepper. Then dead space is unimportant (other than in the info message on what was actually measured).
c) If starting from encoder (or technically just past it) then then calibrated sample would be the amount initially measured by encoder (the overshoot) + the actual stepper movement as per comment b).

So in a nutshell I this a rewrite of the auto-calibration logic is not due because of the different sensor combinations.

2) I don't think anything special is required for `MMU_TEST_MOVE`, `MMU_RECOVER` or `MMU_EJECT`.  For test move the encoder movement is simply returned as information and does not influence the move.  Recovery will only use the encoder if there aren't other sensors to give sufficient feedback.  In the case the filament is in the dead zone, the `mmu_gate` sensor will have already reported presence so encoder is not needed.   Ejection doesn't care about dead space...(?)

3)

moggieuk:
I'll give a rewrite of "automated" bowden calibration a go tomorrow.  I can post and update in my "development" branch when done so you could try...

dflemstr:
Okay sounds great!

> Extruder unload is almost certainly because the tip creation is pushing the filament out of the extruder or it is configured incorrectly...

I have rather narrowed it down to some sort of noise in the filament sensor.  I have wired the switch to be NO (normally open) so that if the communication to the switch is broken, it is detected as a filament runout rather than happily continuing.  However some of the built-in pull-up resistors in the toolhead MCU seem to assume a NC switch, so I might be getting resistor values fighting each other slightly... But I think this is something I want to rule out on my end, I don't think it is Happy Hare related.

I might also switch to Filametrix; I already have a cutter on the toolhead but have been looking for a good servo mounting option; haven't had the time yet.  Might do that if it helps with debugging this issue!

> Ejection doesn't care about dead space...(?)

It would if `calibrated_bowden_length` would be computed from the encoder reading without adding the dead space value to it. In that case an extra dead space worth of unload movement would be needed to finally get the filament to the gate switch sensor.  With your approach of either just trusting the gear movement directly or manually adding the dead zone to the `calibrated_bowden_length` value, I agree it shouldn't be a problem.

dflemstr:
Something to consider for the extruder filament sensor malfunctioning case is to do something intelligent here:
https://github.com/moggieuk/Happy-Hare/blob/12c7d3e8bccfe4e249559abf24fb9f937035d347/extras/mmu.py#L3654-L3655

...ie if we know the filament was just in the extruder and left the extruder gears, would it be possible to do a blind move from the gears to where the filament sensor was expected to be, ie to basically move `hlength` without a homing move?

This would avoid the current issue that I face which is that the filament has barely left the extruder gears, but the extruder sensor isn't triggering, so HH assumes the filament is at the extruder sensor, and then tries a fast unload sequence with the MMU gear motor.  This of course usually results in the filament being stripped at the MMU gear motor, since the tip is being held by the extruder gears.  Doing a blind unload from the extruder would at least prevent that from happening.

moggieuk:
> This would avoid the current issue that I face which is that the filament has barely left the extruder gears, but the extruder sensor isn't triggering, so HH assumes the filament is at the extruder sensor, and then tries a fast unload sequence with the MMU gear motor. This of course usually results in the filament being stripped at the MMU gear motor, since the tip is being held by the extruder gears. Doing a blind unload from the extruder would at least prevent that from happening.

On this point.  HH has to assume it can believe the sensors else all kinds of problems would occur.  The logic tries to assume nothing and make the best possible assumptions (except where it would be pedantic to try to gain more info - especially if that would cause the servo to cycle up/down.

In this case, we are trying to unload the extruder and move filament to the homing point.  The extruder (entry) sensor says that the filament is not present.  Two things could be true: 1) the sensor is malfunctioning. 2) the unload_extruder() was called without filament in the extruder.  The best this method can do is assume that the filament is out of the extruder (true). It could consider it out but not homed (i.e. at end_of_bowden state). This would prevent the fast unload but instead to the slow homing unload all the way to the gate.

What am I missing?

moggieuk:
> It would if `calibrated_bowden_length` was computed from the encoder reading without adding the dead space value to it. In that case an extra dead space worth of unload movement would be needed to finally get the filament to the gate switch sensor. With your approach of either just trusting the gear movement directly or manually adding the dead zone to the `calibrated_bowden_length` value, I agree it shouldn't be a problem.

What I'm proposing for this is that the recorded "calibrated_bowden_length" remembers what was the gate starting point: encoder or gate.  Then adjusts from there.  Really it shouldn't practically matter so long as it was calibrated according to user setup but for me.. it would allow for dynamic switching between choice of gate homing which would be hugely helpful in testing.

dflemstr:
> What am I missing?

I agree that in general it sounds like a good principle to trust the sensors :sweat_smile: and in the end this is something I should fix with my hardware.

I guess where the logic *could* do something intelligent is that it could take into consideration that it is in the middle of an unloading process, and that it knows that the filament was just at the extruder gears (since the toolhead filament sensor registered filament a few secs ago).

...but this might be something that's not worth fixing.

moggieuk:
Ok. I have the changes for this (and some unrelated pause temperature save logic) in my "development" branch.

Ideally you would try before I push it live:
```
cd ~/Happy-Hare
git checkout development
git pull
```

Don't forget to switch back on "main" when I push this live (don't stay on development for too long -- no telling what I might do. lol)

dflemstr:
Testing this now!

dflemstr:
Log of my tests below.

Lazy way of checking that the encoder is OK:

```
$ MMU_SERVO POS=down
$ MMU_TEST_MOVE MOVE=100
Moved 100.0mm (measured 59.0mm)
$ MMU_TEST_MOVE MOVE=100
Moved 100.0mm (measured 101.0mm)
$ MMU_TEST_MOVE MOVE=100
Moved 100.0mm (measured 100.0mm)
$ MMU_TEST_MOVE MOVE=100
Moved 100.0mm (measured 100.0mm)
$ MMU_RECOVER
...
$ MMU_EJECT
Unload of -400.0mm filament successful (encoder measured 362.0mm)
```
Looks fine I suppose.  This was using slightly dented filament from having just completed a print job (ie gear teeth in the filament) so I expect there will be some material drift due to that.

Calibrating bowden:

```
$ MMU_CALIBRATE_BOWDEN BOWDEN_LENGTH=1100 REPEATS=7
Calibrating bowden length on reference Gate #0
Tool T0 enabled
Finding extruder gear position (try #1 of 7)...
Pass #1: Filament homed to extruder, encoder measured 1177.0mm, filament sprung back 0.0mm
- Bowden calibration based on this pass is 1177.0
Finding extruder gear position (try #2 of 7)...
Pass #2: Filament homed to extruder, encoder measured 1176.0mm, filament sprung back 0.0mm
- Bowden calibration based on this pass is 1176.0
Finding extruder gear position (try #3 of 7)...
Pass #3: Filament homed to extruder, encoder measured 1176.0mm, filament sprung back 0.0mm
- Bowden calibration based on this pass is 1176.0
Finding extruder gear position (try #4 of 7)...
Pass #4: Filament homed to extruder, encoder measured 1176.0mm, filament sprung back 0.0mm
- Bowden calibration based on this pass is 1176.0
Finding extruder gear position (try #5 of 7)...
Pass #5: Filament homed to extruder, encoder measured 1175.0mm, filament sprung back 0.0mm
- Bowden calibration based on this pass is 1175.0
Finding extruder gear position (try #6 of 7)...
Pass #6: Filament homed to extruder, encoder measured 1175.0mm, filament sprung back 0.0mm
- Bowden calibration based on this pass is 1175.0
Finding extruder gear position (try #7 of 7)...
Pass #7: Filament homed to extruder, encoder measured 1175.0mm, filament sprung back 0.0mm
- Bowden calibration based on this pass is 1175.0
Recommended calibration bowden length is 1175.7mm. Clog detection length: 23.5mm
Bowden calibration and clog detection length have been saved
```

... so seeing the usual deforming drift as expected, this time only with a 2mm span/0.17%.

Not sure where I can find the new calibration values, they don't seem to be saved in `mmu_vars.cfg`?

moggieuk:
The calibration should be saved to `mmu_vars.cfg`.  It is only not saved if you specify `SAVE=0`.   I'll double check but it was working (BTW don't refresh development .. I'm deep in another change).

The params in `mmu_vars.cfg` should look something like:
```
mmu_calibration_bowden_home = 'encoder'
mmu_calibration_bowden_length = 698
```

dflemstr:
Hmm you're right, I must've forgotten to restart something... retrying today after having restarted the host just to be sure.

Now unloading seems to consistently fail when calibrating:
```
$ MMU_CALIBRATE_BOWDEN BOWDEN_LENGTH=1100 REPEATS=7
Calibrating bowden length on reference Gate #0 using ENDSTOP 'mmu_gate' as gate reference point
Tool T0 enabled
Finding extruder gear position (try #1 of 7)...
Pass #1: Filament homed to extruder after 1176.7mm movement
(encoder measured 1184.0mm, filament sprung back 0.0mm)
- Bowden calibration based on this pass is 1176.7
An issue with the MMU has been detected whilst out of a print
Reason: Calibration of bowden length (on Gate #0) failed. Aborting, because:
Unloading gate failed
```

Maybe some logic needs to be added to take into consideration the right reference during unloading too? Just guessing, happy to look deeper into it if you prefer.


> (BTW don't refresh development .. I'm deep in another change).

Did you consider creating a separate branch per change? :sweat_smile:

moggieuk:
Hmm.. I'm far to disorganized...   usually it is one thing at a time, but I conflated two by accident.

> Did you consider creating a separate branch per change? 😅

I'll take a look. We are close I think.  It would be so much easier if I have an MMU with gate sensor. Lol.

moggieuk:
BTW base on my experience you want to occasionally change the piece of filament you are testing over. Once groves start to appear it upsets the repeatability of the tests

> so seeing the usual deforming drift as expected, this time only with a 2mm span/0.17%.

dflemstr:
Re-running the test to get the log files; I had reverted to `main` since I needed to keep having a working printer ... lucky I have `git reflog` so I am able to find the previous version of the `development` branch, heh!  I *think* you wanted me to test commit `9eb249d98cd66435707708e5884f96707d0b6648`, right?

Config stuff:
```
gate_homing_max: 70
gate_unload_buffer: 50
```

Log from `MMU_CALIBRATE_BOWDEN BOWDEN_LENGTH=1100 REPEATS=7`:

```
17:24:54 - DEBUG: Resetting TTG map
17:24:54 - DEBUG: Setting servo to move (filament hold) position at angle: 131
17:24:54 - DEBUG: Tool already unloaded
17:24:54 Calibrating bowden length on reference Gate #0 using ENDSTOP 'mmu_gate' as gate reference point
17:24:54 - DEBUG: Selecting tool T0 on gate #0...
17:24:54 Tool T0 enabled
17:24:54 - - TRACE: Setting gear motor rotation distance: 22.763648 (ratio: 1.000000)
17:24:54 - - TRACE: Processing idle_timeout 'printing' event
17:24:54 - DEBUG: Setting servo to down (filament drive) position at angle: 0
17:24:56 - - TRACE: Initial homing to gate sensor. Stepper: 'gear' homed after moving 16.1mm (of max 70.0mm), encoder measured 0.0mm (delta 16.1mm). Pos: @16.1, (0.0mm)
17:24:56 - DEBUG: Gate endstop reached after 16.1mm (measured 0.0mm)
17:24:56 - DEBUG: Loading bowden tube
17:25:03 - - TRACE: Course loading move into bowden. Stepper: 'gear' moved 1100.0mm, encoder measured 1082.0mm (delta 18.0mm). Pos: @1100.0, (1082.0mm)
17:25:03 Finding extruder gear position (try #1 of 7)...
17:25:03 - DEBUG: Homing to extruder 'extruder' endstop, up to 150.0mm
17:25:05 - - TRACE: Homing filament to extruder endstop. Stepper: 'gear' homed after moving 64.7mm (of max 150.0mm), encoder measured 65.0mm (delta -0.3mm). Pos: @1164.7, (1147.0mm)
17:25:05 - DEBUG: Extruder endstop reached after 64.7mm (measured 65.0mm)
17:25:06 - - TRACE: Aligning filament to extruder gear. Stepper: 'gear' moved 8.0mm, encoder measured 8.0mm (delta 0.0mm). Pos: @1172.7, (1155.0mm)
17:25:06 - DEBUG: Setting servo to up (filament released) position at angle: 131
17:25:07 Pass #1: Filament homed to extruder after 1172.7mm movement
         (encoder measured 1180.0mm, filament sprung back 0.0mm)
         - Bowden calibration based on this pass is 1172.7
17:25:07 - DEBUG: Unloading bowden tube
17:25:07 - DEBUG: Setting servo to down (filament drive) position at angle: 0
17:25:15 - - TRACE: Course unloading move from bowden. Stepper: 'gear' moved -1072.7mm, encoder measured 1080.0mm (delta -7.3mm). Pos: @-1072.7, (1080.0mm)
17:25:17 - - TRACE: Reverse homing to gate sensor. Stepper: 'gear' did not home after moving -70.0mm (of max -70.0mm), encoder measured 71.0mm (delta -1.0mm). Pos: @-1142.7, (1151.0mm)
17:25:17 - DEBUG: Did not home to gate sensor
17:25:17 - DEBUG: Setting servo to up (filament released) position at angle: 131
17:25:17 An issue with the MMU has been detected whilst out of a print
         Reason: Calibration of bowden length (on Gate #0) failed. Aborting, because:
         Unloading gate failed
17:25:18 - - TRACE: Processing idle_timeout 'ready' event
```

Logs from `MMU_EJECT` in case you're curious about the measurements:

```
17:26:27 - DEBUG: Unloading tool T0
17:26:27 - - TRACE: Running macro: _MMU_PRE_UNLOAD
17:26:27 Unloading filament...
17:26:27 - - TRACE: Running macro: _MMU_ACTION_CHANGED ACTION='Unloading' OLD_ACTION='Idle'
17:26:27 MMU [T0] <<< [En] <<<<... [Ex] .. [Ts] .. [Nz] 0.0mm (e:0.0mm)
17:26:27 - DEBUG: Setting servo to down (filament drive) position at angle: 0
17:26:27 - - TRACE: Processing idle_timeout 'printing' event
17:26:27 - DEBUG: Performing safety synced pre-unload bowden move
17:26:28 - - TRACE: Bowden pre-unload move. Stepper: 'gear+extruder' moved -5.0mm, encoder measured 6.0mm (delta -1.0mm). Pos: @-5.0, (6.0mm)
17:26:30 - - TRACE: Reverse homing to gate sensor. Stepper: 'gear' homed after moving -26.1mm (of max -1195.7mm), encoder measured 0.0mm (delta 26.1mm). Pos: @-31.1, (6.0mm)
17:26:30 - - TRACE: Final parking. Stepper: 'gear' moved -15.0mm, encoder measured 0.0mm (delta 15.0mm). Pos: @-46.1, (6.0mm)
17:26:30 MMU [T0] <.. [En] ....... [Ex] .. [Ts] .. [Nz] UNLOADED -46.1mm (e:6.0mm)
17:26:30 - DEBUG: Setting servo to up (filament released) position at angle: 131
17:26:31 Unload of -46.1mm filament successful (encoder measured 6.0mm)
17:26:31 - - TRACE: Running macro: _MMU_ACTION_CHANGED ACTION='Idle' OLD_ACTION='Unloading'
17:26:31 - - TRACE: Running macro: _MMU_POST_UNLOAD
17:26:31 - - TRACE: Processing idle_timeout 'ready' event
```

dflemstr:
>BTW base on my experience you want to occasionally change the piece of filament you are testing over. Once groves start to appear it upsets the repeatability of the tests

Yep, I'm only doing it like this in this case to make the test more reproducible.  But maybe it's not a problem since the drift is at most 2-5mm

moggieuk:
Anything is in "main" now.  I've tested and am pretty happy.   I'm happy to accept PR for anything I've missed but it seems complete to me....

dflemstr:
OK, thanks. I will test when my print queue is empty; hoping that will be later today.  If you want to wait for my validation that's fine by me, or if you want to close this PR I can re-open if I find anything...

dflemstr:
I have tested this for a while now and not noticed any issues, thanks!

PR #143, opened 2024-01-08
Provide sane defaults for mmu_vendor=Tradrack

dflemstr:
The 25mm endstop-to-encoder distance was determined empirically by homing to the gate endstop and advancing the filament 1mm at a time until the encoder detects movement.  This might not match the CAD 100% because of the different shapes of the detection mechanisms; the gate sensor uses a 4mm bearing ball and the encoder a BMG gear, so they might need different amounts of filament movement to trigger.

The last gate offset is likely typically 0 if you follow the official tradrack instructions, but might be higher if a longer rail is used.  For example, in my case I have a 600mm rail which supports 34 filament lanes, by using a MGN9H block and allowing the block to slide ~3mm off the rail (where the rubber lubrication block is, before the bearings in the block risk being lost).  The movable rail stopper can be used to position the rail block correctly. This means that there's really 0mm distance between the last gate and end-of-travel.

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -303,7 +303,7 @@ class Mmu:
             self.cad_gate0_pos = 0.5
             self.cad_gate_width = 17.
             self.cad_bypass_offset = 0 # Doesn't have bypass
-            self.cad_last_gate_offset = 1. # TODO this is a guess
+            self.cad_last_gate_offset = 0. # This can vary a lot depending on the length of the rail
 
             self.gate_parking_distance = 17. # Using Gate switch (had user reports from 15 - 17.5)
             self.encoder_default_resolution = bmg_circ / (2 * 12) # If fitted, assumed to by Binky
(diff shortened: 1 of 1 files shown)


moggieuk:
Thanks for the feedback.  In my upcoming v2.4 push I have taken a lot of the logic out of the main Happy Hare module for differences in vendors (limit it to cad dimensions now,  The image before shows some starter values I have on initial setup to create a "starter" config files.  

param

I got a dimension of 31 from another user, but happy to change to 25 if that is correct. Is there just one Binky encoder modification to Tradrack?

Also, a dimension of 0 for the last gate offset now means pretty much no limit to accomondate TR design.  Let me know how v2.4 works and them I'm happy to accept PRs to improve from there...

dflemstr:
> Is there just one Binky encoder modification to Tradrack?

I think so, yes!  Although maybe some folks are using the original ERCF encoder with the "Binky" TR mod too.  I don't think that would explain the 31->25mm difference though...

> Also, a dimension of 0 for the last gate offset now means pretty much no limit to accomondate TR design. Let me know how v2.4 works and them I'm happy to accept PRs to improve from there...

Sure, I'm happy to hold off and test the new version.  The gate auto calibration hasn't worked well for me anyway since I use a way too powerful selector motor for stallguard.  Is there a way to reduce the motor current during selector homing that I've missed in the code/docs?

moggieuk:
Ok. New version v2.4 with install support for Tradrack has been launched.

I'm going to close this PR but please submit new ones with any mistakes you find. I'm anxious to provide the same love to Tradrack that ERCF has recieved...

PR #167, opened 2024-01-28
remove double subtraction of gate_unload_buffer

thisiscam:
Gate_unload_buffer appears already subtracted in unload_bowden -- this fix my broken bowden calibration


moggieuk (commented):
Thanks, yes. This is legacy. Probably would still unload but fast movement would be short.

Issue #184, opened 2024-02-15
Filametrix cut with restore position tries to return to 0,0 usually destroying pin located near there.


jwhite:
Title pretty much says it all.

If you do a filametrix cut directly after power on, and are using restore position the toolhead attempts to return to 0,0.

Most people with cable chains are pultting the filametrix pin there, so it than attempts to remove said pin.

Pull request incoming.

moggieuk:
Merged, but I did then rework to simply move the position saving to AFTER the optional homing to achieve the same thing.

PR #185, opened 2024-02-15
Fixes issue #184 - Filametrix cut with restore position tries to return to 0,0 usually destroying pin located near there.

jwhite:
Adds a save state command.


moggieuk (approved):
I'll pull this but probably will simply move the save state until after the homing rather than before.

PR #191, opened 2024-02-20
Fix unable to resume after clog

jwhite:
Calling through to base even if mmu state isn't paused.

The root issue is that fluidd supplies its own pause and resume macros.

Happy hare also provides pause and resume, but only calls through to the base resume in the case that the mmu internal state is paused.  This is insufficient for supporting the state also being maintained by fluidd.  This change ALWAYS calls through and allows the fluidd macro to update the UI and begin printing.

### DIFF
--- a/config/optional/client_macros.cfg
+++ b/config/optional/client_macros.cfg
@@ -12,6 +12,7 @@ description: Pause the print and park
 gcode:
     {% if printer.pause_resume.is_paused %}
         RESPOND MSG="Print is already paused"
+        BASE_PAUSE
     {% else %}
         {% if printer.extruder.can_extrude %}
             G92 E0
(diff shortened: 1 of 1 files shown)


moggieuk:
I have a question about this.  While it seems fairly safe, it is going to cause duplicate "print is pause/print is not pause" messages.  I'm not looking at whether HH thinks it is pause, but at whether the base klipper thinks it is paused or not.

Did you test without the Fluidd macros?

jwhite:
Fluidd macros are required by Fluidd and are provided in a read only file, so hard to turn off.

jwhite:
I wonder if Fluidd requires this VELOCITY property that I see here:

RESUME_BASE VELOCITY={params.VELOCITY | default(sp_move)}

moggieuk:
Any parameters passed to RESUME will be passed down by HH including VELOCITY.   But something isn't right here because klipper cannot function if the pause_resume state is not correct.  What HH does checking the pause_resume.is_paused is EXACTLY what will be done in base_pause.   So even if we bypassed the check in HH it would hit the same check in klipper...

My best guess is that either something is silently failing or somehow a "CANCEL_PAUSE" is creeping into the workflow but I can't see how..

jwhite:
Closed.  If there is a fix here, it has to be at the root of the issue.

PR #199, opened 2024-02-24
Added OctoPrint Service Detection, Added condition in Mookraker direc…

JesNoFear:
…tory detection, Adjust Klipper Config path if OctoPrint Detected

Add OctoPrint Detection - Runs before directory check, sets OCTOPRINT to 0 or 1, used as a condition in the directory check.
Skip Exit if Moonraker directory not found if OctoPrint detected (last condition) - specifically looks for printer.cfg in the ~/ directory since the directory will always exist.
Adjust Klipper Config path if OctoPrint Detected

**tried to cherry pick out lines 1055 and 1073, but was unsuccessful.**

moonraker calls later in the script are soft errors and do not cause the script to abort. I started to edit them to skip, but didn't want to potentially break functionality if users have mainsail or fluidd and have partial or lite installs of OctoPrint. But could be mitigated by setting a 0/1 if moonraker directory is found instead of if OctoPrint is found.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -15,6 +15,7 @@ ARGS=( "$@" )
 KLIPPER_HOME="${HOME}/klipper"
 MOONRAKER_HOME="${HOME}/moonraker"
 KLIPPER_CONFIG_HOME="${HOME}/printer_data/config"
+OCTOPRINT_KLIPPER_CONFIG_HOME="${HOME}"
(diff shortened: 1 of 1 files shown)


moggieuk (approved):
Thanks for putting this into a PR. Very helpful!!

PR #200, opened 2024-02-25
Fix typo preventing sync multipliers from updating

seth-cohen:
Simple typo fix. Was having issues using the `MMU_TEST_CONFIG` command to update these two runtime values. Aaaand... then I found out why :D


moggieuk (approved):
Thanks!!

PR #207, opened 2024-02-28
Fix to the mmu pre check and flushing volumes

Dendrowen:
Due to the change and flexibility in recent updates to flushing volumes, a regex group was added to the regex while the data extraction was still gathering group 1. I changed this to group 2.

The pre-check for gates ran into an 'mmu not home' error. I added a conditional MMU_HOME move in that check.


moggieuk (approved):
Using unicode instead of character to support Python2..

PR #216, opened 2024-03-07
Fixed color pre-processing mismatch in SS. empty colors were picked up

Dendrowen:
BEWARE! I have not tested this with slicers, because I don't have SS set up. However, testing in a simulated environment seemed to work.


moggieuk (approved):
Looks ok. I'll merge but let me know if other problems are found..

PR #220, opened 2024-03-10
Update blobifier.cfg

Saikedo:
Replaced servo with mmu_servo to resolve multiple issues encountered with Klipper's default servo library, particularly the WIDTH=0 issue, which was causing unexpected servo movements. 

Additionally, corrected an inaccurate comment regarding minimum_pulse_width; it actually controls the servo's outward position, contrary to the previous assertion. The comment has been updated to reflect its true function.

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -38,13 +38,13 @@
 ##########################################################################################
 # The servo hardware configuration. Change the values to your needs.
 # 
-[servo blobifier]
+[mmu_servo blobifier]
 # Pin for the servo.
 pin: PG14
-# Adjust this value until a 'BLOBIFIER_SERVO POS=in' retracts the tray fully without a 
+# Adjust this value until a 'BLOBIFIER_SERVO POS=out' extends the tray fully without a 
 # buzzing sound
 minimum_pulse_width: 0.00053
-# Adjust this value until a 'BLOBIFIER_SERVO POS=out' extends the tray fully without a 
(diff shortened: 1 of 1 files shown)


moggieuk:
I have no problem with this PR (definitely should be using mmu_servo), but the direction is likely related to type of servo -- MG90 moves in opposite direction to Savox for example.

@Dendrowen are you able to validate the movement comment? Then I can merge..

Dendrowen:
The mmu_servo is probably the way to go. I've had a small number of complaints regarding the WIDTH=0

The switched positions is very setup dependant. Simply swapping them because it doesn't suit your specific setup is a no-go.
@moggieuk, how does the install script handle these settings? If a user changes them, does the install.sh restore it to the default?

moggieuk:
@Dendrowen user changes to "variables" are preserved whilst updating the rest of the config/macro.

So looking at this change, is the change to comment appropriate?

Dendrowen:
Looking at the code once more, @Saikedo seems to be right. PR approved :)

Dendrowen:
Looking at the code once more, @Saikedo seems to be right. PR approved :)

Saikedo:
@moggieuk The documentation for Blobifier does not even mention Savox for now and explicitly asks for SG90 or MG90  so the comment change in my opinion is certainly a plus. Also for Savox, we usually keep the min/max pulse width almost the same right? It`s the BLOBIFIER_SERVO that will need to change for Savox.

I can get my hands on a Savox to experiment and open another PR by the way if you guys think that might be worth it.

moggieuk:
Got it. Thanks. Merging.

PR #223, opened 2024-03-10
Add ability to specify alternative service name from kiauh install

ejsears:
There was a request from an user on Discord to be able to specify an alternative service name for klipper only:

https://discord.com/channels/460117602945990666/909743915475816458/1216405832456212652

While this is somewhat accomplished with the `-r` flag, this depends on the Repetier-Server. 

This added functionality allows setting the `${KLIPPER_SERVICE}` variable to a specific name. This would commonly be used if a user installed multiple klipper services with the commonly used `kiauh` utility.


moggieuk:
No problem. Merging now.

PR #264, opened 2024-04-09
replace egrep calls with grep -E

bakatrouble:
`egrep` and `fgrep` aliases have been deprecated for over 15 years, and for almost 2 years they are raising warnings (including on Arch Linux, which I'm running this software on): https://www.phoronix.com/news/GNU-Grep-3.8-Stop-egrep-fgrep

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -385,7 +385,7 @@ cleanup_manual_stepper_version() {
 
     # Upgrade mmu_hardware.cfg...
     hardware_cfg="${KLIPPER_CONFIG_HOME}/mmu/base/mmu_hardware.cfg"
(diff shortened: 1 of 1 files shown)


moggieuk:
What!? I grew up with egrep. Am I really that old.  :-)

I've never seen a warning on a system before, but easy to fix.

moggieuk (approved):
Looks good. I'll merge tomorrow when I can sanity test. Thanks.

bakatrouble:
> What!? I grew up with egrep. Am I really that old. :-)

You're not alone
I only found out because of these warnings as well, looks like it was a secret well kept in plain sight until then, huh

Issue #254, opened 2024-04-05
All gcode files are modified after a print is finished


eMapleLuke:
I use the ERCF V2 with orca slicer. When I after a print is finished, all gcode files sliced with more than one color are modified. All have the same date and time, and the tracker of printed parts dont work anymore, because the files are recognized as new files.

iamnotahippy:
I have the same issue. It's breaking my gcode archiving script as well.

iamnotahippy:
> why do you believe happy hare is updating the g-code?

In moonraker.log it shows mmu_server.py attempting to preprocess every gcode file on the system.

ningpj:
Ok, the preprocessor & parsing and expanding placeholders/tokens like !referenced_tools!
This is triggered on gcode upload and was trying to work out why file timestamps on the gcode would be updated post print.  Do you use place holders like !referenced_tools! as you can disable to preprocessor in moonraker.conf

```[mmu_server]
enable_file_preprocessor: False
enable_toolchange_next_pos: False```

eMapleLuke:
> Ok, the preprocessor & parsing and expanding placeholders/tokens like !referenced_tools! This is triggered on gcode upload and was trying to work out why file timestamps on the gcode would be updated post print. Do you use place holders like !referenced_tools! as you can disable to preprocessor in moonraker.conf
> 
> ```
> enable_file_preprocessor: False
> enable_toolchange_next_pos: False
> ```

When I change these to false, what does it change? Do I have any disadvantages with it?

ningpj:
Your log doesnt seem to have the message "Processing Uploaded File:...." preceding the shell_command.py call out...
What version of moonraker are you running?  I'm current and running v0.8.0-325-g99b97afc and only see moonraker triggering the preprocess script each time i upload a new gcode prior to printing it.

eMapleLuke:
> It turns off the preprocessor so HH wont be able to validate required gates have filament ready to go. more detail here. https://github.com/moggieuk/Happy-Hare?tab=readme-ov-file#11-useful-pre-print-functionality https://github.com/moggieuk/Happy-Hare/blob/main/doc/gcode_preprocessing.md
> 
> So the process flow was you uploaded and printed a gcode file and at the conclusion of the print all mmu gcode files in the directory were updated again? Are you using fluidd/mainsail or octoprint to upload and print your files?

I use mainsail. Yes, after a print is finished, all the files I uploaded with more than one filament is updated, mot the old ones.

[...]

PR #271, opened 2024-04-16
Fix gcode rewrites

mpsnp:
Fixes https://github.com/moggieuk/Happy-Hare/issues/254. Ideally we should not rewrite files, but rather extract metadata for initial checks and save it elsewhere: either in moonraker or manually in filesystem.

Also I did add marker for already processed files, so script can early exit and don't read whole gcode file for processed ones.


moggieuk:
Thank you for this.  I was just about to look into it based on a couple of private messages and I thought, "let me run through PR's first" and bingo!  :-)

PR #278, opened 2024-04-22
Update mmu_server.py

Saikedo:
There was an extra `"` left at the end of the string. Most likely a leftover from NEXT_POS version.


moggieuk (approved):

moggieuk:
Thanks. Yes, that's an bug!

PR #283, opened 2024-05-03
Update configuration.md

NortonTech3D:
Updated the CW2 dimensions for the Rapido/Rapido2 hotend setup with verified dimensions


NortonTech3D:
Error with Pull Request, will resubmit

PR #291, opened 2024-05-09
Automatic pull-back of filament after loading long bowden cables

Cyberwizzard:
My Voron is using a long bowden cable, which is also extra wide as 2mm resulted in high friction loads and unloads.

The result is that after a tool swap, the filament was pushed into the extruder head and is slack, resulting in the clog detection triggering right after.

This is not a complete pull request as the amount of rollback and torque should probably be adjustable from the configuration files. Also, I still have trouble getting reliable swaps, with or without this change - so either I am missing something or I need to configure my printer better.

Anyway, comments welcome.

**What does this do?**
At each tool load, the filament is pushed down the tube, picked up by the extruder and then put into the parking position inside the extruder (both the ERCF feeder motor and extruder motor run in parallel for this).
This new code then reverses the feeder motor on the ERCF (after reducing the pulling torque) to pull the filament tight from the ERCF side before disengaging the servo (letting go of the filament).

At this point, the filament should be tight and as soon as the extuder pulls it in further, the encoder wheel should move, thereby accurately showing that the filament is indeed loaded and fed through.

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -4062,6 +4062,15 @@ class Mmu:
             self._set_filament_pos_state(self.FILAMENT_POS_LOADED)
             self._log_debug("Filament should loaded to nozzle")
 
+            # Roll back up to N mm using low torque to pull the filament tight in long bowden tubes with extra space
+            self._log_info("Filament pull after load using low torque")
+            self._adjust_gear_current(percent=30, reason="Pull filament tight")
+            self._set_filament_direction(self.DIRECTION_UNLOAD)
+            
+            _,_,measured,delta = self._trace_filament_move("Loading filament to nozzle", -25, speed=speed, motor="gear", wait=True)
+            self._log_info("Filament pull: rolled back %0.0f mm" % measured) 
+
+
     # Extract filament past extruder gear (to end of bowden). Assume that tip has already been formed
     # and we are parked somewhere in the extruder either by slicer or by stand alone tip creation
     # But be careful:


moggieuk:
Oh, this is an interesting problem.  So after completed loading the filament has more slack than the clog detection threshold so that it triggers a clog before the encoder has moved?

What is your clog detection threshold?  I see the theoretical issue but haven't heard of the problem before.

Cyberwizzard:
@moggieuk I suppose you mean mmu_calibration_clog_length? That is set to 33.5 at the moment, but I did not lower if after my changes to be honest.

I also have some issue with the dynamic clog detection still triggering on swaps - where the static one is now working after the pull back method.

About the need for it when synced: I suppose then it does not matter that it acts like a spaghetti noodle in the bowden tube - as the encoder will measure right after the ERCF and the ERCF pushes it into the tube, the clog detection would likely not see the difference. So I suppose it should depend on sync_to_extruder=0 for the pull-back to happen?

moggieuk:
Yes, to you comment about "spaghetti noodle" -- with synced gear/extruder, the encoder will reliably see movement.

Getting unreliable clog detection is often because of too high extruder stepper movement.  It seems that almost everyone has a purge that is too fast for the extruder and the missed steps will (correctly) trigger a clog.  If this is the cause you can disable and re-enable during this operation but it is better to reduce purge speeds.

33.5mm clog detection length is very long and I'm sure if you set it to automatic mode it would tuning in back down.  10mm would be more typical.  That said, I do want to think about the "spaghetti noodle" fix to tighten the filament when printing without synchronization.  It's on my TODO list.

christi-ko:
I can confirm the same issue with a bowden tube length of about 1000mm. I had several clog detections during my first prints, now i enabled sync_to_extruder = 1 and the problem is gone ...
If you need any measurements or more information about my configuration to support further improvements just let me know.

moggieuk:
Testing this now. Will be in the next patch to v2.6.0

christi-ko:
Hi i tested it the last two days and it's working fine, no clog detection directly after load / unload. But still some clogs during "normal" printing. Now increasing the head room.

PR #293, opened 2024-05-14
Fix resume after pause during toolchange

bakatrouble:
My use case: I've installed Blobifier, so if an error happens during toolchange, klipper saves pause position at the Blobifier position, right on the tray. I do not want the print to continue from that position, so I call `_MMU_SAVE_POSITION` in my `PAUSE` macro and then restore the saved position in the `RESUME` macro. Before this change, I had nozzle dragging across both printed parts and the tray edge back to the saved position on the Blobifier tray.

Now the position at which was the toolchange initiated is set as the return position used by klipper's resume logic: https://github.com/Klipper3d/klipper/blob/master/klippy/extras/pause_resume.py#L83

Took me a while to debug and figure out where that return to a position was exactly happening


bakatrouble:
Just noticed that perhaps `restore_xy_pos` should be respected here?

moggieuk:
Thanks! I need to think about all the possible corner cases for this carefully.  I've been busy the past couple of weeks but have some time in the next few days...

moggieuk:
This is still on my list of things to review... soon ;-)

leweljun:
I have the same issue and was fixed by this PR
I merged this change with v2.5.2-33-gecb57559 and completed a 49hrs print, too bad I don't have logs.

moggieuk:
Sorry, it took a while (I'm been distracted) and frankly I was being very careful because making sure all combinations of operation-fail, and fail-operation were covered.  Your fix completely makes sense now I reminded myself of the logic.  

Anyway I pulled the PR.  Modified the logic very slightly and cleaned up some unnecessary "gcode_move" lookups that I introduced.

leweljun:
Thanks. This is great news.
Actually the PR was from Bakatrouble, I merely merge and tested the PR on my environment, since I found it to fix my issue.
________________________________

PR #309, opened 2024-06-07
Added `toolhead_ooze_reduction` logic into `_MMU_FORM_TIP` and `_MMU_CUT_TIP` macros, added commits from main as well

NortonTech3D:
After much trial and error the `toolhead_ooze_reduction` value has been successfully incorporated into the `_MMU_FORM_TIP` and `_MMU_CUT_TIP` macros. This prevents the hot tip of the filament from being pulled too far into the extruder, thus preventing clogs

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Controlling an ERCF with companion [customized KlipperScreen](https://github.com
  
 ## ![#f03c15](https://github.com/moggieuk/Happy-Hare/wiki/resources/f03c15.png) ![#c5f015](https://github.com/moggieuk/Happy-Hare/wiki/resources/c5f015.png) ![#1589F0](https://github.com/moggieuk/Happy-Hare/wiki/resources/1589F0.png) Installation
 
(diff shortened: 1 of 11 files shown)


NortonTech3D:
This took way too long to work out haha

moggieuk:
Ok. I'm merging this now with some related but similar logic for toochange retraction

One thing I want to do is not change the semantic meaning of 'retract_length'.  The reason is doc and confusion on the intent of this setting relative to the blade pos.   I.e. the user sets to 0 to `blade_pos`.

Then instead, calculate:
```
{% set effective_retract_length = retract_length - toolhead_ooze_reduction - toolchange_retract  %}
```
And use this distance for the retract.

So the logic is that say, blade pos is 40mm.  The user wants 5mm filament left, they would set retract_length to 35mm.

If we know that HH has retracted 2mm (this is the new stuff I've been working on) and that 10mm of filament is left behind in the nozzle, then the effective retraction length would be 40 - 10 - 2 = 28mm

In theory this would leave the 5mm stub of filament behind.

One reason for this is that your change greatly complicates upgrade. Install.sh reads all old parameters and variables and restores them.  If we change the meaning or name I need to write upgrade logic in the installer and update all docs.

WDYT?

NortonTech3D:
Great points on how it would complicate the install.sh script, I think what you propose here is a great idea.

PR #336, opened 2024-07-08
Updated Blobifier.cfg

smwoodward:
There was a difference between the blobifier.cfg that I had on my machine and the one that was included here.  This update adds temp check and sets temp, and it homes all axes and QGL if not done.

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -504,26 +504,40 @@ gcode:
   {% set pos_max = printer.toolhead.axis_maximum %}
   {% set position_y = pos_max.y %}
 
+
   SET_GCODE_VARIABLE MACRO=BLOBIFIER_PARK VARIABLE=restore_z VALUE={pos.z}
 
   SAVE_GCODE_STATE NAME=blobifier_park_state
+
+  
+  {% if "xy" not in printer.toolhead.homed_axes %}
+    RESPOND MSG="Automatically homing XY"
+    G28 X Y
+  {% endif %}
+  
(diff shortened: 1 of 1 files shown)


Dendrowen (commented):
Homing is not the responsibility of a park movement and temperature setting also isn't. Thanks for the input though!

smwoodward:
> Homing is not the responsibility of a park movement and temperature setting also isn't. Thanks for the input though!

While I do agree, is it better for if someone issues BLOBIFIER_PARK for it to error if homing and temps aren't correct or to would it be better for BLOBIFIER_PARK to fix the deficiencies to complete the intended action of parking the toolhead on blobifier?

moggieuk:
I think the core issue is that Blobifer is not designed to run separate from everything else. It should display suitable error message if some conditions are not met but if you start implementing "if not homed, then home" logic here then why not to add to EVERY macro that moves the toolhead...

Dendrowen:
I agree with @moggieuk, if you wan't to use certain macro's that use the toolhead, then you'll need to home first. That's how the default klipper macro's handle it, and others should do the same to stay consistent.

smwoodward:
Ok, so I think I see what's going on here.  I'm not sure how I got the config file from the original Blobifier and not the cfg file from Happy Hare, and that's why things weren't making sense or working like I thought.  It primarily started with the original cfg had a G1 E line which was causing it to error because it wasn't up to temp.  That line and all of the other check for homing/QGL was also removed from the HH file, but somehow I still had the original from Blobifier Github.  This isn't needed and I will make sure my files are correct.

Issue #272, opened 2024-04-18
MMU_START_LOAD_INITIAL_TOOL doesn't load tool on single color prints


iamnotahippy:
As the title says, MMU_START_LOAD_INITIAL_TOOL won't actually load the tool on a single color print. I'm not sure if this is intended behavior, but the culprit seems to be the line which reads:

``` {% if not using_bypass or tools|length > 1 %}```

Changing the line to:

``` {% if not using_bypass or tools|length >= 1 %}```

seems to fix the issue.

**Comment by moggieuk** [comment]
I think you are correct and this is an oversight. I need to review and remind myself that there isn't a valid reason.... Some folks don't use the bypass and just load the tool/color they want to print with.  The slicer default for `initial_tool` is 0 and so it could result in frustrating behavior of undoing intent, but...

On reflection I do think that loading the initial tool (even on single color prints) might be more logical.

**Comment by moggieuk** [comment]
Fixed. Update pushed.

**Comment by demitrix** [comment]
The call to MMU_START_LOAD_INITIAL_TOOL now gives me this error. I'm not using bypass just a single color print. I can manually change to correct tool with T1 and resume.

MMU issue detected whilst printer is paused
Reason: Load bypass or initial tool before resuming print

**Comment by mpsnp** [comment]
@demitrix do you have referenced tools in gcode? there should be at least one i guess 🤔

**Comment by mpsnp** [comment]
I suppose you're removing/adding each time list of filaments in your slicer. Guess whole intention of TTG is to keep list of filaments static in slicer and change those only in case of swapping spools.
So instead of leaving only one in the list, you'd select right filament from the list for single color prints. 
Or i'm wrong @moggieuk ?

**Comment by demitrix** [comment]
I didn't chance any spools in the my slicer, or gate mapping. The filament on gate 1 is also the second filament in my slicer. There is no tool change during the print so tool 1 should be the tool loaded by MMU_START_LOAD_INITIAL_TOOL, from my understanding.
Relevant g-code from a file
```
[...]

PR #352, opened 2024-07-25
Fix load initial tool in case of using bypass

mpsnp:
Related to #272. MMU shouldn't try to load initial tool in case of using bypass.


moggieuk (approved):

moggieuk:
Thanks!

PR #367, opened 2024-08-02
Updated variable_feed_length for EREC Cutter for ERCF V2

xF4m3:
Updated variable_feed_length for EREC cutter to work with V2 out of the box, which is based on the gate_parking_distance from mmu_parameter.cfg and adjusted the comment to show the values for ERCF V1.1 and V2+ others. This is based on the install.sh gate_parking_distance of 13 (ERCF V2 + others) vs 23(ERCF V1.1).

If not adjusted properly the cut length is not correct. Initial Macro came from me with these Values, since i still had gate_parking_distance of 23 left when upgrading from V1.1 to V2.

### DIFF
--- a/config/addons/mmu_erec_cutter.cfg
+++ b/config/addons/mmu_erec_cutter.cfg
@@ -35,7 +35,7 @@ variable_servo_duration       : 0.4	; Time (s) of PWM pulse to activate servo
 variable_servo_idle_time      : 1.0	; Time (s) to let the servo to reach it's position
 
 # Controls for feed and cut lengths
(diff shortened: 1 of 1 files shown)


xF4m3:
Might be a good idea to couple this value with the install.sh since the gate_parking_distance is set by the installer when asking for V1.1, V2 or other. I just do not know how to do that properly, so i leave that up to you or someone who knows.

moggieuk:
Thanks. Yes I can do this.  I can add to the installer, but how many ERCFv1.1's are still evolving..?

PR #370, opened 2024-08-07
Update regex pattern for OrcaSlicer filament color (fix for single to…

CooperGerman:
Minor fixes to the spoolman integration features which i stumble across while beta testing

### DIFF
--- a/components/mmu_server.py
+++ b/components/mmu_server.py
@@ -721,7 +721,7 @@ METADATA_TOOL_DISCOVERY = "!referenced_tools!"
 COLORS_REGEX = {
     'PrusaSlicer' : r"^;\s*extruder_colour\s*=\s*(#.*;.*)$",
     'SuperSlicer' : r"^;\s*extruder_colour\s*=\s*(#.*;.*)$",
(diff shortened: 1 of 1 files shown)


moggieuk:
Thanks. Would this be the same for prusaslicer and superslicer?  Seems like it would be.  I have prusaslicer so can describe how to recreate a test print job...

CooperGerman:
I would think this might be the case:
To recreate launch a print with only one tool (in slicer) but a mmu setup with multiple gates. 
Automap should be on on whatever setting (i think).

Launching this should result in an error at 
`MMU_START_SETUP INITIAL_TOOL=0 REFERENCED_TOOLS=0 TOOL_COLORS=  TOOL_TEMPS=210 TOOL_MATERIALS=PLA PURGE_VOLUMES=0,60 FILAMENT_NAMES="Winkle PLA-HD Glacier White"`
at the beginning of the pre-processed gcode file (missing color information).

This induces then an error in `MMU_START_SETUP` as the `num_slicer_tools` is calculated with this information

moggieuk:
Indeed it is the same on PrusaSlicer (therefore SuperSlicer).

Want to mark PR and final and I can merge...

PR #382, opened 2024-08-15
Allow using SX1509 and MCP23017 IO expander for buttons

idubrov:
Allow using I2C IO expanders to be used as inputs for buttons / filament sensors / etc.

The implementation relies on an interrupt pin provided by the I2C expander board to detect change in the input pins. Once change is detected, Klipper would read the state of the buttons via I2C and invoke callbacks as needed.

To integrate with the existing buttons support, simulates MCU and handles the commands used to setup buttons (`config_buttons`, `buttons_add`, `buttons_ack`, `buttons_query` and `buttons_state`). The motivation of this approach is that it does not require changes to Klipper code.

The I2C board thus requires three GPIO pins to be connected to any Klipper board: two pins for I2C (SDA and SCL) and one GPIO pin for the interrupt.

Add support for inputs pins to SX1509 implementation. The new implementation uses configuration name `[mmu_sx1509 ...]`.

Also, implement support for MCP23017 I2C expander board. This implementation uses configuration name [`mmu_mcp23017 ...`].

Example configuration for the SX1509:

```
[mmu_sx1509 expander]
i2c_mcu: mmb
i2c_address: 62
i2c_bus: i2c3_PB3_PB4
interrupt_pin: !mmb:PC15
[filament_switch_sensor gate5]
switch_pin: ^mmu_sx1509_expander:PIN_5
pause_on_runout: False
```

Example configuration for the MCP23017:
```
[mmu_mcp23017 expander]
i2c_mcu: mmb
i2c_address: 32
i2c_bus: i2c3_PB3_PB4
interrupt_pin: !mmb:PC15
[filament_switch_sensor gate3]
switch_pin: ^mmu_mcp23017_expander:PIN_3
pause_on_runout: False
```

### DIFF
--- a/klippy/extras/mmu_i2c_inputs.py
+++ b/klippy/extras/mmu_i2c_inputs.py
@@ -1,0 +1,122 @@
+# Simulated I2C Inputs support
+#
+# Copyright (C) 2024  Ivan Dubrov <dubrov.ivan@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import pins
+from . import bus
+
+# Used to ignore acks from the buttons code
+class MmuI2cInputsSendAck():
+    def send(self, data=(), minclock=0, reqclock=0):
+        return
+
+class MmuI2cInputs(object):
+    def __init__(self, config):
+        self._printer = config.get_printer()
+        name = config.get_name().split()[1]
+        i2c = bus.MCU_I2C_from_config(config, default_speed=400000)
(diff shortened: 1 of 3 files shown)


idubrov:
This is draft -- I only did a very basic testing. Just to create awareness & maybe allow others to test.

moggieuk:
This is great!!   It would be good to have folks test and validate.  It would be a nice addition for MMU with lots of gates / switches.

You might have to keep asking for testing help in the forums...

Issue #397, opened 2024-08-23
filament sync tension is never corrected with two sensors set up in some cases


ntchris:
filament sync tension is never corrected with two sensors set up in some cases

Steps to reproduce:
1 printer has expand and compress sensors setup, two,  not one.  ( expand sensor for too little filament, compress sensor for too many filament )
2 manually put more filament in tube to make the compressed sensor triggered.
3 manually set the mmu_hardware.cfg file  rotation_distance to a obvious smaller value , ie 20. ( so ERCF shall push more filament than it should)

Restart raspberry PI controller computer or klipper (  sudo  service klipper restart ), things are correct, (this explains why sometimes it works sometimes it doesn't. )
**however, when we do **firmware restart ** from the webpage (key step: klipper software did not restart)**

**Expected behaviour:**
filament sync state get updated by actual sensor reading --  compressed,  and adjust the rotation_dist

**Actual behaviour**
sync state remains neutral ( should be compressed), 
and if we start printing, since the ERCF will keep pushing more filament than it should (by the smaller rotation_distance value in config) , the already triggered compressed sensor will remain the same, the event _handle_sync_feedback() will never happen. the state will be neutral forever. 
thus the multiplier and gear ratio remains the same.
It's natural that sensor is in compressed state since the rotation distance is pushing too much filament, so , not a very rare and corner case.
 
It should have the same problem in another similar case - rotation_distance is too large , ERCF pushes too little filament, and the sensor is already in expanded state. it shall remain the incorrect neutral state forever.
It's natural that sensor is in expanded state since the rotation distance is pushing too little filament, so , not a very rare and corner case.

**Propose Fix**

read the actual sensor reading and set  **self.sync_feedback_last_state** when software restarts. (firmware reboot)
 def handle_ready(self)  ...seems it couldn't read sensor state at the time it runs.
or  maybe  def _enable_sync_feedback(self):
`    def _enable_sync_feedback(self):

    def _enable_sync_feedback(self):
        if self.sync_feedback_operational: return
        self.sync_feedback_operational = True
        self.reactor.update_timer(self.sync_feedback_timer, self.reactor.NOW)

        eventtime = self.reactor.monotonic()
        state_expanded = self.mmu_sensors.get_status(eventtime)[self.SWITCH_SYNC_FEEDBACK_TENSION]
        state_compression = self.mmu_sensors.get_status(eventtime)[self.SWITCH_SYNC_FEEDBACK_COMPRESSION]
[...]

PR #398, opened 2024-08-24
read tension sensor state when sync feedback is enabled

ntchris:
fixed this issue
https://github.com/moggieuk/Happy-Hare/issues/397

fixed a bug that sometimes tension sensor is never read and the tension state is never updated.

filament sync tension is never corrected with two sensors set up in some cases

**Steps to reproduce:**
1 printer has expand and compress sensors setup, two, not one. ( expand sensor for too little filament, compress sensor for too many filament ) 2 manually put more filament in tube to make the compressed sensor triggered. 3 manually set the mmu_hardware.cfg file rotation_distance to a obvious smaller value , ie 20. ( so ERCF shall push more filament than it should)

Restart raspberry PI controller computer or klipper ( sudo service klipper restart ), things are correct, (this explains why sometimes it works sometimes it doesn't. ) **however, when we do firmware restart ** from the webpage (key step: klipper software did not restart)

Expected behaviour:

filament sync state get updated by actual sensor reading -- compressed, and adjust the rotation_dist

**Actual behaviour**
sync state remains neutral ( should be compressed), and if we start printing, since the ERCF will keep pushing more filament than it should (by the smaller rotation_distance value in config) , the already triggered compressed sensor will remain the same, the event _handle_sync_feedback() will never happen. the state will be neutral forever. thus the multiplier and gear ratio remains the same. It's natural that sensor is in compressed state since the rotation distance is pushing too much filament, so , not a very rare and corner case.

It should have the same problem in another similar case - rotation_distance is too large , ERCF pushes too little filament, and the sensor is already in expanded state. it shall remain the incorrect neutral state forever. It's natural that sensor is in expanded state since the rotation distance is pushing too little filament, so , not a very rare and corner case.

**The Fix**

make it read the actual sensor reading and set self.sync_feedback_last_state when software restarts. (firmware reboot) 
def handle_ready(self) ...seems it  sometimes couldn't read sensor state at the time it runs , both sensors are 0 even it's actually triggered, also there is no browser log at the time this is called/run, seems this is not a perfect place.
instead, let's add it inside  def _enable_sync_feedback(self):

` def _enable_sync_feedback(self):
`
log file shows the incorrect behaviour is now fixed.
 
[fixed_read_sensor_when_enable_sync.log](https://github.com/user-attachments/files/16734873/fixed_read_sensor_when_enable_sync.log)

manually trigger tension sensor in 3 states, expanded, neutral, compressed, run **MMU_SYNC_GEAR_MOTOR** ,
watch log shows the correct state , also the self.sync_feedback_last_state is updated.

Also added the confusing tension state constant definition and text definition for easier future work

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -238,6 +238,23 @@ class Mmu:
     SPOOLMAN_OPTIONS       = [SPOOLMAN_OFF, SPOOLMAN_READONLY, SPOOLMAN_PUSH, SPOOLMAN_PULL]
     SPOOLMAN_CONFIG_ERROR  = "Moonraker/spoolman may not be configured (check moonraker.log)"
 
+    # filament sync tension states
+    class SyncTensionState:
+        SYNC_NEUTRAL = 0
+        SYNC_COMPRESSED = 1
+        SYNC_EXPANDED = -1
+        #  a helper function easily output tension state to log
+        @classmethod
+        def tension_state_to_text(cls, state):
+            if math.isclose(float(state), cls.SYNC_NEUTRAL):
+                return "neutral"
+            elif math.isclose(float(state), cls.SYNC_COMPRESSED):
+                return "compressed"
+            elif math.isclose(float(state), cls.SYNC_EXPANDED):
+                return "expanded"
+            else:
+                return f"unknown ({state=})"
+
     # Automap strategies
     AUTOMAP_NONE           = 'none'
     AUTOMAP_FILAMENT_NAME  = 'filament_name'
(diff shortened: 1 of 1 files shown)


moggieuk:
Firstly, thanks for all your PR's!  Some great additions and fixes.

I have quite a large code refactor due to be pushed to the main branch which addresses some of the bugs you have found (I think).  After than I can run through your PR's although I will likely need you help to refactor them.

One this one in particular, you are correct that this is an oversight, but I'm trying to understand the need to reset sync state every time the feedback is enabled.. you comment that the switches do not always give the correct values?

My intention was only to set the sync state on startup (ready() event) and then always listen for sync updates so the state of the switch should always be valid regardless of wether the sync is currently enabled/disabled.

ntchris:
thanks, seems all fixed and improved.
closing.
also yes, that mmu_var.cfg auto save got me confused for a while, but figured it out later...

PR #400, opened 2024-08-24
fixed a bug that servo up and MMU_SYNC_GEAR_MOTOR it remains unsync

ntchris:
fixed a bug that run servo up and MMU_SYNC_GEAR_MOTOR it remains unsync

moggieuk#399
simply move the get prev_sync_state line after the servo_down block 
because the servo_down function mess up the previous state as the servo_down 
calls this function _sync_gear_to_extruder  itself again

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -5437,12 +5437,13 @@ class Mmu:
         if self.gate_selected < 0:
             sync = current = False
 
-        prev_sync_state = self.mmu_toolhead.is_gear_synced_to_extruder()
         if servo:
             if sync:
                 self._servo_down()
             else:
                 self._servo_auto()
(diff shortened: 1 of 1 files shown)


moggieuk:
Thank again for this fix.  I believe that with the latest updates to the main branch (which is a restructure of the way stepper syncing is performed) the problem you identified has been fixed.  See the use of `with _wrap_sync_mode():` in many places.  The synchronization state should not be preserved.  If you still see a problem let me know.

BTW Since it's obvious that you are doing a thorough review of code I wanted to let you know that I'm working on a substantial restructure to break up what has become a ridiculously large class!  That work is being done in the "development" branch although it's not functional at the moment and is designed to support type-B MMU's like AMS with multiple gear steppers..

ntchris:
I think the fix is already in place.

PR #413, opened 2024-09-01
Support for Y axis cutting motion

drataiczak:
Adds a parameter named cutting_axis in mmu_macro_vars.cfg. Renames several to be axis-independent. Update g code to respect the new cutting_axis variable.

### DIFF
--- a/components/mmu_server.py
+++ b/components/mmu_server.py
@@ -376,8 +376,11 @@ class MmuServer:
         '''
         if self._mmu_backend_enabled():
             gate_dict = {
-                gate: {'spool_id': -1} if spool_id < 0 else (
-                    spool[2].copy() if (spool := self.spool_location.get(spool_id)) else logging.error(f"Spool id {spool_id} requested but not found in spoolman")
+                gate: (
+                    {'spool_id': -1} if spool_id < 0 else
+                    self.spool_location.get(spool_id)[2].copy()
+                    if self.spool_location.get(spool_id)
+                    else logging.error(f"Spool id {spool_id} requested but not found in spoolman")
                 )
                 for gate, spool_id in gate_ids
             }
--- a/config/base/mmu_cut_tip.cfg
+++ b/config/base/mmu_cut_tip.cfg
@@ -45,20 +45,10 @@ gcode:
     {% set cutting_axis = vars['cutting_axis'] %}
 
     {% if cutting_axis == "x" %}
-	{% if printer.configfile.config.stepper_x.homing_positive_dir == True %}
-            {% set pin_park_x_loc = pin_loc_x + pin_park_dist %}
-        {% else %}
-            {%set pin_park_x_loc = pin_loc_x - pin_park_dist %}
-        {% endif %}
-
+	{%set pin_park_x_loc = pin_loc_x - pin_park_dist %}
         {% set pin_park_y_loc = pin_loc_y %}
     {% else %}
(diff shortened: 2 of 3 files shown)


Hobbss27:
Question. how do i implement this?

drataiczak:
> Question. how do i implement this?

As it stands, you have a couple of options, though all of them involve you running an unsupported (by moggieuk) version of the HH macros. You can either SSH into your pi and modify ~/printer_data/config/Happy-Hare/config/base/mmu_cut_tip.cfg to match the modifications I've made here. You can then go into your config through Mainsail (or whatever frontend you're using) and update the mmu_macro_vars.cfg to match the modifications Ive made.

Alternatively you can clone my fork and overwrite your HH install on the pi, but this is more advanced if you are not familiar with how Git works.

Inside of mmu_macro_vars.cfg you set the following:
variable_cutting_axis: 'y' # to cut on y axis
variable_pin_park_dist: 5 # distance behind cutter to wait
variable_pin_loc_compressed: <y coord> # y coordinate where your cutter is fully depressed

Again, none of these options are currently supported my moggieuk, so only do so at your own risk until/if they are pulled into the HH repo.

Hobbss27:
apprciate the answer. i need a Y travel for my cutter vs an X travel. will probably do it the first way you mentioned.

moggieuk:
JFYI. I intend to pull this PR in the next versioned update to HH.  I can't do it without revving version because I need to implement automatic upgrade of old param/variable names, etc.

A version change is always necessary when there are changes to any parameter name in `mmu_parameters` or variable name in `mmu_macro_vars`

moggieuk:
Again, thanks for this work.

This PR has been merged into the "auto-calibrate" branch which represents v2.7.1.  when it is shipped.  This PR will disappear from github once that is merged into main. I added upgrade logic so old variable names will automatically be fixed.

The auto-calibrate branch is still beta, but it is essentially the same as "main" but with new auto-calibration/tuning added. My only concern with the auto calibration is how it works with sync-feedback sensors like Belay (but you can disable with the two new settings in mmu_parameters).

To opt in:
```
cd ~/Happy-Hare
./install.sh -b auto-calibrate
```

You can stay on this branch until it becomes main if you want, but if you need to get back to main BEFORE v271 is released, you must do two steps:
```
./install.sh -b main
```
AND
copy the last backed up mmu_vars.cfg back into place because auto-calibrate also removes all the old "ratio" calibration metrics and replaces with a single list of rotation_distances.

drataiczak:
Thanks for the update moggie. Glad to hear of some of the new changes on the way.

CarolynRemacle:
I need Y axis cutting motion also and am interested in using this.  I just have a question about when you said 'My only concern with the auto calibration is how it works with sync-feedback sensors'.  Can you tell me more about this concern and how it would manifest?

moggieuk:
I have not tested the auto-calibration features with a belay like sync-feedback sensor so I don't know how well they would work.  You can turn off the two auto-calibration options in `mmu_parameters` though.  But this branch will also dissappear when integrated into main so you need to know how when when to switch back if you use now.

moggieuk:
Update on this:
I'm closing in on a v2.7.1 release that will formerly include this feature. It's in the `v271beta` branch.
It also contains lots of updates on the toolchange movement configuration so if you use, be sure to read the doc inside the MOVEMENT section of `mmu_macro_vars.cfg` carefully.

PR #414, opened 2024-09-01
fix mmu_eject and mmu_recovery command

ntchris:
fix mmu_eject and mmu_recovery command make it work with tension sensor free move distance

add capability to check if filament is in extruder with tension sensor device
1. added new function retract_from_extruder_until_mmu_can_freely_pull() to reliably pull filament out of extruder

2. make _unload_sequence() work reliably with tension sensor setup

3. added helper function with_tension_sensor_check_is_filament_in_extruder_by_weak_move() to test filament is in extruder by weak pull

4. modified _test_filament_in_extruder_by_retracting() to make it work with tension sensor setup

5 can test and output a help message to suggest user adjust toolhead_extruder_to_nozzle value in parameter.cfg file like this had to unload extra 11.0mm to unload from extruder, suggest adjust toolhead_extruder_to_nozzle in config

6 added a sanity check if unload from extruder, the unloaded length must not be too far off from the bowden length, so to detect a true unload failure, so to avoid EREC servo cutter to cut in the middle of the filament while it's still loaded in extruder, what a disaster

Test:

* load filament in extruder, can load
* mmu_recover can detect correctly
* load 50mm more filament into extruder, and run  mmu_eject, can eject correctly
* put filament in encoder only, unloaded from extruder, MMU_TEST_MOVE MOVE=50, mmu_recover can recovery correctly
* set a obviously shorter value for toolhead_extruder_to_nozzle * in parameter.cfg file, filament still can be pulled up by MMU_EJECT command

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -4813,6 +4813,17 @@ class Mmu:
                 self.log_debug("Unloading last %.1fmm to exit the extruder%s" % (length, " (synced)" if synced else ""))
                 _,_,measured,delta = self._trace_filament_move("Unloading extruder", -length, speed=speed, motor=motor, wait=True)
 
+                # filament should be out of extruder now, but is it really? worth double checking before try to
+                # unload the whole bowden length in hi speed.
+                UNLOAD_EXTRUDER_MORE_MAX = 100
+                step = 20
+                unloaded, extra_len = self.retract_from_extruder_until_mmu_can_freely_pull(UNLOAD_EXTRUDER_MORE_MAX,
+                                                                                           step)
+                if unloaded:
+                    self.log_debug("unload extruder done!")
+                else:
+                    raise MmuError(f"failed to unload from extruder even with {extra_len}mm more length.")
+
                 # Best guess of filament position is right at extruder entrance or just beyond if synced
                 if synced:
                     self._set_filament_position(-(self.toolhead_extruder_to_nozzle + self.toolhead_unload_safety_margin))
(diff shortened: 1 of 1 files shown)


moggieuk:
There is a lot of logic to parse here. Few things to be aware of:
1. Has to work with python2 because Klipper still supports. It's really frustrating but necessary.  I'd estimate about 10%+ of users
2. Need to make sure this works with all sync_feedback sensors including those that don't have switches but are proportional feedback.  May need a config option.
3. In my "auto-calibration" branch I have logic to auto-tune bowden length and gear rotation_distance for none gate 0.  This relies on a couple of pieces of telemetry data.  including the excessive homing that was necessary and the mismatch of encoder and move through the bowden. Need to think through how this is effected by sync-feedback sensor.

Other thoughts. I haven't studied many devices in detail but they really need a spring that attempts to hold them at a fixed point until pressure becomes too great. This "natural home" would help with consistent move lengths.

PaulMellor:
Just a questions regards "I haven't studied many devices in detail but they really need a spring that attempts to hold them at a fixed point until pressure becomes too great. This "natural home" would help with consistent move lengths."

As I have thought about that myself in regards to the belay and other compression/tension sensors as they do not have a neutral position only a compression and tension position. Is the coding already if a third position in the sensor to add as a neutral? Knowing full well it will add significant length to the sensor(I am currently using a Hall effect switch) so need that to provide a gap between the switchs. Not an expert but can certainly modify to test this out. 
As I think the sensor is beneficial as have a 1200mm Bowden length, and certainly has improved prints as less strain on extruder.

PR #416, opened 2024-09-02
Fix PAUSE called twice

miguelangel-nubla:
Fixes `PAUSE` being called twice and out of order.
First `PAUSE` was called before `_save_toolhead_position_and_lift`, so the toolhead would not move to the correct position on resume.


moggieuk:
Thanks. This somehow got added on a previous merge ..

PR #421, opened 2024-09-12
Allow comments on toolchange commands to accommodate verbose slicer mode

GarthSnyder:
I'm using OrcaSlicer and have the Verbose G-code option turned on. That causes T* commands to be emitted with a comment, and Happy Hare's G-code preprocessor no longer recognizes them. This is a simple regex fix to allow (and ignore) comments.


moggieuk:
Oh, this makes a lot of sense.  Thanks.

PR #424, opened 2024-09-16
Fix auto-setting spool ID in spoolman pull mode.

miguelangel-nubla:
I understand pull mode should also imply pushing relevant local changes, for example auto-setting spool ID on filament insertion.


moggieuk:
I'm going to close out this PR because I think it stems from a misunderstanding that "pull" mode means ONLY pull from spoolman and don't write back changes.   The `MMU_SPOOLMAN` command can be use to update the printer and gate assignment only, other attributes must be edited through spoolman.

PR #427, opened 2024-09-16
Fix ‘out of range’ with Blobifier shake

jake-b:
When using skew correction, a Blobifier shake operation can trigger an ‘out of range’ error.  This fix allows the user to apply a skew correction to both x and y axes.

cc: @Dendrowen 

Thank you for your consideration.

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -96,7 +96,14 @@ variable_wipe_spd_xy:       10000          # Nozzle wipe speed in mm/min.
 # correction is set up. If you have skew correction enabled and get 'move out of range'
 # errors regarding blobifier while skew is enabled, try increasing this value. Keep the 
 # adjustments small though! (0.1mm - 0.5mm) and increase it until it works.
-variable_skew_correction: 0          
+variable_skew_correction_y: 0
+
+# Similarly, Blobifier sends the toolhead to minimum x position when shaking, but this 
+# can cause issues when skew correction is setup. If you have skew correction enabled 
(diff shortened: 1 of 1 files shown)


moggieuk:
I think this is a sensible precaution to add, but I've advocate to simplify setup and rather than separate x and y skew correction factor, instead just have the single setting:
```
variable_skew_correction: .. 
```
and apply it in both the x and y coordinates.

It seems pedantic to have two variables rather than one shared.

@Dendrowen WDYT?

Dendrowen:
Yeas, I agree. sub mm precision isn't important in this case and if you have more than a mm of skew, you've got other problems that need fixing.

moggieuk:
@jake-b could you update PR along these lines, then I an merge. Thanks.

jake-b:
In my skew profile, only x-min was affected, but the added parameter is probably not with the complexity.  This works, and is simpler.  A slight, but unnecessary compensation for the other axis doesn't really change much.

moggieuk:
But the skew could effect the y-min too, right?   So keeping it to the single configuration param just makes it easier to setup and allows shipping a default to cover most users. 

@Dendrowen would it make sense to ship a default skew correction of 0.1mm or 0.2mm.   This isn't enough to effect functionality but will reduce "support" requests..

jake-b:
Happy to add a default skew correction value to the open pull request if that's what's decided.

Dendrowen:
I think that is appropriate. Thanks for the work @jake-b

moggieuk:
Great. Ok. I'll merge this into the v271 release..

PR #438, opened 2024-09-23
Fix pre-processor module to support BambuStudio Slicer

jacksky6:
Modify the preprocessor to support bambu studio

PS:
**BambuStudio Slicer has the following advantages:**
1. The support for multi-color in BambuStudio settings is the default, and no complex settings are required.
2. Supports automatic calculation of flush length.
3. Currently,for users with purge buckets, OrcaSlicer currently has a bug, but BambuStudio has no problem.


moggieuk:
Thanks for this.  I'm happy to add but would really like some validate by others before I merge.  I've been bitten with regexp changes before and I can only validate prusaSlicer.

Anyone reading this, please chime in if you have tested... :pray:

PR #441, opened 2024-09-25
Fix _unload_bowden arguments in _calibrate_bowden_length_auto

ronpandolfi:
I noticed that `mmu_calibrate_bowden` wasn't working. It was loading filament up to the extruder gears, but then backing it away at most 135 mm. Since my bowden tube is ~700mm, this raises an error.

The use of `_unload_bowden` in `_calibrate_bowden_length_auto` was incorrect causing `self._unload_bowden(reference)` to do nothing, followed by `self._unload_gate()` trying to back the filament out (but expecting to go at most 135mm). Adding a `False` as a positional argument fixes this.

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -2668,7 +2668,7 @@ class Mmu:
                     self.log_always("Failed to detect a reliable home position on this attempt")
 
                 self._initialize_filament_position(True)
-                self._unload_bowden(reference)
+                self._unload_bowden(False, length=reference)
                 self._unload_gate()
 
             if successes > 0:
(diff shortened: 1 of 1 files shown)


moggieuk:
You are absolutely correct. No idea how this got messed up - must have been a previous merge problem.  Will pull.

AlexTOOT:
> You are absolutely correct. No idea how this got messed up - must have been a previous merge problem. Will pull.

Running v3.1.0-162-g3d367ce, I have got the exact same issue. 
Loading process is fine, but in backing up to gate, it only unloads only same amount with "gate_homing_max". Don't know why that happens.

> 
$ MMU_CALIBRATE_BOWDEN BOWDEN_LENGTH=1000
22:54:44 
// Calibrating bowden length for gate 0 (automatic method) using ENCODER as gate reference point
22:55:06 
// Filament homed to extruder after 941.3mm movement
// (encoder measured 930.0mm)
22:55:11 
!! MMU issue: Calibration of bowden length on gate 0 failed. Aborting because:
Failed to unload gate because did not clear the encoder after moving 100.0mm

Issue #238, opened 2024-03-21
MMU_TEST_TRACKING raises the servo before stepping 1mm to verify track


ningpj:
This must be an oversight - MMU_TEST_TRACKING raises the servo after the initial 100mm load before stepping 1mm at a time to verify encoder/filament tracking. The servo should be engaged for the duration of this test (Tradrack with Binky encoder & gate sensor, no led effects, no cutter or user macro overrides).

``HH v2.5.1-15-ge6cdbddf``

```

// TRACE: Processing idle_timeout 'printing' event
// TRACE: Initial load into encoder. Stepper: 'gear' moved 70.0mm, encoder measured 21.2mm (delta 48.8mm). Pos: @70.0, (21.2mm)
// TRACE: (mmu_gate sensor detects filament)
// [T0] > (*) >> En >........ [Ex ... Nz] 21.2mm (e:21.2mm)
// DEBUG: Loading bowden tube
// TRACE: Course loading move into bowden. Stepper: 'gear' moved 78.8mm, encoder measured 78.8mm (delta 0.0mm). Pos: @100.0, (100.0mm)
// TRACE: (mmu_gate sensor detects filament)
// [T0] > (*) >> En >>>>>.... [Ex ... Nz] 100.0mm (e:100.0mm)
// Load of 100.0mm filament successful (encoder measured 100.0mm)
// TRACE: Running macro: _MMU_ACTION_CHANGED ACTION='Idle' OLD_ACTION='Loading'
// DEBUG: Setting servo to up (filament released) position at angle: 145
// Gear/Encoder : 01.00 / 00.00 mm +
// Gear/Encoder : 02.00 / 00.00 mm ++
// Gear/Encoder : 03.00 / 00.00 mm +++
// Gear/Encoder : 04.00 / 00.00 mm ++++



// MMU: Happy Hare v2.5.1 running Tradrack v1.0e with 9 gates (OPERATIONAL)
// Servo in UP position, Encoder reads 14592.8mm
// Print state is PRINTING. Selector is HOMED. Tool T0 selected on Gate #0
// Gear stepper is at 100% and is not synced to extruder
//
// Load Sequence:
// - Filament loads into gate by homing a maximum of 70.0mm ('gate_homing_max') to ENCODER
// - Bowden is loaded with a fast 1114.0mm ('calibration_bowden_length') move, then homes to extruder using COLLISION detection (at 40% current)
[...]

PR #442, opened 2024-09-25
Set servo to down during MMU_TEST_TRACKING

wdcocq:
Just a simple fix to ensure the servo is down when starting MMU_TEST_TRACKING

Fixes issue #238


moggieuk:
Thanks. Yes, this is bug that got introduced when I default to servo up at the end of the load sequence.

PR #458, opened 2024-10-03
Fix install.sh about mips

jacksky6:
fix mips install


moggieuk:
Ok, happy to merge this but obviously I can't test it.  It does seem reasonable.

PR #462, opened 2024-10-08
Fix unicode filament names not matching with automap

wdcocq:
Certain unicode characters wouldn't match properly, e.g. the ™ in  "PolyLite™ ABS Black" (this is how it's defined in [SpoolmanDB](https://donkie.github.io/SpoolmanDB/)) would fail to be auto-mapped. I suspect that the methods to transfer it from spoolman or the slicer use slightly different unicode paths.

Also, although the documentation in `mmu_macro_vars.cfg` states it would match case-insensitive, it didn't. So I fixed that too. (The wiki doesn't mention it though)
```
#   'filament_name'  - exactly match on case insensitive filament name
```
I'm far from proficient with python, so possibly there's a better way? Basically I copied the implementation from here: https://docs.python.org/3/howto/unicode.html#comparing-strings

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -1230,8 +1230,8 @@ class Mmu:
         a = unicodedata.normalize('NFKC', a)
         b = unicodedata.normalize('NFKC', b)
         if case_insensitive:
-            a = unicodedata.normalize('NFKC', a.casefold())
-            b = unicodedata.normalize('NFKC', b.casefold())
+            a = unicodedata.normalize('NFKC', a.lower())
+            b = unicodedata.normalize('NFKC', b.lower())
         return a == b
 
     # This retuns the hex color format without leading '#' E.g. ff00e0
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -13,7 +13,7 @@
 # This file may be distributed under the terms of the GNU GPLv3 license.
 #
 import sys # To detect python2 or python3
-import random, logging, logging.handlers, threading, queue, atexit, time, contextlib, math, os.path, re
+import random, logging, logging.handlers, threading, queue, atexit, time, contextlib, math, os.path, re, unicodedata
 from extras.mmu_toolhead import MmuToolHead, MmuHoming
 from extras.mmu_sensors import MmuRunoutHelper
 from extras.homing import Homing, HomingMove
@@ -1224,6 +1224,15 @@ class Mmu:
                 return int(s)
             except ValueError:
                 return s
+
+    # Compare unicode strings with optional case insensitivity
+    def _compare_unicode(self, a, b, case_insensitive=True):
+        a = unicodedata.normalize('NFKC', a)
+        b = unicodedata.normalize('NFKC', b)
+        if case_insensitive:
+            a = unicodedata.normalize('NFKC', a.casefold())
+            b = unicodedata.normalize('NFKC', b.casefold())
+        return a == b
 
     # This retuns the hex color format without leading '#' E.g. ff00e0
     def _color_to_rgb_hex(self, color):
@@ -7395,7 +7404,12 @@ class Mmu:
             # 'standard' exactly matching fields
             if strategy != self.AUTOMAP_CLOSEST_COLOR:
                 for gn, gate_feature in enumerate(search_in):
-                    if tool_to_remap[tool_field] == gate_feature:
+                    # When matching by name normalize possible unicode characters and match case-insensitive
+                    if strategy == self.AUTOMAP_FILAMENT_NAME:
+                        equal = self._compare_unicode(tool_to_remap[tool_field], gate_feature)
+                    else:
+                        equal = tool_to_remap[tool_field] == gate_feature
+                    if equal:
                         remaps.append("T%s --> G%s (%s)" % (tool, gn, gate_feature))
                         self._wrap_gcode_command("MMU_TTG_MAP TOOL=%d GATE=%d QUIET=1" % (tool, gn))
                 if not len(remaps):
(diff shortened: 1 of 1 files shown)


moggieuk:
Thank you for the PR.  You are probably correct in the matching.  I did briefly review the code and the biggest problem I see is that it is not Python 2.7 compliant. Since Klipper still supports 2.7 (I wish they didn't) I have to be very careful about constructs used.  About 10% of users are still on Python 2.  BTW one of the biggest differences between python 2 and python 3 is in unicode character handing.

I'm also cc'ing @CooperGerman   who implemented the original pattern matching code.

One thing that would be helpful is to see if you can update in a way that is still compatible with Python 2.7...

wdcocq:
Ah yes I did not consider that at all. Is it the `casefold()` that's not python2 compatible?  Or something else too? Using `lower()` probably should work just as well and i think that's compatible.

wdcocq:
I changed it to `lower()` I feel like that makes it python 2.7 compatible(?). But I don't have an environment set up to test it properly with Python 2.7 (except for some online compiler)

In my case it still works, and I don't see a reason `lower()` wouldn't work just as well as `casefold()`

moggieuk:
Yes, casefold() was what caught my eye.  We need to confirm this works with python 2.7.  Perhaps extracting as a standalone routine and test.  On line runners would be fine for that. 

I've been burned several times before because I run python 3 and accidentally had regressions as a result.  Would like @CooperGerman  to comment too...

wdcocq:
> Yes, casefold() was what caught my eye. We need to confirm this works with python 2.7. Perhaps extracting as a standalone routine and test. On line runners would be fine for that.

According to this it works in 2.7: https://onecompiler.com/python2/42v2ahp2z

CooperGerman:
Hi, i just made a test for this issue here is the tb and associated results:
```py
#!/usr/bin/env python2.7
# -*- coding: utf-8 -*-
# python2.7 file to test unicode string comparison for equality such as "PolyLite™ ABS Black"

# test equality with string coming from argv  using lower() and re.match()

import sys
import re

string = "PolyLite™ ABS Black"

test_string = sys.argv[1]

# test with lower()
# ./py27_unicode_test.py "PolyLite™ ABS Black"  0.01s user 0.00s system 95% cpu 0.009 total
if string.lower() == test_string.lower():
	print("string.lower() == test_string.lower()")

# test with re.match()
# ./py27_unicode_test.py "PolyLite™ ABS Black"  0.01s user 0.00s system 95% cpu 0.010 total
if re.match(string, test_string, re.IGNORECASE):
	print("re.match(string, test_string)")
```

My python version is `Python 2.7.18`. So both lower() and re.match(X,X, re.IGNORECASE) would work and have very similar execution times.

My thought on this would be to use `re.match` as it seems more "elegant" to me but `lower` seems to be doing the job just fine.

_Note : `# -*- coding: utf-8 -*-` seems to be mandatory in the python file in order to handle such non-ASCII characters._

wdcocq:
> My thought on this would be to use `re.match` as it seems more "elegant" to me but `lower` seems to be doing the job just fine.

I think `lower()` shows the intend clearer, as when regex is involved I usually expect more regex shenanigans to be going on than just a simple string comparison. 
Then again, if we use regex we can actually use regex strings in the slicer names, which could be cool too :) But probably not something we want to support. So we would have to escape our input.

> _Note : `# -*- coding: utf-8 -*-` seems to be mandatory in the python file in order to handle such non-ASCII characters._

I noticed this too in my online example. But this should only be required if you put actual utf-8 string in the source. This works without the  `# -*- coding: utf-8 -*-`

```py
#!/usr/bin/env python2.7
# test equality with strings coming from argv  using lower() and re.match()

import sys, unicodedata, re

test_string = sys.argv[1].decode('utf-8')
match_string = sys.argv[2].decode('utf-8')

def compare_unicode_lower(a, b, case_insensitive=True):
    a = unicodedata.normalize('NFKC', a)
    b = unicodedata.normalize('NFKC', b)
    if case_insensitive:
        a = unicodedata.normalize('NFKC', a.lower())
        b = unicodedata.normalize('NFKC', b.lower())
    return a == b

def compare_unicode_regex(a, b, case_insensitive=True):
    a = unicodedata.normalize('NFKC', a)
    b = unicodedata.normalize('NFKC', b)
    if case_insensitive:
        return re.match(a, b, re.IGNORECASE) is not None
    else:
        return re.match(a, b) is not None

if compare_unicode_lower(test_string, match_string):
    print("lower equal")
else:
    print("lower not equal")

if compare_unicode_regex(match_string, test_string):
    print("re equal")
else:
    print("re not equal")
```

CooperGerman:
> I noticed this too in my online example. But this should only be required if you put actual utf-8 strings in the source. This works without the `# -*- coding: utf-8 -*-`

Indeed ! 

Let's run with lower then, this looks ok for me. I understand using regex for this much might be overkill ;)

@moggieuk : we can go with `lower` and `unicodedata` package ! Worked for me with python2.7 



> Then again, if we use regex we can actually use regex strings in the slicer names, which could be cool too :) But probably not something we want to support. So we would have to escape our input.

Could be cool indeed but I'm not sure onthe real world applications this could have (at least in my scenario). To keep in mind though.

wdcocq:
> Could be cool indeed but I'm not sure about the real world applications this could have (at least in my scenario). To keep in mind though.

The use cases I see would be basically the same as the material and color matches already do now, but you can get more elaborate with it.  Like making groups of compatible filaments that you give an group name and reference that in the slicer regex.  But we should see how far the current implementation gets us before we start over engineering this. And of course supporting regex in user-orientated interfaces is asking for lots of unwanted support issues :sweat_smile: 

A better solution would probably be to have a placeholder system akin of how  Spoolman does it for it's labels. e.g: `{filament.brand} - {filament.name}: {filament.color}`

But hey I'm getting off topic here! let's get this merged so I can get off my own branch.
Thanks everybody!

PR #464, opened 2024-10-14
Blobifier V1.5

Dendrowen:
## version 1.5
- Purge with a pulsating (and occasionally, retracting) motion. This will prevent laminar flow from occuring and purge the old filament more quicly. (Up to 30% EXTRA filament saved!)
- Changed the blob calculation mode. The blob will now split into equal parts if it is too large (e.g. 2x85mm instead of 150mm + 20mm). This will eliminate/reduce the small blobs scattering over the buildplate.
- Due to the above changes, certain variables have been added and removed:
  - new:
    - `z_raise`: The total amount the nozzle should be raised during a blob. The value should be somewhere around `iteration_z_raise * max_iterations_per_blob - triangular(iteration) * iteration_z_change`
    - `z_raise_exp`: The rate at which the hotend reduces raising speed during a blob. `0.85` seems to be a good starting value.
    - `purge_length_maximum`: The maximum length of the entire purge. Use `max_iterations_per_blob * max_iteration_length`.
  - removed:
    - `iteration_z_raise`
    - `max_iterations_per_blob`
    - `iteration_z_change`
    - `max_iteration_length`
- Revised some parameter checks to work with the new Happy Hare version 2.7
- Add settings:
  - `shaker_arm_z` to allow for different bed/shaker arm heights.
  - `brush_accell` to be able to change the acceleration of the brush movement.
- Keep in mind I do not maintain the `SW Only Blobifier.cfg`

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -120,7 +120,12 @@ variable_brush_start:           67  # For 300mm build
 #variable_brush_start:          84  # for 350mm build
 
 # width of the brush
-variable_brush_width:          35
+variable_brush_width: 35
+
+# The acceleration to use when using the brush action. If set to 0, it uses the already 
+# set acceleration. However, in some cases this is not desirable for the last motion 
+# could be an 'outer contour' acceleration which is usually lower.
+variable_brush_accel: 0
 
 # Location of where to purge. The tray is 15mm in length, so if you assemble it against 
 # the side of the bed (default), 10mm is a good location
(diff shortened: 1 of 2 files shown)


moggieuk:
It would be great to get a couple of folks to validate this...

I will need to rev the HH version number to force re-running of the installer and so I don't want to do this more often than necessary..

Dendrowen:
Maybe merge it to a v2.7.3 branch? Would that be better?

moggieuk:
Ok, branch "blobifier_v15" created with v2.7.3 version.

Your PR will disappear when this is merged into main.

moggieuk:
@Dendrowen one thing I've been meaning to alert you too is a new printer variable that appeared in v2.7.1 on...
```
During a toolchange the printer variable printer.mmu.toolchange_purge_volume will contain the suggested TOTAL purge volume comprised of the sum of components:
purge_volume_from_matrix + residual_filament + cut_tip_fragment
This is recommended to be read by any purging logic
```
I didn't check blobifer but this would the idea place to get the starting volume for purge.  Note that with the future v3.0.0 release there is also a command `MMU_CALC_PURGE_VOLUMES` that sets the purge volume matrix based on filament colors.  This is similar to the advanced logic in some slicers but brings it into HH so that it can be used in the absence of slicer purge_map or off the gatemap colors that can be synced with spoolman...

Long way of saying that simply pulling this one printer variable takes into account all the user controls for purge volume.

Dendrowen:
Oh, that's awesome. I'll add it to this merge. 

Would it then also be a good idea to include the purge multiplier in happy hare? That way the 'purge_matrix' would represent the actual volumes.

moggieuk:
Sorry for the delay, been really busy.  Yes a multiplier in HH would make a lot of sense.   This can be supplied to the new "CALC_PURGE_VOLUME" calculator but that is independent from the purge matrix that previously existed.    Maybe an option to the `MMU_SLICER_TOOL_MAP` to scale the purge volumes..   If a parameter than we would need to remove from CALC_PURGE_VOLUME and apply at runout for all purge_volumes...

moggieuk:
@Dendrowen WDYT? Ready to merge this change?

Dendrowen:
I haven't received any complaints, but I'm not sure it has been tested. I think it's ok though

moggieuk:
Ok. I put out a "version update" message tomorrow...

Issue #468, opened 2024-10-18
Incomplete filament position recovery in setups with Extruder Entry sensor


smb209:
My MMU is configured with an extruder entry sensor but no toolhead sensor. I had several restarts where the filament was past the entry sensor but not yet engaged with the extruder gears. Despite the filament detected in the entry sensor, the EREC_CUTTER_ACTION (`variable_user_post_unload_extension`) was performed while the filament was still (partially) loaded.

moggieuk:
Unfortunately I can't help without an `mmu.log` file attached.  I assume that the MMU "state" is out of sync.  What is the status when in this situation?  Did you try "MMU_RECOVER", etc..

moggieuk:
I just saw your PR ... I will look at this, makes sense!

moggieuk:
Closing. PR pulled and incorporated.

Heads up: In v3.0.0 release (currently the v3alpha branch) I reworked the recovery logic. I'm pretty sure it will match your expectations but something to watch for when you get to that release...

PR #469, opened 2024-10-18
Check Extruder Entry sensor during filament position recovery

smb209:
Update the logic in `_recover_filament_pos` to include a check for the extruder entry endstop. This should help catch cases where a restart left filament past the entry sensor but not yet engaged with the extruder. May also help for devices without a toolhead sensor.

For conditions that would set the position to `FILAMENT_POS_IN_BOWDEN`, now perform a check on the extruder entry endstop and if it is installed and triggered, set the position to `FILAMENT_POS_EXTRUDER_ENTRY` instead of `FILAMENT_POS_IN_BOWDEN`.

Addresses issue #468.

### DIFF
--- a/extras/mmu.py
+++ b/extras/mmu.py
@@ -5284,7 +5284,7 @@ class Mmu:
                     else:
                         es = self._check_sensor(self.ENDSTOP_EXTRUDER_ENTRY)
                         if es is not None and es: # Installed and triggered
-                            self._set_filament_pos_state(self.FILAMENT_POS_EXTRUDER_ENTRY, silent=silent)
+                            self._set_filament_pos_state(self.FILAMENT_POS_HOMED_ENTRY, silent=silent)
                         else:
                             self._set_filament_pos_state(self.FILAMENT_POS_IN_BOWDEN, silent=silent) # This prevents fast unload move
                 else:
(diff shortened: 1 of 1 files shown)


moggieuk:
Thanks for this.  This looks basically ok, except that the setting to `FILAMENT_POS_EXTRUDER_ENTRY` isn't a correct assumption.   I think all you can infer by the entry sensor bring triggered without further tests  is the `FILAMENT_POS_HOMED_ENTRY` position.  That said all of these are possible:
 
```
    FILAMENT_POS_END_BOWDEN
    FILAMENT_POS_HOMED_ENTRY
    FILAMENT_POS_HOMED_EXTRUDER 
```

The only practical difference is that the one you used (`FILAMENT_POS_EXTRUDER_ENTRY`)will force an unload_extruder() operation before the fast bowden unload.
Can you confirm if in your situation the filament is still caught in the extruder? If not, I think `FILAMENT_POS_HOMED_ENTRY` is more accurate.

Here is an improved explanation of states:
```
    FILAMENT_POS_UNKNOWN = -1
    FILAMENT_POS_UNLOADED = 0       # Parked in gate
    FILAMENT_POS_HOMED_GATE = 1     # Homed at either gate or post-gate sensor (currently assumed mutually exclusive sensors)
    FILAMENT_POS_START_BOWDEN = 2   # Point of fast load portion
    FILAMENT_POS_IN_BOWDEN = 3      # Some unknown position in the bowden
    FILAMENT_POS_END_BOWDEN = 4     # End of fast load portion
    FILAMENT_POS_HOMED_ENTRY = 5    # Homed at entry sensor
    FILAMENT_POS_HOMED_EXTRUDER = 6 # Collision homing case at extruder gear entry
    FILAMENT_POS_EXTRUDER_ENTRY = 7 # Past extruder gear entry
    FILAMENT_POS_HOMED_TS = 8       # Homed at toolhead sensor
    FILAMENT_POS_IN_EXTRUDER = 9    # In extruder past toolhead sensor
    FILAMENT_POS_LOADED = 10        # Homed to nozzle
    ```

smb209:
Thanks for the clarification. Yes I believe `FILAMENT_POS_HOMED_ENTRY` would be the correct assumption in this case.

PR #472, opened 2024-10-19
Update installer (development)

ammmze:
There's a few updates included here...

- Updates the installer with the following:
  - The `prompt_123` now allows numbers greater than 9 and has additional validations and feedback to the user when the value is invalid.
  - Add a `prompt_option` function to make menu selection more similar to the bash built-in `select`, while still allowing prompts with less than 10 options to function without having to press enter/return. Similar to the `select` command we have the ability to get the numerical response or the text response (which may be useful if we decide to do some dynamic menus where numerical values may vary)
  - For the prompt option/menu things we can define a key for the options and use that as a variable. This removes the need to rely on the option order in the case statement that handles the selected option, instead relying on the defined option key. This should make it easier to build dynamic menu options if we wanted and then have one case after the menu to handle all the possible options. For example, we may choose to only show a subset of the boards when the mmu vendor is armored turtle (because several of the boards do not have enough connections for a box turtle).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -745,7 +745,7 @@ copy_config_files() {
                 sed "/^# ADDITIONAL FILAMENT DRIVE/,+10 d" ${dest} > ${dest}.tmp && mv ${dest}.tmp ${dest}
             fi
 
-        # Conifguration parameters -----------------------------------------------------------
+        # Configuration parameters -----------------------------------------------------------
         elif [ "${file}" == "mmu_parameters.cfg" ]; then
             update_copy_file "$src" "$dest" "" "_param_"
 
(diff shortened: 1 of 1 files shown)


ammmze:
Pulling this back to draft, it may need some updates after the latest refactoring.

moggieuk:
Thanks for this.  Would you mind fixing the merge with the latest install.sh.  Probably better if you did that rather than me.

install.sh
```diff
@@ -1067,17 +1097,18 @@ questionaire() {
...
+        'ERCF v2.0'
+        'Tradrack v1.0'
+        'Angry Beaver v1.0'
+        'Armored Turtle v1.0'
+        '3MS (Modular Multi Material System) v1.0'
+        'Other / Custom (or just want starter config files)'
+    )
+    prompt_option opt 'MMU Type' "${options[@]}"
+    case $opt in
+        *'ERCF v1.1'*)
```
ammmze, line 1111: I'm on the fence between matching the text vs the number. Using this `prompt_option` we can do either one. If we want to match by number we would change the case to use `$REPLY` instead of `$opt`.

```shell
case $REPLY in
    1)
        # ...
```

Matching by text makes it less reliant order which could be useful for dynamic option lists but potentially prone to typo errors. Matching by item number is more reliant on the order of the options. Any preference one way or another? Happy to update these if we want to match by number instead...just let me know what you'd prefer.

PR #478, opened 2024-10-23
fix: use 0 as position for dummy rail

ammmze:
This fixes PRELOAD functionality with type-b mmu using `gate_homing_endstop: mmu_post_gate`


moggieuk (commented):
Thanks. Not sure why my original logic was the way is was!

Issue #340, opened 2024-07-11
Endless Spool function change


platinumalloy01:
I have a suggestion for the endless spool function in happy hare on the ercf.
Rather than unload the filament once a runout has been detected, I recommend that it instead passes a specified amount of filament through the encoder/selector while continuing to print, then after that specified amount has passed through signifying that the gate is cleared and the selector is able to move to another gate loading the next roll of filament, and using a sensor less homing type function that detects when it pushes against the currently loaded filament then continues the print as normal, and not wasting filament, or pausing the print. Thank you for reading, I understand if this may be a limitation of the host computer, but maybe there’s a way.

moggieuk:
This has been asked quite a few times before.  The problem is that it could work (see not super easy) ONLY with single color endless spool prints...  once the filament break is past the MMU gear, there is no way to color change. Also, if the print happens to end with the break still in the bowden it also becomes a hassle.

Frankly, loosing 600mm - 1.5m of filament seems like a better compromise than the extra complexity trying to special case this would bring.  That said, the application of this idea to single color prints is on my "might do" list but near the bottom :-)

platinumalloy01:
Thank you for responding, I figured it might not be feasible at this time, otherwise it would have been implemented already. Only reason I even thought about it was because of an issue I came across where the filament was unable to backfeed because it was getting trapped in the filament block at the buffer system, ended up being the Bowden but yeah, anyways issue resolved. Thanks again for the response and explanation.

moggieuk:
You are right about the trapping in the buffer system .. some are more prone than others. Filamentalist will work quite reliably I think.  The default ERCT can have issues.   I know KinematicDigit is working on some improvements.  Also, the v2.6.0 of HH includes a designated "waste gate" option so one gate can be used to dump filament fragment.  This is experimental and suggested by KinematicDigit, but the downside is that the pre-gate sensors are ignored for runout because the filament must pass clear through the selector so it can select the waste gate...

k1-801:
I know this was already marked as "wontfix", but I think there is a way to improve the EndlessSpool feature for the scenarios where it would be possible to continue printing with what is left in the filament channel.

The idea is to extend the gcode preprocessing in order to add a parameter to __every__ Tx macro that would specify the length of filament that will be extruded in the specified toolchange (i.e., before the next Tx macro or before the end of file). After the toolchange, we reset the counter to whatever length was specified, and, if we encounter a runout, we compare what is left with the length of our bowden tube.

[...]

PR #491, opened 2024-10-30
Extrude Length Preprocessing

k1-801:
This change aims to provide a basis for Endless Spool functionality improvement.
It scans each block of GCode between toolchanges and records the total length of filament used within each block. This data is then attached to the `MMU_CHANGE_TOOL` macro (in the `EXTRUDE_LENGTH=...` parameter).

Additionally, this change removes buffering of GCode lines between the toolchange macro and the next G1 move. This potentially can reduce RAM consumption. This is achieved by recording the file position after the Tx macro and then rewinding the file stream to that point.


moggieuk:
This is an interesting idea.  What kind of overhead does it add to the parsing?  I persuaded the moonraker crew to accept my PR for specifying the timeout of the pre-processing step so at least we can bump that up.

Can you give me the use case of how and when you would use this or is this a precursor to something else?

k1-801:
> This is an interesting idea. What kind of overhead does it add to the parsing?

The parsing is currently pretty slow =\
But I'm working on it, I know what is taking the longest and I have a few ideas that could help avoid it.

> Can you give me the use case of how and when you would use this or is this a precursor to something else?

This is indeed a precursor to the Endless Spool feature upgrade. The idea was described in my comment in #340, but in short - for it to work, it needs to know in advance how much more plastic is planned to be extruded, and if the amount is more than what we have left in the bowden, we follow the already loaded bit of plastic with the next spool in sequence without pausing the print.

k1-801:
> persuaded the moonraker crew to accept my PR for specifying the timeout of the pre-processing step so at least we can bump that up.

I don't think long preprocessing is a good idea in general. I'd look into optimizing it instead.
For example, the placeholders for "MMU_START_SETUP" (`!referenced_tools!` etc.) - those don't have to be substituted in _every_ line, so substituting them everywhere may be a bad idea. I'd say, a good marker that we no longer need to look for those macros is extruding moves, as we definitely don't expect them AFTER the printer starts printing.
So, if at least one extruding move was found, I suggest we stop looking for those placeholders.

k1-801:
Made the MMU-related parsing into it's own class. Hid some of the parameters into object attributes. This should reduce some overhead.

moggieuk:
Any performance optimization of current parsing would be appreciated but it is actually quite fast.  Once a placeholder is found it won't look in subsequent lines, but the logic parses to the end of file to do the NEXT_POS insertions.

JFYI I'm not going to be able to spend time looking at this for some time (i know it is still in development).  The reason is that I'm really busy getting type-B MMU support polished (in the v3alpha branch)

k1-801:
> The approach you are taking is certainly an option but what about thinking about it another way (this assumes a extruder entry sensor):

In the absence of preprocessing, be it disabled or just had it failed, AND with the presence of an extruder entry sensor, this is certainly how this would work. However, there are plenty of machines that don't have an extruder entry sensor, so without preprocessing there isn't much we could do.

Furthermore, there are instances where it woud be wasteful to go on with the print after a runout trigger on MMU, as it would lead to a purge of a whole meter of filament - it would be much better to eject it (if possible), change the spool and __maybe__ use the ejected bit the next time this tool is requested.

I think we are going to need a flowchart :)

moggieuk:
Flowchart, ha ha. Yes it is a complex challenge.  The statement above _"it would be much better to eject it (if possible), change the spool and maybe use the ejected bit the next time this tool is requested."_  can't work though ... if you are past the point of no return you cannot eject (because the filament cannot be grabbed be the MMU gear stepper) thus can either continue as far as possible and if the remaining filament is beyond a user defined threshold, say 300mm, then PAUSE and wait for user interaction. else do the very long purge.  You can't eject a filament that is too short...

k1-801:
> because the filament cannot be grabbed be the MMU gear stepper

But there is some more filament between the extruder gears and the nozzle, I always assumed that unloading from the extruder  makes the filament reach the MMU gears again. This assumption, however, also heavily relies on the MMU sensor to gears distance being significantly shorter than extruder gears to nozzle distance...

k1-801:
>  if you are past the point of no return you cannot eject

This gave me another thought. What if we used the pre-gate sensor, when available? That would notify the system about a runout in advance, thus, allowing to run checks earlier

moggieuk:
>  I always assumed that unloading from the extruder makes the filament reach the MMU gears again. This assumption, however, also heavily relies on the MMU sensor to gears distance being significantly shorter than extruder gears to nozzle distance...
Correct, and this is the case today ... BEFORE passing the point of no return then filament fragment is unloaded and another (endless spool) is substituted.

My point is that often users will have over 1m of filament between that "point of no return" and the extruder.  So, once past, you have no choice but a massive purge IF a toolchange occurs.

moggieuk:
> > if you are past the point of no return you cannot eject
> 
> This gave me another thought. What if we used the pre-gate sensor, when available? That would notify the system about a runout in advance, thus, allowing to run checks earlier

Indeed the pre-gate sensor IS used today for runout unless the user explicitly turns it off.   If can be turned off by disabling the mmu_pre_gate sensor in Mainsail OR by configuring a "waste gate".  This functionality was added at the request of the community so the waste filament can be dumped out a single gate -- that obviously requires the filament to clear the selector before triggering runout.

moggieuk:
I'm going to close this .. no activity and will not work with current branch.

PR #492, opened 2024-10-30
Refactor install script

wdcocq:
This is still work in progress but I wanted to make a PR/Draft to have a place to frame questions.
The basic structure is there, but it probably needs some more fine tuning and quite a bit of testing.

Currently does the following:
- `make menuconfig` creates a `.config` from the Kconfig file. which will provide most of the parameter that should be consistent with the requested options. But there are still some inconsistencies here and there that need to be fixed
- The `Makefile` takes care of most of the copying, backups, installing and checking of file/path existence
- Uses an intermediate folder (`out` by default) to process all the updates in. So a botched install shouldn't affect the Klipper setup.
- `make install` will build and only if the build is successful install all the necessary files
- Moved all the pin defs to `pins.kconfig` so they can be configured through `menuconfig`. 
- `install.sh` mostly only bothers with updating of variables/parameters in the `.cfg` files

 

For now I have the following questions:
- How married are we to do an automatically git pull when installing? currently it has to be called manually with `make update`. 
- Do the B-Type MMU's only ever support MMB MCU's? I only see these supported currently in pins.def. If so I can encode this better in Kconfig
- Will Octoprint/repetier/klippain be more 1st class citizens in the future? I'm not too familiar with any of these so I probably have a lot of gaps here.

Probably a few more to come...


moggieuk:
Great.  Just a heads up that the installer has been updated again..  changes should slow down now and be targeting subtle setup differences for the different vendor machines.

moggieuk:
Answer to questions:
1. The automatic git pull has been a HUGE help in reducing support.  It forces an update to the latest regardless of Mainsail update manager slowness in refreshing its cache.
2. Type-B MMU's can be run on most / all of the boards I just haven't created pin mappings yet.  Some of the new Type-B MMU designs have simple two channel so the older mcus (like EASY-BRD) could still be used.
3. Not sure.  Octoprint is supported by HH (and there is nothing special to setup anymore).  Klippain has it's own setup but the separation of pin aliases into `mmu.cfg` was original to help Klippain.

wdcocq:
> Answer to questions:
> 
>     1. The automatic git pull (with -z flag to opt out) has been a HUGE help in reducing support (as has the -b  option).  It forces an update to the latest regardless of Mainsail update manager slowness in refreshing its cache.

Ok. then I will make it a prerequisite of the `install` target. 

>     2. Type-B MMU's can be run on most / all of the boards I just haven't created pin mappings yet.  Some of the new Type-B MMU designs have simple two channel so the older mcus (like EASY-BRD) could still be used.

I thought so but I wasn't to familiar with them. I'll leave it as is then

wdcocq:
I see in the dev branch's install script you use the following values for `gate_homing_endstop`: encoder, mmu_gate, gate, post-gate (and in `mmu_parameters.cfg` doc: mmu_post_gate)

see: https://github.com/moggieuk/Happy-Hare/blob/05b4f65af230bf53f427ca98e3515caaa133dd0c/install.sh#L1308
and: https://github.com/moggieuk/Happy-Hare/blob/05b4f65af230bf53f427ca98e3515caaa133dd0c/install.sh#L1365
I presume the `mmu_gate` and `mmu_post_gate` are old names, and it should be just `gate` and `post_gate`?

wdcocq:
Currently values from the existing installation are paramount to the values in the `.config` file.
I can see that this can potentially be confusing to some users..  Where they change some settings in menuconfig but they won't actually get used. because those values will be overridden by the existing installation.

The installer does mention during installation which parameters get reused. If they don't want that they can do a `make uninstall` first and get the default install with their menuconfig settings

I think this is the best solution, but I can imagine people wanting the .config/menuconfig to be their leading configuration

Just something we probably should mention in the wiki.

moggieuk:
Sorry, you can see I've been busy with HH v3.0 beta launch (same as development branch now). But development on v3 has slowed now...

Gate endstops:
`mmu_gate`
`mmu_gear` (was briefly called mmu_post_gate). 
Are the correct names.  If "gate" is reference in the install script it must be a bug..

I need to catch up with your work, but the intent in the current installer was to:
- read "default" config into memory
- apply "current" config over the top
- apply some custom upgrade logic like name changes
- write the new template files with now upgraded values
- (add in a few optional params like `cad_*`
- report on any parameter not used

That last step was important because it allows me to remove parameters.  E.g. in v3.0 the "type-B" mmus like Box Turtle don't require the selector or servo configuration parameters.  So if they don't appear in the starting template they won't appear in the final upgraded config.

wdcocq:
> Sorry, you can see I've been busy with HH v3.0 beta launch (same as development branch now). But development on v3 has slowed now...

That's ok :) I'll try to get up to par with the dev branch again soon.

> 
> Gate endstops: `mmu_gate` `mmu_gear` (was briefly called mmu_post_gate). Are the correct names. If "gate" is reference in the install script it must be a bug..

Ok, I'll check whether i use the right ones and fix it

> 
> I need to catch up with your work, but the intent in the current installer was to:
> 
>     * read "default" config into memory
My implementation doesn't read it into memory anymore, instead it uses the `*.cfg` files as the base, so the default values are used unless overwritten later. And replaces the `{placeholders}` with the values from `.config` 

>     * apply "current" config over the top
This is still intact and happens after the the above steps

>     * apply some custom upgrade logic like name changes
I've split this into separate functions (see `upgrades.sh`) per upgrade path. e.g. `upgrade_2_72_to_3_00()` to keep the main script cleaner and smaller. And to prevent collisions with different versions (though this would probably rarely happen) and do upgrades in steps. i.e. 2.70 -> 2.71 -> 2.72 -> 3.00, etc. 

>     * write the new template files with now upgraded values
It now copies the build files after everything has finished instead of writing/updating in place. So a botched upgrade wouldn't leave your klipper configuration in a in-between state (though there are of course still backups)

>     * (add in a few optional params like `cad_*`
This is currently unchanged but this might change if i feel like there is a better more generic way to handle this.

>     * report on any parameter not used
> 
> That last step was important because it allows me to remove parameters. E.g. in v3.0 the "type-B" mmus like Box Turtle don't require the selector or servo configuration parameters. So if they don't appear in the starting template they won't appear in the final upgraded config.

I still have to look more into the new type-B stuff and I saw you've added `mmu_parameters.cfg.vs` for this too. I haven't incorporated this yet (what does the 'vs' stand for?)

I hope to get to a place where `build.sh` would basically be static and only changes to `Kconfig` and `upgrade.sh` would be required for new updates. And have a well-defined install/upgrade path.


To recap it works like this now:

- Copy base files from `config/*.cfg` to `out/config/*.cfg` (Makefile)
- Copy existing `printer.cfg` and `moonraker.conf` to `out/config/...` (Makefile)
- Create links from `components/*.py` and `extras/*.py` to `out/...` (Makefile)
- Read `{placeholder}` values from `.config` and apply to `out/config/*.cfg` where needed (build.sh)
- Read values from existing klipper/happy-hare configuration if it exists (build.sh)
- Apply upgrades to values depending on the version (build.sh/upgrades.sh)
- Apply existing (upgraded) values to `out/config/*.cfg` (build.sh)
- Add the extra parameters to `out/config/*.cfg` (build.sh)
- If all was successful, create backups of `printer.cfg` (if required), `moonraker.conf` (if required) and `.../klipper/config/mmu` (always) (Makefile)
- Copy all the files/links from `out/` to the correct locations (Makefile)
- Restart necessary services
- Done. Happy Hare!

wdcocq:
Ok this was a way deeper rabbit (hare?) hole than I expected.. Mostly because the dev branch added way more types/boards :sweat_smile: So the Kconfig/menuconfig part exploded a bit, hence why I restructured it in the later commits.

Now, I still have quite a bit of testing to do, but I feel the refactoring is kinda where I want it to be for this PR. Though I'll probably still clean some stuff up a long the way, I don't expect anything major.

So I feel now would be a good time to reflect whether this is going in a direction that seems feasible/wanted. 
I know I've added a lot of new files...  I hope it adds to the readability and extensibility of new configurations by splitting them up in separate files per type, board, addon, etc. And easier to configure with the new menu.

scripts/addons/Kconfig.espooler
ammmze, line 1: Any chance we can rename this addon to be `dc_espooler`?  Originally when adding the addon we had it named respooler, then espooler, then finally dc_espooler. The intent there is that it could certainly be possible for someone to come in later an add a stepper driven espooler and adding the `dc_` prefix allows them to co-exist as different addons without giving preferential treatment to one or the other.

wdcocq:
Is there, besides the install script, another reason why HH uses x.xx versioning internally though showing semver x.x.x versions? 
If not maybe this would be the right moment to pick one and stay with that? :)

ammmze (commented):
I want spend more time reviewing this when I've got some availability, but at first glance I think this is looking good. I've never dug too deep into implementing Kconfig, but I know it's been around a long time and is a proven way to do menu config. I also like the explicit upgrade migrations.

ammmze:
> * `make install` will build and only if the build is successful install all the necessary files

It would be cool if we also added a `make diff` target that would build the new config and then print a unified, colorized diff of the changes.

.gitignore
```diff
@@ -1 +1,3 @@
-my_*.cfg
\ No newline at end of file
+my_*.cfg
+out/
+.config
```
ammmze, line 3: Perhaps we should also add ignores for `*.out/` and `*.config`? That way I can easily use other paths without them "dirtying" the repo. I've currently got a couple mmu type units that i switch between, so being able to have `bt.out` and `bt.config` for my box turtle and `tr.out` and `tr.config` for my tradrack would be nice. I do this for building klipper on my mcu's so I don't have to re-configure the mcu each time I build and flash.

ammmze:
> Though I didn't add it as a make target, I did already have a script that crudely does this (though no colors). So at least a basic version wouldn't be a lot of work

👍🏻 Potentially we'd be able to just use the `diff` command. Not sure if it's installed by default on most distro's. My pi running mainsail os has it. Not sure if it was installed originally or got installed as a side effect of other things i've installed though. I doubt it would be installed on creality k1 series...they don't even have `bash` by default apparently...only `sh`.

wdcocq:
> doubt it would be installed on creality k1 series...they don't even have `bash` by default apparently...only `sh`

Wait is this really true? the old install.sh requires bash, no? How did this work then.. Did people manually change the shebang and bash test and hoped for the best? :sweat_smile:

wdcocq:
That being said, even if `diff` doesn't exist on creality K1. I don't think that's a problem as it's not a crucial part of installing/using HH. So just printing a message if it doesn't should be enough.

wdcocq:
One thing I was still debating with myself whether it's worth it is the following:

Currently `copy_config_files()` does a bit of deleting, adding, (un)commenting of sections in the `*.cfg` files. This isn't really my preference. As I see this function becoming very large and unwieldy when adding more configurations.
I think a better approach would be to split certain parts of the config files in separate files and merge them depending on the config. 

e.g:
Put everything for selectors/servos in `mmu_parameters.cfg.servos` and only add it to the main `mmu_parameters.cfg` when required. Probably by replacing a placeholder in `mmu_parameters.cfg` This could be done for things like LED support, encoder, etc. too

Basically this would be a more general approach to what `mmu_parameters.cfg.vs` (A-Type vs B-Type) is currently doing. But would have the benefit of not having duplicates of the same parameters that need be kept in sync in each file.

But so far I've tried not too stray too far from the dev branch's `*.cfg` files. To keep staying up to track easier.

ammmze:
> Wait is this really true? the old install.sh requires bash, no? How did this work then.. Did people manually change the shebang and bash test and hoped for the best? 😅 or I guess install bash..

I'm just going off the issue I helped a person with in the HH discord in the 3ms channel (starting [here](https://discord.com/channels/1305204275315474452/1305204466374672474/1305605991286702200))

ammmze:
> Currently `copy_config_files()` does a bit of deleting, adding, (un)commenting of sections in the `*.cfg` files. This isn't really my preference. As I see this function becoming very large and unwieldy when adding more configurations.
> I think a better approach would be to split certain parts of the config files in separate files and merge them depending on the config.

💯 I've was thinking the same thing as we were adding support for type-b mmus.

wdcocq:
I split parts of the cfg into seperate components and only insert them when required. Imho this makes the intend of `copy_config_files` much clearer and less error prone.

ammmze:
> Hmm I see.. recon this would come up often? I have no clue how many people have a K1 with a mmu..
> In the end it might have been a better choice to make a python builder

Yea...i'm not sure either. But realistically if people are getting in there to run HH, they _probably_ had to flash klipper to the MCU on the board they are using for the MMU. And to do that you would have needed `make` installed anyways. I think as long as we know that they _can_ install `bash` (and other things needed), then we should be good to go. Potentially we could check to see if bash is installed and if not then ask them to install it (or maybe even ask them if they want us to install it...but that potentially gets annoying because of different package managers).

wdcocq:
I tried to look up what kind of distro/tools creality use or has, but they basically give no information at all. It's kind of hard to support a platform that tries to keep you in the dark. But I did find this decompressed repo: https://github.com/Guilouz/Creality-K1-Extracted-Firmwares/tree/main/Firmware (from the maker of the creality helper script). So it looks like the K1 uses busybox for it's tooling (and thus ash) for sh. They do seem to have make & git so that's good.

Currently the script does use quite a bit of bash only features (associative arrays, extra parameter expansion features, regex tests...) that would be quite a bit of work to convert to be sh compliant, but not impossible.

On another note. Did anyone try the menuconfig yet and feel like it would be intuitive to a general end user?

wdcocq:
> I split parts of the cfg into separate components and only insert them when required. Imho this makes the intend of `copy_config_files` much clearer and less error prone.

To continue on this.. is it preferred to always install all addon cfg files, or only when they are selected/enabled in the menu? They can always go back and change the menu and add them later

wdcocq:
In the current installer buzz_gear_on_down is set to both 3 and 1 for ERCF types see:
https://github.com/moggieuk/Happy-Hare/blob/28b0826bbc1dae896b8f25c0dbacf69c32ca3ca0/install.sh#L1192-L1195

Is this an eSpooler thing @ammmze? I see it's been added for eSpooler support

wdcocq:
Ah ok, honestly I actually have no clue what the eSpooler does :sweat_smile: I just copy most settings as they are now in the current installer. I've set them to 3 for ERCF for now.

wdcocq:
The state of this PR should now be up to par with the current installer in v3beta branch. And my goal of streamlining it to work in NixOS has also been achieved (I currently use it like that now)

I think it would be nice to add a few more things to the menuconfig but at least the current implementation should at least do everything `install.sh` does, plus a few small extras.

The only thing currently missing is some of the `install.sh` flags. Some functionality is there but you need to override env variables for that. e.g: `make SKIP_UPDATE= install`  for `./install.sh -z`. Not sure if we want to go that route. Maybe it's an idea to reintroduce a small `install.sh` that just forwards everything to `make`?

moggieuk:
Hi. Just a quick update.  I'm going to get v3.0 out first, then get back to looking at this PR.  There is a lost to process as well as some doc impact so it makes sense to make it entirely it's own release.   Anyway, I think the current install and cfg files are now 100% done for v3.0...

wdcocq:
It seems like this PR got closed automatically because of the deleting of the v3beta branch.
I don't think i have to permission to re-open it again and change the base branch?

moggieuk:
Glad I found this.  I see that it was against the v3beta branch (which is now main)..   I didn't think about the impact to PR's in progress..  Sorry!!

I'm not an expert with git so will have to see if this could be switched to main branch.

moggieuk:
Reopening..

PR #494, opened 2024-11-02
Minor changes to z-hop min floor logic & Prusa style stamping option for tip forming

### LINKED ISSUE #486: HH 2.7.3 z-hopping when park_toolchange is set to -1,-1 to disable move / park
Use case - standard, purge block based filament changes, no blobifier or cutter.

Setting variable_park_toolchange        : -1,-1, 0, 0, 2 to prevent HH from moving/parking on toolchange so it remains above the purge block now causes it to z-hop 2mm following 2.7.3 update. Presume variable_min_toolchange_z       : 2.0 now enforces this when z <= min height whether parking or not. Since I'm not moving the head from the purge block min_toolchange logic should be conditional and only applied when x & y >= 0 or when z-hop is explicitly set.

14:42:05 // Parking toolhead at (x:n/a, y:n/a, z:2.0) for toolchange operation
14:42:05 // Restoring MMU gear stepper run current to 100% configured
14:42:08 // Run Current: 0.89A Hold Current: 0.21A
14:42:08 // Unsynced MMU from extruder
14:42:08 // Unloading filament...

ningpj:
Suppressed z-hop min floor when park moves aren't defined for the mmu sequence to keep nozzle on purge block during toolchange.  e.g. when the user has used ``variable_park_toolchange : -1,-1, 0, 0, 2`` 

Also incorporated Prusa style stamping moves as part of tip form macro in cooling move loop (only stamps on even moves). Controlled using 3 variables: ``use_stamping`` (``true`` | ``false``), ``stamping_insertion_distance`` mm & ``stamping_insertion_speed``.  Currently using ``unloading_speed_start`` value for the retraction rather than creating another parameter.  Need to check and review move offsets as ``stamping_insertion_distance`` "should" technically be from the middle of the cooling tube according to Prusa slicer  

Still testing so not ready to merge.

Fixes #486

### DIFF
--- a/config/base/mmu_form_tip.cfg
+++ b/config/base/mmu_form_tip.cfg
@@ -58,6 +58,7 @@ gcode:
     {% set parking_distance = vars['parking_distance']|default(0)|float %}
     {% set orig_temp = printer.extruder.target %}
     {% set next_temp = params.NEXT_TEMP|default(orig_temp)|int %}
+    
     {% set use_stamping = vars['use_stamping']|default(false)|lower == 'true' %}
     {% set stamping_distance = vars['stamping_distance']|float %}
     {% set stamping_insertion_speed = vars['stamping_insertion_speed']|int %}
@@ -71,6 +72,11 @@ gcode:
     G91 # Relative positioning
     M83 # Relative extrusion
     G92 E0
+
+    # undo HH retract for tip forming
+    {% if park_vars.retracted_length > 0 %}
+        _MMU_UNRETRACT
+    {% endif %}
 
     # Step 1 - Ramming
     # This is very generic and unlike slicer does not incorporate moves on the wipetower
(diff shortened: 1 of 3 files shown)


moggieuk:
Happy to add this when you are complete. Can you ensure the defaults in FORM_TIP macro default to turning stamping off in the absence of the variables defined in `mmu_macro_vars.cfg`.  That way I don't need  a version number change to force users to so the "upgrade install".   I.e. everything will work with a simple update-manager refresh.  Users who want the functionality can of course run `./install.sh` to upgrade.  The reason is that I've had a lot of forced updates lately and I have v3.0 coming very soon..

ningpj:
Seems to be working fine.  I've kept stamping distance consistent with prusa's definition which is the cooling tube position rather than the top of the tube which is used by skinnydip.  I wanted to keep it consistent for peeps not using it in a standalone manner to confirm slicer settings.   I also undo the HH ooze retraction at the start of tip forming as ramming, stamping, and skinny dip all need to reliably extend back into the melt pool. With the retract in place as it was ramming is ineffectual unless you rammed significant volumes to allow the filament to reach the melt pool.  People who use tip forming and have manually increased ramming and skinny dip distances after the ooze retraction was incorporated will need to reduce and retune them.  Unsure what affect the initial HH retraction will have on the shape of the tip as this is a difference between standalone and slicer tip forming processes. in my case its a 2mm retract and extrude in fairly short order.

I also added an adjustment when the calculated top of the melt pool is less than the slicer enforced 15mm and a warning to increase the cooling tube position.  I considered allowing users to override the 15mm distance when operating in standalone mode for v.compact hotends but haven't done so as yet.

ningpj:
Finished ad tested.  just needs final review and merge

PR #497, opened 2024-11-03
Auto bowden calibration w/o encoder

ammmze:
The main thing I'd like to introduce here is the ability to perform the auto bowden calibration when you have an extruder entry sensor. 

There's a handful of additional changes included which can be seen as individual commits.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -2462,14 +2462,15 @@ class Mmu:
 
         try:
             with self.wrap_sync_gear_to_extruder():
-                self.calibrating = True
-                if manual:
-                    self._calibrate_bowden_length_manual(approx_bowden_length, save)
-                else:
-                    # Automatic method with encoder
-                    self._reset_ttg_map() # To force tool = gate
-                    self._unload_tool()
-                    self._calibrate_bowden_length_auto(approx_bowden_length, extruder_homing_max, repeats, save)
+                with self._wrap_suspend_runout():
+                    self.calibrating = True
+                    if manual:
+                        self._calibrate_bowden_length_manual(approx_bowden_length, save)
+                    else:
+                        # Automatic method with encoder
+                        self._reset_ttg_map() # To force tool = gate
+                        self._unload_tool()
+                        self._calibrate_bowden_length_auto(approx_bowden_length, extruder_homing_max, repeats, save)
         except MmuError as ee:
             self.handle_mmu_error(str(ee))
         finally:
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -2153,8 +2153,10 @@ class Mmu:
                     self.log_always("Failed to detect a reliable home position on this attempt")
 
                 self._initialize_filament_position(True)
+                self._wrap_gcode_command(self.pre_unload_macro, exception=True, wait=True)
                 self._unload_bowden(reference)
                 self._unload_gate()
+                self._wrap_gcode_command(self.post_unload_macro, exception=True, wait=True)
 
             if successes > 0:
                 average_reference = reference_sum / successes
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -3953,7 +3953,19 @@ class Mmu:
             length = self.toolhead_extruder_to_nozzle - self.toolhead_sensor_to_nozzle if self.sensor_manager.has_sensor(self.ENDSTOP_TOOLHEAD) else self.toolhead_extruder_to_nozzle
             length = min(length + self.toolhead_unload_safety_margin, homing_max)
             self.log_debug("Performing synced pre-unload bowden move to ensure filament is not trapped in extruder")
-            _,_,_,_ = self.trace_filament_move("Bowden safety pre-unload move", -length, motor="gear+extruder")
+            if self.gate_homing_endstop == self.ENDSTOP_ENCODER:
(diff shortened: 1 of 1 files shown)


extras/mmu/mmu.py
```diff
@@ -2456,14 +2459,14 @@ def cmd_MMU_CALIBRATE_BOWDEN(self, gcmd):
...
         try:
             with self.wrap_sync_gear_to_extruder():
-                self.calibrating = True
-                if manual:
-                    self._calibrate_bowden_length_manual(approx_bowden_length, save)
-                else:
-                    # Automatic method with encoder
-                    self._reset_ttg_map() # To force tool = gate
-                    self._unload_tool()
-                    with self._require_encoder():
```
ammmze, line 2466: We had another check here that was requiring the encoder for the auto calibration, but removed that, the condition check earlier in the method should be sufficient and covers encoder or extruder entry.

extras/mmu/mmu.py
```diff
@@ -3193,7 +3196,7 @@ def _initialize_encoder(self, dwell=False):
             self.encoder_sensor.reset_counts()
 
     def _get_encoder_dead_space(self):
-        if self.gate_homing_endstop in [self.ENDSTOP_GATE, self.ENDSTOP_POST_GATE_PREFIX]:
+        if self.has_encoder() and self.gate_homing_endstop in [self.ENDSTOP_GATE, self.ENDSTOP_POST_GATE_PREFIX]:
```
ammmze, line 3199: Reviewing the body of code in the bowden auto calibration it appeared to be mostly written to work with and without an encoder. This was the only method call in there that I found where it was missing an encoder check before using data related to the encoder.

extras/mmu/mmu.py
```diff
@@ -3883,7 +3886,19 @@ def _unload_gate(self, homing_max=None):
...
+                actual,homed,_,_ = self.trace_filament_move("Bowden safety pre-unload move", -length, motor="gear+extruder", homing_move=-1, endstop_name=self._get_gate_endstop_name())
+                # In case we ended up homing during the safety pre-unload, lets just do our parking and be done
+                # This can easily happen when your parking distance is configured to park the filament past the
+                # gate sensor instead of behind the gate sensor and the filament position is determined to be
+                # "somewhere in the bowden tube"
+                if homed:
+                    self._set_filament_pos_state(self.FILAMENT_POS_HOMED_GATE)
+                    self.trace_filament_move("Final parking", -self.gate_parking_distance)
+                    self._set_filament_pos_state(self.FILAMENT_POS_UNLOADED)
+                    return max(actual - self.gate_unload_buffer, 0)
```
ammmze, line 3901: This is my fix to avoid having this safety pre-unload eject my spool from the MMU completely. Basically just enables the homing check on the gate and if we happen to home in this move, then we'll consider us homed and move on same as if we weren't running the safety pre-unload.

extras/mmu/mmu.py
```diff
@@ -3903,7 +3918,8 @@ def _unload_gate(self, homing_max=None):
                 msg = "did not clear the encoder after moving %.1fmm" % homing_max
 
         else: # Using mmu_gate or mmu_post_gate sensor
-            actual,homed,_,_ = self.trace_filament_move("Reverse homing to gate sensor", -homing_max, motor="gear", homing_move=-1, endstop_name=self._get_gate_endstop_name())
+            endstop_name = self._get_gate_endstop_name()
+            actual,homed,_,_ = self.trace_filament_move("Reverse homing to %s sensor" % endstop_name, -homing_max, motor="gear", homing_move=-1, endstop_name=endstop_name)
```
ammmze, line 3922: Everything here and beyond is just including sensor names in the logs/errors for easier debugging.

extras/mmu/mmu.py
```diff
@@ -2132,6 +2132,16 @@ def _calibrate_bowden_length_auto(self, approximate_length, extruder_homing_max,
...
                 self._load_bowden(approximate_length)
+
+                if self.extruder_homing_endstop == self.ENDSTOP_EXTRUDER_ENTRY and self.sensor_manager.check_sensor(self.ENDSTOP_EXTRUDER_ENTRY):
+                    msg = "Fast move overshoot. The %s sensor triggered before homing. Try reducing your estimated BOWDEN_LENGTH." % self.extruder_homing_endstop
+                    self.log_error(msg)
+                    # Home back to the gate
+                    self._wrap_gcode_command(self.pre_unload_macro, exception=True, wait=True)
+                    self._unload_gate(approximate_length)
+                    self._wrap_gcode_command(self.post_unload_macro, exception=True, wait=True)
+                    raise MmuError("Fast move overshoot. The %s sensor triggered before homing. Try reducing your estimated BOWDEN_LENGTH." % self.extruder_homing_endstop)
```
ammmze, line 2143: This prevents "successful" bowden calibration when you give it too large of an estimated bowden length. What would happen if you provided too large of an estimated length is it could collide with the extruder and slip on the filament. Then when it does the `_home_to_extruder` it would see the sensor is already triggered so it doesn't do any additional homing and we end up with an "calibrated" length that is too long. These changes will detect that sensor triggering before it was supposed to and instructs the user to try again with a shorter bowden length.

moggieuk:
Nice.  I reviewed and everything looks good, except for the wrapper calling load/unload macros.  I *think* a call to _unload_sequence() will work and be more consistent/safer.

moggieuk:
I merged, but still want to review the calls to start/stop rewinder..

PR #501, opened 2024-11-05
Move retract to directly after the blob.

Swamp-Ig:
Moved the final retract to the end of the blob, before the z-lift.  Cuts the blob off cleanly, instead of leaving a string on it. This means the blobs separate cleanly in the bucket instead of all tangling together.

My initial testing shows this works better.

I will do some more testing on this and see how it goes, but also released here so others can review it. Overall it makes sense to me to do it this way.

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -470,6 +470,9 @@ gcode:
           
         {% endfor %}
 
+        # Retract to match what Happy Hare is expecting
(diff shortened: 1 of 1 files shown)


moggieuk:
Makes sense to me. @Dendrowen ok with you? (I'm trying to change blobifier without your thumbs up)

Dendrowen:
Yes, looks good to me!

Swamp-Ig:
Nice! ♥️

PR #505, opened 2024-11-06
refactor: Finalize respooler > eSpooler rename

ammmze:
Also includes the one change to the runout sensor to use `print_stats` instead of `idle_timeout` to determine if we are "printing".

### DIFF
--- a/extras/mmu_sensors.py
+++ b/extras/mmu_sensors.py
@@ -93,7 +93,7 @@ class MmuRunoutHelper:
 
     def _process_state_change(self, eventtime, is_filament_present):
         # Determine "printing" status
-        is_printing = self.printer.lookup_object("idle_timeout").get_status(eventtime)["state"] == "Printing"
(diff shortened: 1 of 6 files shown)


extras/mmu_sensors.py
```diff
@@ -93,7 +93,7 @@ def note_filament_present(self, is_filament_present):
 
     def _process_state_change(self, eventtime, is_filament_present):
         # Determine "printing" status
-        is_printing = self.printer.lookup_object("idle_timeout").get_status(eventtime)["state"] == "Printing"
+        is_printing = self.printer.lookup_object("print_stats").get_status(eventtime)["state"] == "printing"
```
ammmze, line 96: This change is un-related to the respooler rename. It is in a separate commit and can put it in a separate PR if preferred.

moggieuk: want be to pull and add idle_timeout fallback?

PR #506, opened 2024-11-07
feat: Update installer for eSpooler and add AFC Lite v1.0 board

ammmze:
Individual changes are grouped in separate commits in the PR.


install.sh
```diff
@@ -1466,6 +1485,7 @@ questionaire() {
     option EASY_BRD_RP2040      'EASY-BRD with RP2040'
     option MELLOW_BRD_1         'Mellow EASY-BRD v1.x (with CANbus)'
     option MELLOW_BRD_2         'Mellow EASY-BRD v2.x (with CANbus)'
+    option AFC_LITE_1           'AFC Lite v1.0'
```
ammmze, line 1488: Do we want to restrict this to only BoxTurtle or perhaps anything with virtual selector? I didn't set up pin defs for type-a systems for this, so probably should either omit this option for type-a or will need to pick some pin defs for type a using this board.

moggieuk: Ideally we would define something for type-A printers, but since it is unlikely to be used we can just leave the pin defs empty for now..

PR #515, opened 2024-11-13
Fix mmu_sequence.cfg wrong variable names

luckyandy0714:
Change variable name "lift_speed" to "park_lift_speed".


moggieuk:
Can you give more more context to the problem that was being caused but the former name?  I have no problem changing variable names for clarify (to do this a lot) but then I'd want the upgrade to handle this.   Would need some logic added to the:
```
read_previous_config()
```
section of `install.sh` to perform the upgrade -- they are many example of this.

moggieuk (approved):

PR #516, opened 2024-11-14
Remove duplicate section in espooler

wdcocq:
The section `[gcode_macro _MMU_ESPOOLER_VARS]` was defined in both `dc_espooler.cfg` and `dc_espooler_hw.cfg` 
I presume it was some leftover from a refactor. I removed the redundant section in `dc_espooler_hw.cfg`

### DIFF
--- a/config/addons/dc_espooler_hw.cfg
+++ b/config/addons/dc_espooler_hw.cfg
@@ -10,56 +10,6 @@
 # See https://www.klipper3d.org/Config_Reference.html#output_pin
 #
 
-####################################
-# Variables for the eSpooler macros
-#
-# Configure these for your setup.
-#
-[gcode_macro _MMU_ESPOOLER_VARS]
-# Prefix of name of the `output_pin` for the eSpooler.
-# The `output_pin` name must follow the pattern {prefix}_rwd_{gate}
-# and {prefix}_en_{gate}. By default we prefix it with an underscore
-# to prevent them from being shown in UI applications (like mainsail).
-#
-variable_pin_prefix: '_mmu_dc_espooler'
-
-# Default max number of seconds for eSpooler to run
-# Note: Each time any eSpooler **starts** it will restart the timeout
-#
-# variable_default_timeout: 60
-
-# The step speed where you want to max out the eSpooler to run at full speed.
-#
-# variable_max_step_speed: 200
-
-# Minimum distance of a move required to activate the eSpooler
-# The lowest valid value for this is 50 because eSpooler macros
-# will not even be considered if the distance is less than 50.
-#
-# variable_min_distance: 200
-
-# Adjusts the speed conversion ratio
-# For the following examples, let's assume max_step_speed = 50.
-# And remember actual eSpooler speed values are between 0.0 (off) and 1.0 (full speed) (inclusive)
-#
-# The forumula looks like this:
-# ({step_speed} / {max_step_speed}) ^ {step_speed_exponent}
-#
-# With step_speed_exponent of 1 would have a linear ratio:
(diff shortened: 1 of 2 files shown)


wdcocq:
@ammmze If this is not the case, please let me know!

ammmze:
@wdcocq Thanks for bringing this up. It was actually intentional. With the HH addons the intent is that the non-hw file (e.g. `dc_espooler.cfg`) should be read-only and users should make changes only to the hw files (e.g. `dc_espooler_hw.cfg`). So in the `dc_espooler.cfg` I defined all the default values and then in the `dc_espooler_hw.cfg` users are intended to override variables as-needed. And I chose to leave most of them commented out by default so that by defaults users would use and inherit the "managed" default values. However, as you bring this up, I suppose as long as we update the control (aka ctl) macro itself to handle default values, then I think we can remove the ones in the read-only file (`dc_espooler.cfg`), but leave the ones in the user managed file (`dc_espooler_hw.cfg`).

ammmze:
To clarify my comment above about read-only and user managed files. Technically in mainsail currently the user is able to edit both files, but from what I understand, when you run the installer, the hw file will get left alone, but the non-hw file will get updated. I didn't actually verify that, so might be worth verifying. And if it does get updated, does it preserve the variables? If does preserve them, then I think we could certainly just remove them from the hw file as you have them.

ammmze:
I just verified that it does indeed preserve the values for `variable_...` lines on when running `./install.sh`. So yea, I think we can go ahead and remove these from hw file. Thanks again for bringing this up!

ammmze (approved):

config/addons/dc_espooler_hw.cfg
ammmze, line 1: Oh I almost forgot, I had a note in the `dc_espooler.cfg` about the variables intended to be read-only. We should remove that note as well.

moggieuk:
LMK if this PR is ready to merge... just want to be sure.

Note on `cfg` file philosophy with HH..
- For addon's the `*_hw.cfg` is never upgraded because I don't want to overwrite the users configuration and it is often hard to parse out previous configuration because of duplicated terms (at least with the borne shell script currently used).
- Other  ("read-only") `cfg` files follow the pattern of reading previous config into memory, transforming any changes like name change, writing the NEW `cfg` file substituting the in-memory values.  This allows for upgrade of base file, for example to add new parameters, but retain user settings.

wdcocq:
It's because I was reworking the install script that I hit on this. as the double entry of the same section created warnings of unused parameters (though they were used in the other entry)

> For addon's the *_hw.cfg is never upgraded because I don't want to overwrite the users configuration and it is often hard to parse out previous configuration because of duplicated terms (at least with the borne shell script currently used).

The issue with duplicated terms has been addressed (in my PR anyway) by parsing parameters by not just name but also by their section. So similar parameters aren't an issue anymore

> Other ("read-only") cfg files follow the pattern of reading previous config into memory, transforming any changes like name change, writing the NEW cfg file substituting the in-memory values. This allows for upgrade of base file, for example to add new parameters, but retain user settings.

So this can also be applied to `_hw.cfg` files too if needed. and I currently do.

config/addons/dc_espooler_hw.cfg
wdcocq: I've replaced it with the note from _hw.cfg

ammmze:
> LMK if this PR is ready to merge... just want to be sure.

I just reviewed the latest updates and it looks good to me!

moggieuk:
I've pushed this to the `v3beta` branch. @wdcocq I recommend you switch to that because really "development" is my personal dev branch and often unstable:
```
./install.sh -b v3beta
```

PR #522, opened 2024-11-18
Blobifier: Amend display values to better match what is actually purged. Tweak pulsatile flushing.

igiannakas:
**A few proposed changes:**

1. When displaying the BLOBIFIER: Purging %dmm of filament message, dont include de-retraction moves that do not actually do any purge (i.e. park_vars.retracted_length)
2. The slow vs fast pulse volume ratio has been slightly tweaked

### Proposed change 1: When displaying the BLOBIFIER: Purging %dmm of filament message, don't include de-retraction moves that do not actually do any purge (i.e. park_vars.retracted_length)

The way the macro was currently structured, it was considering the park_vars.retracted_length as part of the purge length. However technically this is not a purge but rather a de-retraction move, bringing the filament to the start point, right before extrusion. Hence the display message was incorrect as it was incrementing the true purge value with a de-retraction move that results in no material exiting the nozzle.

In addition the application of the "purge minimum" ceiling, also considered the retracted length as part of the purged amount. Therefore, the produced blob would technically be smaller with a higher retract length vs. with a smaller retract length, even though the user would be expecting a fixed minimum purge amount to be extracted.

The proposed changes in the macro split the purge length calculation into two parts
1. The actual purge value that is incremented by the user requested purge_length_addition and that value is displayed in the UI. 
2. It then adds the residual de-retraction moves and executes the purge move (actual purge amount + deretraction move)

### Proposed change 2: Tweaks to pulsatile flushing
Finally more of a tweak - the slow vs fast pulse volume ratio has been amended from 95-5% (fast - slow) to 90-10% (fast - slow). While this will ever so slightly increase the purge time, it helps a lot with reducing the required purge amount by increasing flushing effectiveness ever so slightly (experimentally).

### Notes:
Raising as draft as currently in testing.

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -345,7 +345,7 @@ gcode:
       {% endif %}
       {% set purge_len = purge_vol / filament_cross_section %}
 
-      {% set purge_len = purge_len + printer.mmu.extruder_filament_remaining + park_vars.retracted_length + purge_length_addition %}
+      {% set purge_len = purge_len + park_vars.retracted_length + purge_length_addition %}
 
     {% else %} # ========================= USE CONFIG VARIABLE =============================
       {action_respond_info("BLOBIFIER: No toolmap or PURGE_LENGTH. Using default")}
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -325,16 +325,9 @@ gcode:
     # ==================== DETERMINE PURGE LENGTH ==========================================
     # ======================================================================================
 	
-	
     {% if params.PURGE_LENGTH %} # =============== PARAM PURGE LENGTH ======================
       {action_respond_info("BLOBIFIER: param PURGE_LENGTH provided")}
       {% set purge_len = params.PURGE_LENGTH|float %}
-      # APPLY PURGE MINIMUM
-      {% set purge_len = [purge_len,purge_length_minimum]|max|round(0, 'ceil')|int %}
-      {action_respond_info("BLOBIFIER: Purging %dmm of filament" % (purge_len))}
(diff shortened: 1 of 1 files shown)


Dendrowen:
Proposed change 1: Happy Hare functionality is built with slicer compatibility in mind. Therefor taking these settings into account is not desirable. Like the ramming settings which are also done in the firmware rather than the slicer. 

The changes also introduce duplicate code. For maintainability I try to maintain a strict DRY policy.

igiannakas:
> Proposed change 1: Happy Hare functionality is built with slicer compatibility in mind. Therefor taking these settings into account is not desirable. Like the ramming settings which are also done in the firmware rather than the slicer.
> 
> The changes also introduce duplicate code. For maintainability I try to maintain a strict DRY policy.

Indeed Orca seems to be the only one implementing the residual filament in nozzle parameter and including in the purge matrix. I'll revert change 1.

igiannakas:
OK code reworked and tweaked. Please take a look when you can on the other two points? Thank you :) 

I'm testing it now as well.

Dendrowen:
When using G1 moves, F determines the movement speed rather then the extrusion speed, therefor the speed is the same, but the z-movement is scaled to the correct proportion.

igiannakas:
> When using G1 moves, F determines the movement speed rather then the extrusion speed, therefor the speed is the same, but the z-movement is scaled to the correct proportion. At least in my experience?

You're absolutely right - I need more coffee. I've reverted it and instead adjusted the %ages to do 90-10 fast-slow instead of 95-5 as that was a bit on the short side for the slow extrusion. My apologies for this!!

igiannakas:
Closing this as has been superseded with further experimentation as documented in the discord channel.

PR #524, opened 2024-11-18
Parse flush multiplier and adjust flush matrix accordingly to match Orca and Bambu requested purge volumes

igiannakas:
As I was experimenting with Blobifier I noticed that the purge matrix was not changing as the flush multiplier was increased or decreased in Orca slicer as I was trying to validate this Orca slicer PR I've raised here: https://github.com/SoftFever/OrcaSlicer/pull/7508

**For example:**
Multiplier of 1.2
image
Gcode file:
image

Multiplier of 0.5
image

Gcode file:
image

This results in an incorrect purge matrix being sent to the blobifier for use when purging that does not match the intended action from the user in the UI. Instead the purge matrix is always as if 1.0 multiplier was set in the UI.

This happens because both Orca and Bambu generate a base purge matrix and a separate multiplier and they output both to the gcode for use by Happy Hare etc. To calculate the effective purge volume, these two values need to be multiplied so they match the user intent.

This PR proposes to parse the flush multiplier and apply it to the purge matrix before use in the placeholder variables.

**Results:**
For example, now with a 0.5 multiplier the purge volumes correspond to what the UI displays
image

### DIFF
--- a/components/mmu_server.py
+++ b/components/mmu_server.py
@@ -772,6 +772,8 @@ METADATA_MATERIALS = "!materials!"
 PURGE_VOLUMES_REGEX = r"^;\s*(flush_volumes_matrix|wiping_volumes_matrix)\s*=\s*(.*)$" # flush.. in Orca/Bambu, wiping... in PS
 METADATA_PURGE_VOLUMES = "!purge_volumes!"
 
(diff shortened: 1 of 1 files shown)


moggieuk (approved):
Thanks!  I guess most folks must be using a multiplier of 1.0 .. I know I'm guilty of that..

igiannakas:
Or believed it was using less plastic to purge as also orca didn’t produce the purge estimates up until the other Pr over there 😁 let’s hope there will be limited reports of color bleed on discord 😇

PR #532, opened 2024-11-20
chore: add a way to simply run the installer via docker in an isolated environment

ammmze:
Here's that quick little docker environment I've been using to test the installer in an isolated environment.


moggieuk (approved):
Awesome. Thanks.

PR #539, opened 2024-11-26
Update MMU_SELECT_TOOL to MMU_SELECT

Luoti:
I noticed, that my menu was not working for this command, and looks like it has been updated some time ago. This should fix the default menu.


moggieuk (approved):
THanks.  Not many users use the 12864 display anymore.

PR #544, opened 2024-12-06
handle klipper installed to /usr/share and /usr/data on MIPS

pellcorp:


pellcorp:
For SimpleAF (https://github.com/pellcorp/creality/wiki/Simple-AF) klipper actually gets installed to /usr/data/klipper, rather than /usr/share/klipper, so I am hoping we can make this minor change to the installer to handle both Simple AF and Helper Script.

moggieuk:
No problem. Merged.

Issue #512, opened 2024-11-12
The use of demultiplexers.


ghost:
Hello everybody.
If you want to use a 3MS or BoxTurtle mmu, then there is a need for a board with a sufficient number of stepper motor drivers. Thus: how many colors are there - so many stepper motors and, accordingly, stepper motor drivers. At the same time, each stepper motor requires at least three free pins of the microcontroller. If TMC drivers are used, then four pins are already required: en, step, dir, uart. It turns out that if I want to build myself a 3MS mmu with twelve colors, then I need to use two boards with support for six drivers. This sounds unreasonably expensive and inefficient, because usually half of the legs on the board are used to control the heater, fans, etc.

I had the idea of using demultiplexers to reduce the number of microcontroller legs used. As an example: to create an mmu like 3MS with 16 colors, only 7 microcontroller legs are required without using TMC drivers and 8 legs if TMC is used. It turns out that an additional board may not be useful for assembling a 16-color mmu if you use a demultiplexer. 

What is a demultiplexer? This is a digital logic device. It allows you to redirect the outgoing signal to the desired address. This very address is set in binary code. An illustration of the work is below. Thus, for the driver to work, three or four demultiplexers are needed, the address inputs of which will be connected together.  An approximate scheme is shown in the figure.

In this configuration, the system was tested, and for a fairly long time (about 10 days of continuous operation). There were no failures. In fact, this is an intermediate option between ERCF and 3MS. Each coil has a drive and a driver, but Klipper thinks that the motor is the same, because the en, step, dir signals for all motors come from the same pins. The disadvantage is the impossibility of simultaneous operation of several drives, but this is not required. The limit switches are connected individually and they will still need a number of pins equal to the number of limit switches.  I have a suggestion to add a function to work with demultiplexers in Happy Hare, of course, if possible. In fact, all that is required is to switch Switch0-Switch3 when changing the color. Example:

[output_pin switch_0_demultiplexor]
pin: EAZY:gpio3

[output_pin switch_1_demultiplexor]
pin: EAZY:gpio2

[output_pin switch_2_demultiplexor]
pin: EAZY:gpio1

[output_pin switch_3_demultiplexor]
pin: EAZY:gpio0

[gcode_macro Demultiplexor_chanel_0]
gcode = 
    SET_PIN PIN=switch_0_demultiplexor VALUE=0
    SET_PIN PIN=switch_1_demultiplexor VALUE=0
[...]

PR #545, opened 2024-12-09
MacroSelector Implementation

3DCoded:
This PR adds a new `MacroSelector` selector type that allows for several niche MMU setups. After setting `selector_type: MacroSelector` in `mmu_hardware.cfg`, you have two parameters you can set in `mmu_parameters.cfg`:

- `select_tool_macro`: This is **required** and is the macro run when selecting a tool. A `GATE` parameter gets passed to this macro
- `select_tool_num_switches` This is **optional** and will convert the selecting gate to binary, allowing setups with demultiplexers #512 or relay arrays. This will pass the binary number to the `select_tool_macro` as `S0`, S1`, S2`, etc. with each `S` parameter reflecting one binary digit.

Example `select_tool_macro` for a demultiplexer- or relay-style setup:

```
[gcode_macro SELECT_TOOL]
gcode:
    SET_PIN PIN=d0 VALUE={params.S0}
    SET_PIN PIN=d1 VALUE={params.S1}
    SET_PIN PIN=d2 VALUE={params.S2}
    ...
```

Example `select_tool_macro` for an optocoupler-style setup:

```
[gcode_macro SELECT_TOOL]
gcode:
    SET_PIN PIN=o{printer.mmu.gate} VALUE=0
    SET_PIN PIN=o{params.GATE} VALUE=1
```


moggieuk:
Oh, this is a nice "catch-all" selector idea....

3DCoded:
@moggieuk 

Thank you for the quick reply and review. I have updated my fork as requested with Python 2 support, updated `get_mmu_status_config()`, and `_wrap_gcode_command()` usage.

PR #546, opened 2024-12-09
strip # symbol before alpha channel truncation

CooperGerman:


moggieuk:
Thanks for this (and for cleaning up trailing spaces ... those were from another PR).

I can merge this now but just checking if you wanted to/needed to do something similar in mmu.py?

CooperGerman:
> Thanks for this (and for cleaning up trailing spaces ... those were from another PR).
> 
> I can merge this now but just checking if you wanted to/needed to do something similar in mmu.py?

I don't think so as the color only seems to be retrieved in mmu_server i think. 

For the trailing spaces, my IDE removes them automatically (vscode) i was wondering if it would bother you but i see it is OK ;)

PR #572, opened 2024-12-22
Move park_operation variable update

kohlerj:
I would argue that setting this variable should only be done once an actual park operation has been performed.

### DIFF
--- a/config/base/mmu_sequence.cfg
+++ b/config/base/mmu_sequence.cfg
@@ -200,9 +200,9 @@ gcode:
             {% elif x < 0 and y >= 0 %}
                 G1 Y{y} F{park_travel_speed}						# Y move only
             {% endif %}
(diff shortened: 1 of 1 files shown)


moggieuk:
You might be right by I worry about corner cases...  this variable is used to choose what is done during the "RESTORE_POSITION" (e.g. whether to unretract, etc).  Therefore if not set this logic would be incorrect.

Really the "is homed" test is to prevent unnecessary errors but continue as best we can.

If it was called "attempted_park_operation" it's current use would be correct so I think it is just the name that is confusing...

kohlerj:
Mmm the issue is that the machine will not move on subsequent calls to the same park position, because the software thinks it's at the correct location. 
 
 I see your point with the use in RESTORE_POSITION. 
 
 Seems the only way to have two variables...

PR #582, opened 2025-01-02
Fix bowden load correction moves

Leonti:
This fixes logic for bowden correction moves.

Example scenario:
 1. We move `1500mm` with fast bowden move. Encoder measured `1480mm`. The delta is `20mm` which we need to correct.
 2. First correction move will attempt to move `20mm` which will move for example `18mm` according to the encoder, so now our delta is `2mm`. So we are away from `1500mm` destination by only `2mm` and we are at `1498mm`
 3.  The delta is updated like this `delta -= d`, so the new delta will be `20mm - 2mm = 18mm`. So when we do the second correction move we will move past the target and end up at `1498mm + 18mm = 1516`

This PR is a simple change to update the original `delta` with the delta from the correction move. Updated logic:
 1. We move `1500mm` with fast bowden move. Encoder measured `1480mm`. The delta is `20mm` which we need to correct.
 2. First correction move will attempt to move `20mm` which will move for example `18mm` according to the encoder, so now our delta is `2mm`. So we are away from `1500mm` destination by only `2mm` and we are at `1498mm`
 3.  The delta is updated like this `delta = d`, so the new delta will be `2mm`. If it's less than the `bowden_allowable_load_delta` then we are done and we are at `1498mm`, if we need `1mm` precision the second correction move will move only remaining `2mm` to end up at `1500mm` as expected.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -4190,9 +4190,10 @@ class Mmu:
                         if delta >= self.bowden_allowable_load_delta:
                             msg = "Correction load move #%d into bowden" % (i+1)
                             _,_,_,d = self.trace_filament_move(msg, delta, track=True)
-                            delta -= d
+                            delta = d
                             self.log_debug("Correction load move was necessary, encoder now measures %.1fmm" % self.get_encoder_distance())
                         else:
+                            self.log_debug("Correction load complete, delta %.1fmm is less than 'bowden_allowable_unload_delta' (%.1fmm)" % (delta, self.bowden_allowable_load_delta))
                             break
                     self._set_filament_pos_state(self.FILAMENT_POS_IN_BOWDEN)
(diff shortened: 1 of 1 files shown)


igiannakas:
Awesome! I had noticed something similar to you - the filament crashing into the extruder when correction moves were applied. This looks promising!

moggieuk:
Thanks for this.  Despite my comment to you on the associated issue and the improvements with v3.0 this may well be a bug.  Will validate shortly!

moggieuk:
Ok.  Yes, work is the "delta" has tripped me up before, I think I was thinking to reduce the correction by the amount actually moved, i.e:
`delta -= (delta - d)`

which is of course
`delta = d`

@Leonti  you win the award for finding the oldest bug in Happy Hare .. it must originate from the very first release!

PR #593, opened 2025-01-09
various spelling fixes

burkfers:
I've given the entire repo a once-over with the dictionary.

There is only two changes in executed code, both in `action_respond_*`s. All other changes are in comments only.


moggieuk:
Oh, thank you.  I like polish like this... will merge very soon.

PR #605, opened 2025-01-21
Warn if user defined accels exceed max_accels defined for the printer.

ningpj:
Apply sane limits to prevent users from defining brush and shake accel's that exceed max_accel settings of the printer to minimise potential for missed steps.  Checks added to delayed_gcode blobifier init block

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -867,7 +867,15 @@ gcode:
     {action_respond_info("BLOBIFIER: variable_z_raise_exp has value: %f. This value is out of spec (0.5 ... 1.0)." % (bl.z_raise))}
   {% endif %}
 
-
+  # cap user defined accels at printer max_accel if greater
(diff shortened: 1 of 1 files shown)


ningpj:
// BLOBIFIER: variable_shake_accel has value: 8000 which is higher than your printer limit of 6180. Resetting to match your printer limit.
// BLOBIFIER: variable_brush_accel has value: 8000 which is higher than your printer limit of 6180. Resetting to match your printer limit.

ningpj:
// BLOBIFIER: variable_shake_accel has value: 8000 which is higher than your printer limit of 6180. Resetting to match your printer limit.
// BLOBIFIER: variable_brush_accel has value: 8000 which is higher than your printer limit of 6180. Reseting to match your printer limit.

ningpj:
Ready for review and merge. simple change

ningpj:
Changed to warning at startup rather than capping accel's at max printer setting

PR #606, opened 2025-01-22
Update blobifier.cfg

trajesus:
- Update some inconsistencies in the variables naming (bb in some macros and bl in others to represent blobifier)
- Reordered some movements to prevent crashing into the tray

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -370,13 +370,14 @@ gcode:
       # ==================== POSITIONING ===================================================
       # ====================================================================================
       
+      # Retract the tray so it is not in the way
+      BLOBIFIER_SERVO POS=in
+
       # Move to the assembly, first a bit more to the right (brush_start) to avoid a 
       # potential filametrix pin if it's not already on the same Y coordinate.
       {% if printer.toolhead.position.y != position_y %}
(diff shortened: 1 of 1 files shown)


ningpj:
I noticed the bb inconsistency as well but are a way off submitting my PR for changes to support left and right hand setups, chaining of wiping across fixed and gantry wipers, etc.  Because I'm reworking movement and collision avoidance logic i'm interested in understanding a little more about the tray collision issue you were trying to solve?  I see you retract the tray and moved  purge_X X positioning after the Z descent before reextending the tray....what were you seeing with your setup?

trajesus:
The change of bb to bl in every macro is just for consistency.
I added the retract of the tray and moved the x move because I had some collisions while setting up my ERCF and while printing, on the first few layers, as my tray sits at 7.5mm and the z_hop was not handling it correctly, so the toolhead was actually moving up against the tray, colliding into it.

trajesus:
I also noticed that not every macro considered the skew correction variable, so I updated that. 

I noticed that some variables are not being used, as well. This might be because you are starting to rework the collision avoidance, so I left them unchanged.

ningpj:
OK, thanks for the insight. that makes sense since your tray is elevated and above the bed.  ``clearance_z`` doesn't help in this situation either as its only used in the wipe macro.  I added a z-hop check in my logic before attempting to position X so have this covered. Thanks for sharing

PR #607, opened 2025-01-22
Cutting motion for two dimensional cutting action

ksummers92:
- Added additional movement axis to cutter action to support non-linear cutter such as [this cutter](https://www.printables.com/model/1142687-filament-cutter-for-reaper-toolhead-and-orbiter-pa) to allow for smoother operation and provide more leverage
- Update mmu_macros_vars.cfg defaults to include a y-axis dimension to the compressed position
- Added some logic to _FILAMETRIX_DO_CUT_MOTION macro to allow backwards compatibility for existing configurations with single dimensional compressed location
	- It assumes that the omitted second dimension is the same as the park location coordinate
- Tested with and without y-axis dimension provided and operation works as before
- Tested switching to y-axis cutting direction as the primary direction and it also works
- Attached video shows exaggerated cutter motion in which x and y position are changing

https://github.com/user-attachments/assets/cde72a52-4829-4b21-8944-a9077e806191

### DIFF
--- a/config/base/mmu_macro_vars.cfg
+++ b/config/base/mmu_macro_vars.cfg
@@ -331,7 +331,7 @@ variable_pin_park_dist          : 5.0		; Distance in mm
 # Position of the toolhead when the cutter is fully compressed. This value is on the x or y axis depending on the value of
 # 'cutting_axis'. Should leave a small headroom (should be a bit larger than 0, or whatever xmin is) to
 # avoid banging the toolhead or gantry
-variable_pin_loc_compressed     : 0.5, 250		; x,y coordinates of fully depressed location
+variable_pin_loc_compressed     : 0.5, 250		; x,y coordinates of fully depressed location, assumes y is equal to pin_loc y position if omitted
 
 # Retract length and speed after the cut so that the cutter blade doesn't
 # get stuck on return to origin position
--- a/config/base/mmu_macro_vars.cfg
+++ b/config/base/mmu_macro_vars.cfg
@@ -320,7 +320,7 @@ variable_simple_tip_forming     : True		; True = Perform simple tip forming, Fal
 
 # This should be the position of the toolhead where the cutter arm just
 # lightly touches the depressor pin
-variable_cutting_axis		: "x"		; "x" or "y". Determines cut direction (axis) during cut motion
+variable_cutting_axis		: "x"		; "x" or "y". Determines cut direction (axis) during cut motion, used for park distance
 variable_pin_loc_xy             : 14, 250	; x,y coordinates of depressor pin
 
 # This distance is added to "pin_loc_x" or "pin_loc_y" depending on the 'cutting_axis'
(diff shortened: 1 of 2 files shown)


config/base/mmu_macro_vars.cfg
moggieuk, line 1: `variable_pin_loc_compressed` would need to be upgraded in `install.sh` otherwise this would break previous installs. There is a function called:
`read_previous_config() {`
Logic should be added at the end of this function (similar to examples that are already there) that convert a previous scalar value to an list of 2 elements.  Make sure this is tested to "upgrade" then run again to verify no further changes. I.e. the logic correctly spots a scalar value

Ah, I missed the fact that you handle scalar and a list in the code.  Perhaps update the comment for `variable_pin_loc_compressed` to explain x or y or x,y capability...

ksummers92: I am not familiar with bash enough to update the install.sh to do what would be appropriate to update the parameter, so I decided to handle it this way. I tried to take a stab at it, but I am not sure how to detect if an variable(?) has a length of two? Or contains a ","? The syntax of what it is seeing when pulling out the parameters is lost on me...

moggieuk:
oming_extruder=${_param_homing_extruder}
    fi

moggieuk:
Here is what I'd recommend (sorry been really busy):
1. You rename "variable_pin_loc_compressed" to "variable_pin_loc_compressed_xy" for consistency with other variables that take x and y coordinates.
2. Add logic similar to this (completely untested but directionally correct) to install.sh at the end of the `read_previous_config()` method:
```
    # v3.1.0
    if [ "${_variable_pin_loc_compressed}" != "" ]; then
        pin_loc_y=$(echo ${_variable_pin_loc_xy} | cut -d ',' -f2)
        _variable_pin_loc_compressed_xy="${_variable_pin_loc_compressed}, ${pin_loc_y}"
    fi
 ```
 The logic is to cut the "y" parameter out of "pin_loc_xy", then create a new variable ` _variable_pin_loc_compressed_xy` composed of old `_variable_pin_loc_compressed` with added ", <pin_loc_y>".  The the installer will substitute that into the new `mmu_macro_vars` and delete the old variablel. On subsubsequent upgrades the `_variable_pin_loc_compressed` variable will not exist so the upgrade logic will not run again..
 3. Change the shipped `mmu_macro_vars.cfg` template to rename `variable_pin_loc_compressed` to `variable_pin_loc_compressed_xy` and adjust doc to match

ksummers92:
@moggieuk thank you for the feedback. I have made the appropriate updates that you suggested. I did not revert the changes in the _FILAMETRIX_DO_CUT_MOTION macro. Since the updates you suggest should make those changes unnecessary, do you suggest removing at least the backward compatibility part of the changes?

moggieuk:
Next in queue for review.  Thanks for your patience!

PR #619, opened 2025-02-04
Fix current spool in spoolman being activated after the load sequence is complete

### LINKED ISSUE #558: Spoolman doesn’t count filament use correctly in multi material prints
When doing multi color prints, I’ve noticed that spoolman doesn’t account for the full material consumption. For example the print below was about 30grams off, mostly on the white color (as this was the one consumed the most).

image

screenshot from mainsail - notice the discrepancy between the slicer estimate and the actual consumption as measured by klipper:

IMG_5122

I weighed the purged material, model and prime tower and the slicer is within a few grams of the actual print - see also here for testing on the slicer side: https://github.com/SoftFever/OrcaSlicer/pull/7508

basically what I’m seeing is spoolman recording less filament consumed, which can be a bit of an issue as I need to periodically re weigh the spools and adjust the material remaining manually.

for example, the below white spool should have had around 90 grams remaining according to spoolman yet it was under 20 grams left (it was on the last set of loops around the Center of the spool)
IMG_5119

please note, this only happens with multi color prints. Single color prints work correctly and spoolman reconciles down to the gram between slicer, actual and spool weight.

Observations when printing a multi color model:
1. During printing the model, spoolman deducts the weight correctly
2. During purging with the blobifier I do notice filament consumed, so that also is being measured

That leads me to believe that somewhere in the retractions/ cut/ unload / reload logic spoolman doesn’t potentially count the filament remaining in the hotend correctly (?) hence the discrepancy which grows the more tool swaps you do. 

This is probably expected as the extruder retracts more than the filament does when pulling the filament out from the toolhead, because it carries on retracting when the filament has already cleared the extruder top by design to ensure the filament is free from the extruder before running the Bowden unload move. 

This is probably being counted as extra filament being re wound but obviously it’s not there as the filament has already cleared the extruder. Specifically this variable I’m suspecting: 
toolhead_unload_safety_margin: 15

15mm over a few hundred swaps accumulates.

Or maybe the toolhead residual filament plus the remaining fragment after the cut is not counted?

Maybe a better strategy would be to re consider at what points the filament is counted as loaded in the extruder? Or increment the filament consumption by the remaining fragment after cut plus the toolhead residual filament value?

In general going conservative and having more material left over would be better than having less on the spool than what spoolman reports.

igiannakas:
Fixes active spool in spoolman being set after the load sequence is complete. 

The current behaviour triggers the post load macros and then associates the spool from spoolman, meaning that any blobifier purging is not counted against the active filament.

Have moved both the load and unload association from Spoolman for consistency in the code.

Fixes #558

image

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -4187,7 +4187,7 @@ class Mmu:
                 self._set_filament_direction(self.DIRECTION_LOAD)
                 self.selector.filament_drive()
 
-                # Record starting position for bowden progress tracking 
+                # Record starting position for bowden progress tracking
                 self.bowden_start_pos = self.get_encoder_distance(dwell=None) - start_pos
 
                 if self.gate_selected > 0 and self.rotation_distances[self.gate_selected] <= 0:
@@ -4715,6 +4715,9 @@ class Mmu:
             if full and not extruder_only and not self.gcode_load_sequence:
                 self._autotune(self.DIRECTION_LOAD, bowden_move_ratio, homing_movement)
 
+            # Activate loaded spool in Spoolman
+            self._spoolman_activate_spool(self.gate_spool_id[self.gate_selected])
+            
             # POST_LOAD user defined macro
             if macros_and_track:
                 with self._wrap_track_time('post_load'):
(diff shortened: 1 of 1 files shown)


moggieuk:
Thanks.  This is perfect!

PR #621, opened 2025-02-05
Enable bowden correction for unload moves

igiannakas:
To address issue where drag in the rewinder system could result in the gate homing max distance not being enough to fully pull out the filament from the encoder.

Also addresses issue where erec would cut the filament at the wrong spot, as the unload operation didn't bring the filament to the expected position, if drag in the buffer system caused slippage.

The logic uses the amended approach from this PR https://github.com/moggieuk/Happy-Hare/pull/582 to ensure that the right amount is retracted (vs the old one that overshot the requested correction).

Tested with load and unload Bowden console commands and the slippage is fully compensated (actually better than I thought!!)

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -4292,7 +4292,7 @@ class Mmu:
                                 if delta >= self.bowden_allowable_unload_delta:
                                     msg = "Correction unload move #%d from bowden" % (i + 1)
                                     # Move "backwards" by delta since we still haven't offloaded enough
-                                    _, _, _, d = self.trace_filament_move(msg, -delta, track=True)
+                                    _, _, _, delta = self.trace_filament_move(msg, -delta, track=True)
                                     delta = d
                                     self.log_debug("Correction unload move was necessary, encoder now measures %.1fmm" % self.get_encoder_distance())
                                 else:
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -4210,8 +4210,7 @@ class Mmu:
                         for i in range(2):
                             if delta >= self.bowden_allowable_load_delta:
                                 msg = "Correction load move #%d into bowden" % (i+1)
-                                _,_,_,d = self.trace_filament_move(msg, delta, track=True)
-                                delta = d
+                                _,_,_,delta = self.trace_filament_move(msg, delta, track=True)
                                 self.log_debug("Correction load move was necessary, encoder now measures %.1fmm" % self.get_encoder_distance())
                             else:
                                 self.log_debug("Correction load complete, delta %.1fmm is less than 'bowden_allowable_unload_delta' (%.1fmm)" % (delta, self.bowden_allowable_load_delta))
@@ -4293,7 +4292,6 @@ class Mmu:
                                     msg = "Correction unload move #%d from bowden" % (i + 1)
                                     # Move "backwards" by delta since we still haven't offloaded enough
                                     _, _, _, delta = self.trace_filament_move(msg, -delta, track=True)
-                                    delta = d
                                     self.log_debug("Correction unload move was necessary, encoder now measures %.1fmm" % self.get_encoder_distance())
                                 else:
                                     self.log_debug("Correction unload complete, delta %.1fmm is less than 'bowden_allowable_unload_delta' (%.1fmm)" % (delta, self.bowden_allowable_unload_delta))
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -4297,7 +4297,7 @@ class Mmu:
                                     self.log_debug("Correction unload complete, delta %.1fmm is less than 'bowden_allowable_unload_delta' (%.1fmm)" % (delta, self.bowden_allowable_unload_delta))
                                     break
                             if delta >= self.bowden_allowable_unload_delta:
-                                self.log_info("Warning: Excess slippage was detected in bowden tube unload after correction moves.Gear moved %.1fmm, Encoder measured %.1fmm. See mmu.log for more details" % (length, length - delta))
+                                self.log_info("Warning: Excess slippage was detected in bowden tube unload after correction moves. Gear moved %.1fmm, Encoder measured %.1fmm. See mmu.log for more details" % (length, length - delta))
                         else:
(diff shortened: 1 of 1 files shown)


extras/mmu/mmu.py
```diff
@@ -4285,8 +4285,23 @@ def _unload_bowden(self, length=None):
...
-                        # Only a warning because _unload_gate() will deal with it
-                        self.log_info("Warning: Excess slippage was detected in bowden tube unload. Gear moved %.1fmm, Encoder measured %.1fmm" % (length, length - delta))
+                        # Check and attempt correction if bowden_apply_correction is enabled
+                        if self.bowden_apply_correction:
+                            self.log_debug("Checking if correction move is needed on unload bowden")
+                            for i in range(2):
+                                if delta >= self.bowden_allowable_unload_delta:
+                                    msg = "Correction unload move #%d from bowden" % (i + 1)
+                                    # Move "backwards" by delta since we still haven't offloaded enough
+                                    _, _, _, d = self.trace_filament_move(msg, -delta, track=True)
```
norris108, line 4295: Rather than create a temporary variable, use `delta`
```suggestion
                                    _, _, _, delta = self.trace_filament_move(msg, -delta, track=True)
```

norris108 (changes requested):

extras/mmu/mmu.py
```diff
@@ -4285,8 +4285,23 @@ def _unload_bowden(self, length=None):
...
+                                    delta = d
```
burkfers, line 4296:
```suggestion
                                    _, _, _, delta = self.trace_filament_move(msg, -delta, track=True)
```
The line below should also be removed following @norris108 's suggestion.

igiannakas:
Updated, also updated the load part that was doing something similar ;)

norris108 (approved):

extras/mmu/mmu.py
```diff
@@ -4285,8 +4284,22 @@ def _unload_bowden(self, length=None):
...
+                                if delta >= self.bowden_allowable_unload_delta:
+                                    msg = "Correction unload move #%d from bowden" % (i + 1)
+                                    # Move "backwards" by delta since we still haven't offloaded enough
+                                    _, _, _, delta = self.trace_filament_move(msg, -delta, track=True)
+                                    self.log_debug("Correction unload move was necessary, encoder now measures %.1fmm" % self.get_encoder_distance())
+                                else:
+                                    self.log_debug("Correction unload complete, delta %.1fmm is less than 'bowden_allowable_unload_delta' (%.1fmm)" % (delta, self.bowden_allowable_unload_delta))
+                                    break
+                            if delta >= self.bowden_allowable_unload_delta:
+                                self.log_info("Warning: Excess slippage was detected in bowden tube unload after correction moves.Gear moved %.1fmm, Encoder measured %.1fmm. See mmu.log for more details" % (length, length - delta))
```
burkfers, line 4300:
```suggestion
                                self.log_info("Warning: Excess slippage was detected in bowden tube unload after correction moves. Gear moved %.1fmm, Encoder measured %.1fmm. See mmu.log for more details" % (length, length - delta))
```
Trivial nitpick: Missing space after full stop; But would be visible to users.

moggieuk:
Here is my concern... HH used to do this on unload (as it does on load).  The problem is that poor calibration or encoder issues can easily lead to "over unloading". This is something that cannot be undone .. the filament is ejected.   However, the safe way to handle is to let "unload_gate()" handle perhaps with a longer length... that will not over unload.  

That is the reason I removed the logic and simply replaced with warning message.  I think a better way to handle is to indicate to unload_gate() to increase the homing distance.  This can actually easily be done by setting the filament position to "IN_BOWDEN" rather than "START_BOWDEN"

igiannakas:
Hm… I get the point. Maybe a different way would be to cap the max correction adjustment up to the gate parking distance? That way you’ll never unload past the gate?

igiannakas:
Also just thinking out loud - if the encoder is not reliably measuring the filament won’t it crash on the extruder when loading and strip? Yes that’s probably reversible but with a long Bowden tube that’s an issue too as the flat spot will cause the unload to fail?

Also interesting idea setting the in Bowden position. The issue I had that spurred me to look at this was during main Bowden unload phase it slipped slightly and it happened to be the right amount of filament left that the gate unload routine pushed it out of the gate completely. And then it couldn’t be grabbed again for the next tool change 😁

Would setting the position to in Bowden do smaller moves to retract the filament?

Or just cap the adjustment up to gate unload buffer minus say 2xgate parking distance at most (this would bring it at most to the gate parking spot exactly). In which case the gate unload after shouldn’t eject the filament as there is no encoder movement?

I’ve got a print lined up for it tomorrow using that slippy lane with the code above and see how it does 😂 

Will add the max check to contain the adjustment as a test and offer some feedback :)

igiannakas:
Quick check before I go off and try your suggestion to set it in the In_Bowden state if excess slippage is observed.

Would this affect EREC in any way? Is that one expecting the state machine to be in the start_bowden state?

igiannakas:
So did some more tests with a ceiling set to avoid going past the gate. I’ll push the code shortly for you all to check out.

kind off see moggie’s point. The moment you add protections to avoid overrunning the gate, the compensation becomes exponentially less useful, as you are capped to how much you can correct. 

Which is counter to the point of correcting in the first place as you are still left with an “unpredictably” long amount left to cut / rewind. And that still exposes you to the same issue of running out of gate homing headroom or cutting too long of a filament tip with the EREC. 

Hm that’s a tough one. 

The thing that I have noticed though is that the homing moves are around 15mm long - I think this is because of the encoder step size which may be a bit too large causing the homing moves backwards to push the filament out of the gate on occasion, if it happens to have the filament just clear the encoder when another push back move is done. That is because in ERCF v2 the distance between the end of the encoder and the gate parking distance is around 10mm or so.

So maybe that is the solution? Reduce the encoder step size to be less than or equal to the gate parking distance?

igiannakas:
Thank you 🙏🏻 my python skills are pretty much non existent and starting with a 5k line python file is quite the undertaking :)

after my little printing accident yesterday need to get the printer back up and running again as it’s a PITA to try and test these with a half working cutter, blobifier servo etc. 

Got some parts coming today so hope to delve into this tomorrow / Friday.

thank you again!!

moggieuk:
No problem, and yes, HH is got a good introduction to python. Lol.   BTW It is my first python program..

Let me know how it goes. I know what you are trying to achieve but just want to protect against replacing one problem with another (unrecoverable overshoot).

PR #622, opened 2025-02-05
Check filament colour in PrusaSlicer as well

trajesus:
My files were not detecting the colours when pre-processed by Happy-Hare, found out, while looking at the gcode file that the line it was looking for was set as '; extruder_colour = ;;;;;;;' as my extruders are set to have the colour of the filament. This solves the problem by also looking at the filament colour (if the extruder colour is not set)

### DIFF
--- a/components/mmu_server.py
+++ b/components/mmu_server.py
@@ -758,7 +758,7 @@ METADATA_TOOL_DISCOVERY = "!referenced_tools!"
 
 # PS/SS uses "extruder_colour", Orca uses "filament_colour" but extruder_colour can exist with empty or single color
 COLORS_REGEX = {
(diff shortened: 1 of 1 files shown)


trajesus:
No problem, I'm happy to help! 
I'm also having some trouble with the purge volumes matrix, as I'm using Blobifier, with no purge block, they seem not to be generated. Do you know any solution for that?

trajesus:
I tested that before submitting the PR. It takes into account the extruder and only then, the filament even though it was only a simple modification to the expression.

The way that the re is set matches both the extruder colour and the filament colour. As the extruder colour always shows up first in the gcode file, if it has valid colours and matches the re, it is set, due to the way you have the logic after in the Python file. If it doesn't match in the extruder, it will match in the filament colour and set those as colours in the placeholder.

PR #629, opened 2025-02-09
Fix calculation of purge volumes from the slicer with only 2 gates

swstim:
When parsing tool map / purge volumes from the slicer, prefer NxN over from->to list. This fixes invalid purge table when there are only 2 gates.


moggieuk:
Nice one. Didn't think of 2 extruder case!

PR #634, opened 2025-02-11
fix: make safe travel move dependent on cutting axis

PocketMiner82:
If we need to do a safe travel move, first the cutting axis should be moved and then the other one. I have my cutter on the y axis and moving x first would cause the lever to crash into the pin.


moggieuk:
Thanks. This mirrors the logic in the main cut macro so makes sense..

Issue #275, opened 2024-04-19
Belay sensor never triggers on printer restart and first print


igiannakas:
There is a pretty common case where the Annex tension belay sensor never triggers when the printer is restarted until the sensor is triggered manually (or through luck) for the first time.

This is especially problematic with the Annex belay as its a tension feedback mechanism. When the filament is loaded for the first time it usually results in a compression state, resulting in the switch never triggering during the print.

This happens because after a reboot the filament is expanded (sensor not triggered) so it initialises it as neutral as the pin has never triggered for it to update its state. Therefore it uses multiplier 1. 

This will result in the filament possibly always expanding if your extruder is ever so slightly under extruding compared to the ercf. 

Just woken up to an overnight print with the belay fully expanded with plenty of tension in the filament. Checking the logs it appears that the sensor never triggered. It started working when I pushed the lever manually.

**Log file:** 
[mmu_2.log](https://github.com/moggieuk/Happy-Hare/files/15038336/mmu_2.log)

I believe this is caused due to the way the sensor is initialised. 

Looking at the below it is set initially to neutral:
![image](https://github.com/moggieuk/Happy-Hare/assets/59056762/c8737ffc-6b3a-4cfe-b65d-6f379e0df109)

Then it is initialised as neutral:
![image](https://github.com/moggieuk/Happy-Hare/assets/59056762/8d7d21b6-019d-43b7-a46f-9d223b75cdee)

However when the filament is loaded into the extruder, the sensor will expand as the filament is pushed on the extruder gears creating some tension in the filament path. If your extruder is not over extruding compared to the ERCF motor then it will never trigger the tension sensor. Worse, if it is slightly under extruding it will keep raising compression in the filament. 

If you have a dual sensor this situation can never occur. Also if you have a compression sensor, it is far less likely to happen as the filament will most likely result in filament compression on initial tool load triggering the sensor.

Ideally, to correct this and make it more reliable, the initialisation routine above needs to check what type of sensor is fitted. If:
1. Tension sensor: Set the sync_feedback_last_state to 1, triggering the sensor to start contracting the filament until the switch is hit.
2. Compression sensor: Set the sync_feedback_last_state to -1, triggering the sensor to start expanding the filament until the switch is triggered
3. Dual sensor or proportional sensor: Leave it to 0 and let the sensors do their thing on first update.

[...]

PR #636, opened 2025-02-13
Sync mechanism fixing and/or enhancing 

CooperGerman:
Hi @moggieuk ,

It's me at it again =) I'm currently experiencing issues regarding the sync_feedback feature and whilst reading through the implementation i noticed some things.
- I might have stumbled upon some bugs see commit 944bbbdc2e67797799005c1ae48e0a6e1a8d40d8. Is this intended ? It seemed weird you would reset the direction to state but it might not have consequences.
- Another strange thing here see commit 0b9695eb236fd27cef46096541ca3d5463661fe1. I feels like the `multiplier = 1` should more depend on the state rather than the direction ?
- In the sensor description, i might not have understood this correctly but it seems we should check for filament_present additionally to enabled ? (see commit 4d92abd8de5790bf3ddf8da1f027178336216b08)
 
Additionally I'm experiencing issues on my compression+tension feedback sensor setup (hence the deep dive into the code) and found that in my case having compression and tension detected simultaneously is actually not an error but more a "buggy" neutral position. In some cases i felt like the some "tension" or "compression" button events wouldn't result in a multiplier change (not sure why but maybe when both are triggered ?).

My current fix (might not be final) is to use the timer you coded and additionally to direction changes check for state changes. I should not have to do this but this would ensure the code would auto-fix it's state periodically see bc949a1c0a1d7fcc0fe66b7a03c1f68f2321eb62.

Let me get your thoughts on this matter whenever you are available. I'll keep on trying to fix the "intended" event based sync_feedback.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -2793,6 +2793,8 @@ class Mmu:
             self.sync_feedback_last_state = float(state)
             if self.sync_feedback_enable and self.sync_feedback_operational:
                 self._update_sync_multiplier()
+        else :
+            self.log_error("Invalid sync feedback state: %s" % state)
 
     def _handle_mmu_synced(self):
         if not self.is_enabled: return
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -2834,20 +2834,12 @@ class Mmu:
             pos = extruder.find_past_position(estimated_print_time)
             past_pos = extruder.find_past_position(max(0., estimated_print_time - self.SYNC_POSITION_TIMERANGE))
             if abs(pos - past_pos) >= self.SYNC_POSITION_MIN_DELTA:
-                # change in direction
                 prev_direction = self.sync_feedback_last_direction
                 self.sync_feedback_last_direction = self.DIRECTION_LOAD if pos > past_pos else self.DIRECTION_UNLOAD if pos < past_pos else 0
                 if self.sync_feedback_last_direction != prev_direction:
                     d = self.sync_feedback_last_direction
                     self.log_trace("New sync direction: %s" % ('extrude' if d == self.DIRECTION_LOAD else 'retract' if d == self.DIRECTION_UNLOAD else 'static'))
                     self._update_sync_multiplier()
-                # change in state
-                prev_state = self.sync_feedback_last_state
-                self.sync_feedback_last_state = self._get_current_sync_state()
-                if self.sync_feedback_last_state != prev_state:
-                    self.log_trace("New sync state: %s" % self._get_sync_feedback_string())
-                    self._update_sync_multiplier()
-
         return eventtime + self.SYNC_FEEDBACK_INTERVAL
 
     def _update_sync_multiplier(self):
@@ -4206,7 +4198,7 @@ class Mmu:
                 self._set_filament_direction(self.DIRECTION_LOAD)
                 self.selector.filament_drive()
 
-                # Record starting position for bowden progress tracking 
+                # Record starting position for bowden progress tracking
                 self.bowden_start_pos = self.get_encoder_distance(dwell=None) - start_pos
 
                 if self.gate_selected > 0 and self.rotation_distances[self.gate_selected] <= 0:
@@ -4240,7 +4232,7 @@ class Mmu:
                             self.log_info("Warning: Excess slippage was detected in bowden tube load afer correction moves. Gear moved %.1fmm, Encoder measured %.1fmm. See mmu.log for more details"% (length, length - delta))
                     else:
                         self.log_info("Warning: Excess slippage was detected in bowden tube load but 'bowden_apply_correction' is disabled. Gear moved %.1fmm, Encoder measured %.1fmm. See mmu.log for more details" % (length, length - delta))
-    
+
                     if delta >= self.bowden_allowable_load_delta:
                         self.log_debug("Possible causes of slippage:\nCalibration ref length too long (hitting extruder gear before homing)\nCalibration ratio for gate is not accurate\nMMU gears are not properly gripping filament\nEncoder reading is inaccurate\nFaulty servo")
 
@@ -4277,7 +4269,7 @@ class Mmu:
                 self._set_filament_direction(self.DIRECTION_UNLOAD)
                 self.selector.filament_drive()
 
-                # Record starting position for bowden progress tracking 
+                # Record starting position for bowden progress tracking
                 self.bowden_start_pos = self.get_encoder_distance(dwell=None)
 
                 # Optional pre-unload safety step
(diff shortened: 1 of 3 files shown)


CooperGerman:
Added A stress test for sync state validation.

`_MMU_TEST SYNC_STATE=loop LOOP=100000`

and will result in :

image

The test randomly triggers compressions and tensions and saves all results, test configurations and expected results in order to check for configurations that would be more prone to generating errors.

Let me know if you need to discuss ;)

moggieuk:
Nice work!   On a superficial level I think you are completely correct on the first couple of fixes .. I must have been having a bad day when I made those changes (frankly too much switching between UI and backend!!) and the fact that my MMU's have been pulled apart to test certain configurations...

I'm looking at the other suggestions more closely...

moggieuk:
Thinking more about a better fix for #3.  What about this logic (untested because I'm not with my printer):
```yml
    # Button event handlers for sync-feedback
    # Feedback state should be between -1 (expanded) and 1 (compressed)
    def _sync_tension_callback(self, eventtime, tension_state, runout_helper):
        from .mmu import Mmu # For sensor names
        tension_enabled = runout_helper.sensor_enabled
        compression_sensor = self.printer.lookup_object("filament_switch_sensor %s_sensor" % Mmu.SENSOR_COMPRESSION, None)
        has_active_compression = compression_sensor.runout_helper.sensor_enabled if compression_sensor else False
        compression_state = compression_sensor.runout_helper.filament_present if has_active_compression else False

        if tension_enabled:
            if has_active_compression:
                if compression_state:
                    logging.info("Malfunction of sync-feedback unit: both tension and compression sensors are active at the same time!")
                    self.printer.send_event("mmu:sync_feedback", eventtime, 0) # neutral for malfunction
                else:
                    self.printer.send_event("mmu:sync_feedback", eventtime, -(tension_state * 2 - 1)) # -1 or 1
            else:
                self.printer.send_event("mmu:sync_feedback", eventtime, -tension_state) # -1 or 0 (neutral)
        else:
            self.printer.send_event("mmu:sync_feedback", eventtime, 0) # neutral

    def _sync_compression_callback(self, eventtime, compression_state, runout_helper):
        from .mmu import Mmu # For sensor names
        compression_enabled = runout_helper.sensor_enabled
        tension_sensor = self.printer.lookup_object("filament_switch_sensor %s_sensor" % Mmu.SENSOR_TENSION, None)
        has_active_tension = tension_sensor.runout_helper.sensor_enabled if tension_sensor else False
        tension_state = tension_sensor.runout_helper.filament_present if has_active_tension else False

        if compression_enabled:
            if has_active_tension:
                if tension_state:
                    logging.info("Malfunction of sync-feedback unit: both tension and compression sensors are active at the same time!")
                    self.printer.send_event("mmu:sync_feedback", eventtime, 0) # neutral for malfunction
                else:
                    self.printer.send_event("mmu:sync_feedback", eventtime, compression_state * 2 - 1) # 1 or -1
            else:
                self.printer.send_event("mmu:sync_feedback", eventtime, compression_state) # 1 or 0 (neutral)
        else:
            self.printer.send_event("mmu:sync_feedback", eventtime, 0) # neutral
```

CooperGerman:
@moggieuk Here we are!

No more bugs (according to the added soak test). There were some strange falling edge cases on the sensors that are now properly handled.

The "fix" timer has been removed and soak testing enhanced.

I also added an attribute (`_is_running_test`) to the mmu class in order flag some test specific actions. In my case the event generation in the `_handle_sync_feedback` method.

I consider this PR as ready, tell me if anything bothers you.

igiannakas:
Quick check - would these changes affect when a single sensor is present (compression / tension?)? There was a bug on this a while ago so just checking we're not about to introduce a regression? https://github.com/moggieuk/Happy-Hare/issues/275

CooperGerman:
I took a look at your issue and understand your worries as i remember actually experiencing the same kind of issues when i was running the stock belay.

If i understand correctly i would have to check that at wake up the system is in the correct correction state according to the sensor (whether it is tension or compression sensor) right ? I'm not sure i can add a specific test in the `_MMU_TEST` suite for this as it would require restarting the firmware each time... I can try the different cases by hand though =)

The "state" value of compression/tension should be sampled once at startup right ? I think that the code here still acts upon your issue:
```py
def handle_ready(self):
       ...
       ...

        # Sub components
        self.selector.handle_ready()

        # Ensure sync_feedback starting state. This is mainly cosmetic because state is ensured when enabling
        self._update_sync_starting_state()

        # Schedule bootup tasks to run after klipper and hopefully spoolman have settled
        self._schedule_mmu_bootup_tasks(self.BOOT_DELAY)
```

ksummers92:
@CooperGerman, unrelated to this PR, do you mind sharing your sensor design? I was planning to make something similar, but you already did! If you have a step file, I would appreciate it. Thanks in advanced.

CooperGerman:
> @CooperGerman, unrelated to this PR, do you mind sharing your sensor design? I was planning to make something similar, but you already did! If you have a step file, I would appreciate it. Thanks in advanced.

Ill put up a git repo with the project 

> @CooperGerman, unrelated to this PR, do you mind sharing your sensor design? I was planning to make something similar, but you already did! If you have a step file, I would appreciate it. Thanks in advanced.

You can find the step file here : https://github.com/print-hive/Filament_Sensor_Unit

I'll be adding the Freecad Project as soon as it is cleaned up a bit. Please note that it is not in a release state and several tolerances and/or quotes might be to update according to your needs ;)

igiannakas:
> I took a look at your issue and understand your worries as i remember actually experiencing the same kind of issues when i was running the stock belay.
> 
> If i understand correctly i would have to check that at wake up the system is in the correct correction state according to the sensor (whether it is tension or compression sensor) right ? I'm not sure i can add a specific test in the `_MMU_TEST` suite for this as it would require restarting the firmware each time... I can try the different cases by hand though =)
> 
> The "state" value of compression/tension should be sampled once at startup right ? I think that the code here still acts upon your issue:
> 
> ```python
> def handle_ready(self):
>        ...
>        ...
> 
>         # Sub components
>         self.selector.handle_ready()
> 
>         # Ensure sync_feedback starting state. This is mainly cosmetic because state is ensured when enabling
>         self._update_sync_starting_state() #<<<<<<<<<<<<<<<<< Here
> 
>         # Schedule bootup tasks to run after klipper and hopefully spoolman have settled
>         self._schedule_mmu_bootup_tasks(self.BOOT_DELAY)
> ```
> 
> Addendum : I see that the code also updates the state in the following cases:
> 
> * when a sync `"mmu:synced"` occurs
> * when a gate gets selected via the `_set_gate_selected` method
> 
> To my best knowledge i would imagine that it would be enough to ensure the state is correct at printer bootup.

Indeed that looks correct :) 👍🏻

PR #640, opened 2025-02-17
feat: support cutter to be placed before extruder

PocketMiner82:
I have my filament cutter placed before my extruder (BIQU H2 500°C) because there is no other way to place the cutter. I needed to modify Happy-Hare a little to make it work.
As I'm still testing this PR will be a draft until I'm sure that it works reliably. I tried to make sure it's backwards compatible but I'm not familiar enough with the code to be sure that it is. So that needs to be reviewed carefully.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -3855,7 +3855,7 @@ class Mmu:
                     # Mimick in print if requested
                     self.sync_gear_to_extruder(self.sync_form_tip and self.is_in_print(force_in_print), grip=True, current=self.is_in_print(force_in_print))
 
-                    _,_,_ = self._do_form_tip(test=True)
+                    _,_,_,_ = self._do_form_tip(test=True)
                     self._set_filament_pos_state(self.FILAMENT_POS_UNLOADED)
 
         except MmuError as ee:
@@ -4425,7 +4425,7 @@ class Mmu:
             self._set_filament_remaining(0.)
 
             # Encoder based validation test if short of deterministic sensors and test makes sense
-            if self._can_use_encoder() and not fhomed and not extruder_only and self.gate_selected != self.TOOL_GATE_BYPASS:
+            if length > self.encoder_min and self._can_use_encoder() and not fhomed and not extruder_only and self.gate_selected != self.TOOL_GATE_BYPASS:
                 self.log_debug("Total measured movement: %.1fmm, total delta: %.1fmm" % (measured, delta))
                 if measured < self.encoder_min:
                     raise MmuError("Move to nozzle failed (encoder didn't sense any movement). Extruder may not have picked up filament or filament did not find homing sensor")
@@ -4932,7 +4932,7 @@ class Mmu:
 
             # Perform the tip forming move and establish park_pos
             initial_encoder_position = self.get_encoder_distance()
-            park_pos, remaining, reported = self._do_form_tip()
+            park_pos, remaining, reported, before_extruder = self._do_form_tip()
             measured = self.get_encoder_distance(dwell=None) - initial_encoder_position
             self._set_filament_remaining(remaining, self.gate_color[self.gate_selected] if self.gate_selected != self.TOOL_GATE_UNKNOWN else '')
 
@@ -4957,7 +4957,11 @@ class Mmu:
             self._set_filament_position(-park_pos)
             self.set_encoder_distance(initial_encoder_position + park_pos)
 
-            if detected or extruder_only:
+            if reported and before_extruder:
+                # The cutter is located before the extruder.
+                # Therefore the filament is definitely somewhere before the extruder entry
+                self._set_filament_pos_state(self.FILAMENT_POS_HOMED_EXTRUDER)
+            elif detected or extruder_only:
                 # Definitely in extruder
                 self._set_filament_pos_state(self.FILAMENT_POS_IN_EXTRUDER)
             else:
@@ -4974,7 +4978,7 @@ class Mmu:
             with self._wrap_pressure_advance(0., "for tip forming"):
                 gcode_macro = self.printer.lookup_object("gcode_macro %s" % self.form_tip_macro, "_MMU_FORM_TIP")
                 self.log_info("Forming tip...")
-                self.wrap_gcode_command("%s %s" % (self.form_tip_macro, "FINAL_EJECT=1" if test else ""), exception=True, wait=True)
+                self.wrap_gcode_command("%s %s %s %s" % (self.form_tip_macro, f"TOOLHEAD_EXTRUDER_TO_NOZZLE={self.toolhead_extruder_to_nozzle}", "SYNC=1" if self.mmu_toolhead.is_gear_synced_to_extruder() else "SYNC=0", "FINAL_EJECT=1" if test else ""), exception=True, wait=True)
 
             final_mcu_pos = self.mmu_extruder_stepper.stepper.get_mcu_position()
             stepper_movement = (initial_mcu_pos - final_mcu_pos) * self.mmu_extruder_stepper.stepper.get_step_dist()
@@ -4986,6 +4990,8 @@ class Mmu:
                 self.log_error("Reported 'output_park_pos: %s' could not be parsed: %s" % (park_pos, str(e)))
                 park_pos = -1
 
+            # assume we are in extruder
+            before_extruder = False
             if park_pos < 0:
                 # Use stepper movement
                 reported = False
@@ -5006,12 +5012,16 @@ class Mmu:
                 else:
                     self.log_trace(msg)
 
-            if not test and park_pos > self.toolhead_extruder_to_nozzle:
+            if reported and park_pos > self.toolhead_extruder_to_nozzle:
+                self.log_debug("park_pos (%.1fmm) is greater than 'toolhead_extruder_to_nozzle' distance of %.1fmm! Assuming cutter is located before extruder." % (park_pos, self.toolhead_extruder_to_nozzle))
+                before_extruder = True
+                filament_remaining = max(0., filament_remaining)
+            elif not test and park_pos > self.toolhead_extruder_to_nozzle:
                 self.log_error("Warning: park_pos (%.1fmm) cannot be greater than 'toolhead_extruder_to_nozzle' distance of %.1fmm! Assumming fully unloaded from extruder\nWill attempt to continue..." % (park_pos, self.toolhead_extruder_to_nozzle))
                 park_pos = self.toolhead_extruder_to_nozzle
                 filament_remaining = 0.
 
-        return park_pos, filament_remaining, reported
+        return park_pos, filament_remaining, reported, before_extruder
 
 
 #################################
--- a/config/base/mmu_cut_tip.cfg
+++ b/config/base/mmu_cut_tip.cfg
@@ -39,6 +39,8 @@ description: Cut filament by pressing the cutter on a pin with a horizontal move
 variable_output_park_pos: 0 # Dynamically set in this macro
 
 gcode:
+    {% set toolhead_extruder_to_nozzle = params.TOOLHEAD_EXTRUDER_TO_NOZZLE|default(-1)|float %}
+    {% set sync = params.SYNC|default(0)|int%}
     {% set final_eject = params.FINAL_EJECT|default(0)|int %}
     {% set vars = printer['gcode_macro _MMU_CUT_TIP_VARS'] %}
     {% set park_vars = printer['gcode_macro _MMU_PARK'] %}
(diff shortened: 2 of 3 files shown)


moggieuk:
Thanks for this.  I'm going to have to concentrate to review this and need to be with my printer (I'm not at the moment).  I hope to get time to look at it next week...

moggieuk:
Sorry, still side-tracked with Mainsail / Fluidd...   will get to this soon!

extras/mmu/mmu.py
```diff
@@ -4974,7 +4978,7 @@ def _do_form_tip(self, test=False):
             with self._wrap_pressure_advance(0., "for tip forming"):
                 gcode_macro = self.printer.lookup_object("gcode_macro %s" % self.form_tip_macro, "_MMU_FORM_TIP")
                 self.log_info("Forming tip...")
-                self.wrap_gcode_command("%s %s" % (self.form_tip_macro, "FINAL_EJECT=1" if test else ""), exception=True, wait=True)
+                self.wrap_gcode_command("%s %s %s %s" % (self.form_tip_macro, f"TOOLHEAD_EXTRUDER_TO_NOZZLE={self.toolhead_extruder_to_nozzle}", "SYNC=1" if self.mmu_toolhead.is_gear_synced_to_extruder() else "SYNC=0", "FINAL_EJECT=1" if test else ""), exception=True, wait=True)
```
PocketMiner82, line 4997: I don't know if there is a better way of accessing the current sync state and extruder to nozzle distance than passing those values as parameters

extras/mmu/mmu.py
```diff
@@ -4957,7 +4957,11 @@ def form_tip_standalone(self, extruder_only=False):
             self._set_filament_position(-park_pos)
             self.set_encoder_distance(initial_encoder_position + park_pos)
 
-            if detected or extruder_only:
+            if reported and before_extruder:
+                # The cutter is located before the extruder.
+                # Therefore the filament is definitely somewhere before the extruder entry
+                self._set_filament_pos_state(self.FILAMENT_POS_HOMED_EXTRUDER)
```
PocketMiner82, line 4979: I didn't find a state that describes "between extruder and entry sensor", so I believe this is the best we can do. Any other ideas to track filament pos after cut?

extras/mmu/mmu.py
```diff
@@ -5006,12 +5012,16 @@ def _do_form_tip(self, test=False):
                 else:
                     self.log_trace(msg)
 
-            if not test and park_pos > self.toolhead_extruder_to_nozzle:
+            if reported and park_pos > self.toolhead_extruder_to_nozzle:
```
PocketMiner82, line 5031: Maybe this should also rather check the blade pos rather than the park pos (like I did in the macro)

PocketMiner82 (commented):
> Sorry, still side-tracked with Mainsail / Fluidd... will get to this soon!

Nice! Looking forward to that integration. I finished my first two color print with 67 swaps and it worked reliably (just the cotton tail tangled the filament sometimes, will look into building some respoolers, lol).

I'll mark this as ready for review. I also self-reviewed the changes again and added comments where I'm not sure if my implementation is the best/correct.

extras/mmu/mmu.py
```diff
@@ -5006,12 +5012,16 @@ def _do_form_tip(self, test=False):
...
+                self.log_debug("park_pos (%.1fmm) is greater than 'toolhead_extruder_to_nozzle' distance of %.1fmm! Assuming cutter is located before extruder." % (park_pos, self.toolhead_extruder_to_nozzle))
+                before_extruder = True
+                filament_remaining = max(0., filament_remaining)
```
PocketMiner82, line 5034: I'm also not sure if that calculation for the remaining filament is correct. It worked for me, but I'm not sure if this will work if the cutter is located further away from the extruder entry.

extras/mmu/mmu.py
moggieuk: In the macro you can use printer variables:
printer.mmu.sync_drive = True|False

Similarly the toolhead_extruder_to_nozzle can be pulled:
printer.configfile.config.mmu.toolhead_extruder_to_nozzle : 72
or better still is the processed one:
printer.configfile.settings.mmu.toolhead_extruder_to_nozzle : 72.0

This is really what the macro should do rather than to overload parameter passing I think.

extras/mmu/mmu.py
moggieuk: After a cut the filament is "free" for fast bowden move so state of "FILAMENT_POS_END_BOWDEN" is probably the best... because the cutting would not be exactly at the extruder gears.  The next state lower is "IN_BOWDEN" and that would work because HH would fallback to slow unload because of unsure of position.

extras/mmu/mmu.py
moggieuk: I still need to review the whole change (just running through these comments first).  What is important is to (i) get the position of the end of the filament, (ii) determine the amount of cut filament left behind so that the purging logic knows how much to purge.

HH uses the reported `output_park_pos` from the macro and the aggregate measured extruder movement during tip forming/cutting operation. 

So really it should only be necessary to set `output_park_pos`.  This would be a value larger than the toolhead_to_nozzle dimension (so we would need to make sure there is no safety logic that prevents this).  Since HH would have measured the total retraction it will subtract the retraction from the `output_park_pos` to calculate the "filament_remaining"

PocketMiner82:
Hey, sorry for the late reply but I'm currently very busy and also not with my printer. I hope I might be able to take a look at your comments tomorrow.

extras/mmu/mmu.py
PocketMiner82: Done!

extras/mmu/mmu.py
PocketMiner82: Yes, that makes more sense. Updated!

extras/mmu/mmu.py
PocketMiner82: currently testing if removing the call to max works reliably, converted the PR to draft again because of that

PocketMiner82:
@moggieuk I adressed most of the review comments and now tested if everything still works, which - until now - seems to be the case. So I'm marking the PR as ready for review again.

moggieuk:
Thanks!  Sorry, I've been overwhelmed with work lately.   I'l get to looking at pulling this PR very soon..

acunninghamks:
I am fairly certain there are issues being encountered with this during the retract then push that's part of unload and load. The behavior I am experiencing is that the cut filament, which is subsequently pushed back into the nozzle, doesn't end up leaving the extruder gears. The reload ends up getting to the extruder gears, but since the gear + extruder movement ends up not being synced, there is a failure to have the new filament reach the heat break. I know that most of this may be complications due to the geometry of my setup, but I am wanting to try and get some ideas here before coming up with workarounds...

Here is a rough drawing of my setup. Dimensioning all the way to the nozzle tip, but obviously would be adjusted for the shoulder.

image

The heat break ends at 37 or so mm above the nozzle tip.

After unload: 
image


Due to the nature of having the cut occur above the extruder. I end up with filament 33.5mm above plus 14.6mm below the extruder. 

Once the filament goes to load, the extruder pushes 30mm back into the heat break. This results in 3.5mm still remaining above the extruder itself. 

From that point on, there is no additional synced move for loading, which results in the filament from above never actually getting to the extruder gears, but there still being a small portion of the remnant which can be extruded.

I suppose I could fiddle around a bit with trying to get the retract even closer to the 37mm at the top of the heat break, but I would like to avoid clogging issues. I know the geometry works out a bit goofy since my distance between the cutter and extruder isn't much smaller than the heat break position (33.5 vs 37).

acunninghamks:
Actually did just find this on the discord. 

> I am fairly certain there are issues being encountered with this during the retract then push that's part of unload and load. The behavior I am experiencing is that the cut filament, which is subsequently pushed back into the nozzle, doesn't end up leaving the extruder gears. The reload ends up getting to the extruder gears, but since the gear + extruder movement ends up not being synced, there is a failure to have the new filament reach the heat break. I know that most of this may be complications due to the geometry of my setup, but I am wanting to try and get some ideas here before coming up with workarounds...
> 
> Here is a rough drawing of my setup. Dimensioning all the way to the nozzle tip, but obviously would be adjusted for the shoulder.
> 
> image
> 
> The heat break ends at 37 or so mm above the nozzle tip.
> 
> After unload: image
> 
> Due to the nature of having the cut occur above the extruder. I end up with filament 33.5mm above plus 14.6mm below the extruder.
> 
> Once the filament goes to load, the extruder pushes 30mm back into the heat break. This results in 3.5mm still remaining above the extruder itself.
> 
> From that point on, there is no additional synced move for loading, which results in the filament from above never actually getting to the extruder gears, but there still being a small portion of the remnant which can be extruded.
> 
> I suppose I could fiddle around a bit with trying to get the retract even closer to the 37mm at the top of the heat break, but I would like to avoid clogging issues. I know the geometry works out a bit goofy since my distance between the cutter and extruder isn't much smaller than the heat break position (33.5 vs 37).

Actually did just find this on the discord

_<@233360208905240577>  <@233438283152097281> This is what I think needs to be done to correctly support cutting outside of the extruder gears...  it's actually quite tricky
0) need to allow for reported output_park_pos outside of extruder and trigger new logic:
1) cut_tip_macro MUST perform initial_retract.  The retract distance MUST BE > distance from blade to extruder entrance and cut portion (which is dependent on other things) MUST BE > distance from blade to extruder entrance.
2) extra logic is needed to push the cut piece  back into the extruder entrance using just the extruder motor (the previous constraints are to ensure this is possible)
3) bypass the current final extruder unload logic
4) for the succeeding bowden unload the expected distance needs to be shortened by the distance from blade to extruder entrance (can be best derived from output_park_pos)
5) the loading strategy should home to the extruder (a few possible homing options - "extruder", "collision"  with encoder, "gear_touch" or "none")
6) cannot make use of toolhead sensor (for loading/unloading). In fact there isn't really any value toolhead sensor in this setup._

PocketMiner82:
Hi @acunninghamks, do you have `sync_to_extruder` and/or `sync_form_tip enabled`? I only tested with both enabled, so to solve your problem, I might need to force sync during load.
Btw, I also updated the branch to the latest main.

PocketMiner82:
> 0. need to allow for reported output_park_pos outside of extruder and trigger new logic:
> 1. cut_tip_macro MUST perform initial_retract.  The retract distance MUST BE > distance from blade to extruder entrance and cut portion (which is dependent on other things) MUST BE > distance from blade to extruder entrance.
> 2. extra logic is needed to push the cut piece  back into the extruder entrance using just the extruder motor (the previous constraints are to ensure this is possible)
> 3. bypass the current final extruder unload logic
> 4. for the succeeding bowden unload the expected distance needs to be shortened by the distance from blade to extruder entrance (can be best derived from output_park_pos)
> 5. the loading strategy should home to the extruder (a few possible homing options - "extruder", "collision"  with encoder, "gear_touch" or "none")
> 6. cannot make use of toolhead sensor (for loading/unloading). In fact there isn't really any value toolhead sensor in this setup._

I also think I adressed everything here, except 4. and 6. For 4. I thought that updating the park_pos is enough, will look into that when I'm with my printer again. And I didn't think about 6. at all (I don't have a toolhead sensor).
And I'm not sure that in 1. the retract distance really has to be bigger than blade to extruder distance because if we use synced movement when loading and purging and have all calculations correct, then we don't need to worry about the filament still being in the extruder gears after unretract. The remaining filament will get purged anyway, so it also wouldn't be a problem in print resuming later. So the retract can be shorter than blade to extruder distance, if I didn't overlook something... :P

<br>

Edit: I thought about the retract distance again. Only when one is not using sync during **purging** then the retract distance must be > distance from blade to extruder. If one is using sync during purging, all the filament will end up getting purged. It would then even be possible to turn off sync again, after purging.

PocketMiner82:
> I do not have sync_to_extruder enabled, but I do have sync_form_tip enabled. I wanted to ensure things were all working properly before turning that on, but it does seem it would be an easy fix to turn it on. I do think that forcing synced movement during loading would be the best approach, but I will give it a try with sync_to_extruder for now.
> 
> Thank you for taking the time to look through this.

No problem! Could you please respond back, if turning on sync_to_extruder fixed your problem? because if it didn't, forcing sync during load probably won't work either... thanks in advance!

acunninghamks:
@PocketMiner82 I think I need a little bit better understanding of how sync_to_extruder actually functions during a print.

Example, I unload and then load filament. After that I trigger an extrusion move by clicking the EXTRUDE button in Mainsail. The result is below. What I think I am missing though is that the MMU gear does not turn when the extrude command is issued. It is locked solid which causes the extruder to just strip the filament instead of loading further into the heatbreak. I feel like I missed something fundamental with how this should work. 

```
SAVE_GCODE_STATE NAME=_ui_extrude
M83
G1 E50 F300
RESTORE_GCODE_STATE NAME=_ui_extrude```

acunninghamks:
@PocketMiner82 after further experimentation I am still in the middle of tweaking my settings. I've ended up having to babysit prints as every 20th or so toolchange ended up not properly clearing the extruder entry sensor. I was retracting 30mm and them pushing back 30mm after the cut, but I think I was ending up with pulling out too close to the end of the heatbreak and causing clogs. 

Another setting I am now playing with is toolhead_unload_safety_margin. I am suspecting it may have been giving me trouble and retracting filament the wrong direction after an unload.

The thing that solved the majority of my issues was turning on the syncing between gear and extruder through the variables along with realizing I need to issue the command manually if doing any kind of extrusion through the UI.

acunninghamks:
Set the toolhead_unload_safety_margin to 0. Just completed a few hour print with 54 tool changes without any issues at all.

acunninghamks:
After some further thought and observation, I think the only issue I have now is that the amount I am able to retract out of the heartbreak is marginally larger than the distance from my extruder entry sensor to extruder. The majority of tool changes result in the sensor being cleared but every now and then it does not. I think I’m going to need to bump the retract/push by a few mm and try again.

PocketMiner82:
@acunninghamks just to make sure: is your entry sensor located before or after the filament cutter?

Like
1. Bowden -> entry sensor -> cutter -> extruder

or

2. Bowden -> cutter -> entry sensor -> extruder (not implemented by this PR)

If you have option 1 then retracted filament \*should\* never end up in the entry sensor, if you have the correct pushback. If it is option 2 then maybe try if it is possible to move the entry sensor so that you have something like option 1.

acunninghamks:
@PocketMiner82 

entry sensor is below cutter. Refer to the images I shared above. It is a very specific tool head, smart orbiter 3, so moving the sensor is not really an option. I could of course add one above with some redesign but I would prefer to not complicate this even further.

PocketMiner82:
Ah, okay. I somehow overlooked the dimensions in the images. Yeah, if you're using option 2, it seems to be a bit more complicated to calibrate those values, good luck!

Edit: I will probably add a comment somewhere in the config that explains/summarizes all we have discussed

madsi1m:
I also have this layout and would like to see this merged.

Bowden -> entry sensor -> cutter -> extruder

moggieuk:
Update: Thanks and I'm definitely interested in this.  With all the HH v4 work I'm just looking for the right time to verify and pull.  Please be patient...

Wazzup77:
@moggieuk would love to see this pulled, this is another Qidi quirk that we are currently working around.
Not sure what the consensus here was, I can work this into my fork and test if needed

moggieuk:
Marked as duplicate because of PR#904 work

PocketMiner82:
Closed in favor of #904

PR #646, opened 2025-02-21
fix mmu_server.py to work with recent update to moonrakers metdata.py

### LINKED ISSUE #641: unable to fetch metadata after using the new klipper_estimator feature in moonraker
there is a bug where the mmu_server.py is unable to fetch the metadata after you used the new klipper_estimator feature in moonraker. moonraker.log is attached; at the end of the file you can see the errors produced by starting a time analysis

[moonraker(1).log](https://github.com/user-attachments/files/18870582/moonraker.1.log)

fabiandj64:
added the new -c argument and edited the main function calls to work with the new klipper_estimator feature provided by moonraker

should fix #641

### DIFF
--- a/components/mmu_server.py
+++ b/components/mmu_server.py
@@ -1079,7 +1079,7 @@ if __name__ == "__main__":
     config: Dict[str, Any] = {}
     if args.config is None:
         if args.filename is None:
-            logger.info(
+            metadata.logger.info(
                 "The '--filename' (-f) option must be specified when "
                 " --config is not set"
             )
@@ -1094,10 +1094,10 @@ if __name__ == "__main__":
             with open(args.config, "r") as f:
                 config = (json.load(f))
         except Exception:
-            logger.info(traceback.format_exc())
+            metadata.logger.info(traceback.format_exc())
             sys.exit(-1)
         if config.get("filename") is None:
-            logger.info("The 'filename' field must be present in the configuration")
+            metadata.logger.info("The 'filename' field must be present in the configuration")
             sys.exit(-1)
     if config.get("gcode_dir") is None:
         config["gcode_dir"] = os.path.abspath(os.path.dirname(__file__))
--- a/components/mmu_server.py
+++ b/components/mmu_server.py
@@ -1068,6 +1068,7 @@ if __name__ == "__main__":
 
     # Parse start arguments (copied from metadata.py)
     parser = argparse.ArgumentParser(description="GCode Metadata Extraction Utility")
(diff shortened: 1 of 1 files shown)


components/mmu_server.py
```diff
@@ -1068,19 +1068,44 @@ def main(path, filename, insert_placeholders=False, insert_nextpos=False):
...
     parser.add_argument("-o", "--check-objects", dest='check_objects', action='store_true', help="process gcode file for exclude object functionality")
     parser.add_argument("-m", "--placeholders", dest='placeholders', action='store_true', help="process happy hare mmu placeholders")
     parser.add_argument("-n", "--nextpos", dest='nextpos', action='store_true', help="add next position to tool change")
     args = parser.parse_args()
-    check_objects = args.check_objects
-    enabled_msg = "enabled" if check_objects else "disabled"
+    config: Dict[str, Any] = {}
+    if args.config is None:
+        if args.filename is None:
+            logger.info(
```
skyguy94, line 1082: I don't think there's a `logger` in scope here. Should these statements us `metadata.logger`?

jonwally:
can confirm this update fixes issues caused by a recent moonraker update

skyguy94 (commented):
I started out with a similar change, but after reviewing this moonraker commit (https://github.com/Arksine/moonraker/commit/b8b28bc0c969aed2d03964866f50077e0d30d9e8), I took a stab at registering mmu_server as a metadata processor and removing the file_manager chaining logic.

One can register a metadata processor in init like this:
```python
self.proc_config: Dict[str, Any] = {
    "name": "mmu_server",
    "command": [
        "python",
        str(self.path),
        "-f",
        "{gcode_file_path}"
    ],
    "timeout": 300,
    "version": self.version,
    "enable_toolchange_next_pos": self.config.getboolean("enable_toolchange_next_pos", True),
    "ident": {
        "regex": MMU_REGEX,
        "location": "header"
    },
    "enabled": self.config.getboolean("enable_file_preprocessor", True)
}

mdst = self.file_manger.get_metadata_storage()
mdst.register_gcode_processor("mmu_server", self.proc_config)

```

But given my lack of familiarity with all things python, there's a couple of awkward parts. The first is invoking `python` to run mmu_server. I'm not sure the best way to invoke it or even if splitting out the moonraker component from the logic that does the preprocessing might make more sense.

The other part is that `Timeout` has a value, but I'd rather rely in the default in file_manager and I'm not sure how to acquire that value.

components/mmu_server.py
fabiandj64: thanks for the heads up; added metadata now.

PR #648, opened 2025-02-22
Update `mmu_server.py` to work with moonraker's metadata processor logic.

skyguy94:
This PR updates mmu_server.py to work with recent changes to moonraker's metadata gcode file processing that allows for gcode file processors that can be added during initialization.  This change implements moonraker initialization hook to register the mmu gcode processor with moonraker. Registration enables the logic in moonraker's metadata.py to invoke the mmu gcode processor and mmu_server.py no longer needs to hook into file_manager in a fragile way to be able to process the gcode file. 

When `mmu_server.py` is loaded, the following is shown in `moonraker.log`
```
2025-02-22 22:58:18,123 [file_manager.py:register_gcode_processor()] - GCode Processor mmu_server registered
2025-02-22 22:58:18,123 [server.py:load_component()] - Component (mmu_server) loaded
...
2025-02-22 22:58:18,480 [server.py:_initialize_component()] - Performing Component Post Init: [mmu_server]
2025-02-22 22:58:18,480 [mmu_server.py:component_init()] - Spoolman not available. Happy Hare remote methods not available
...
```

The log output is also captured when a file is uploaded

```
2025-02-22 22:58:29,792 [shell_command.py:pipe_data_received()] - INFO:metadata:Running processor mmu_server 2.2 on file mcboatybenchyface.gcode...
2025-02-22 22:58:30,957 [shell_command.py:pipe_data_received()] - INFO:metadata:File processor mmu_server successfully complete
```

An alternative approach to fixing the issue is in this PR https://github.com/moggieuk/Happy-Hare/pull/646 which simply updates `mmu_server.py` to process the new argument used by moonraker's `metadata.py`. This works just as well, but neither approach supports backwards compatibility which may be desirable.

### DIFF
--- a/components/mmu_server.py
+++ b/components/mmu_server.py
@@ -28,6 +28,7 @@ from typing import (
 )
 
 if TYPE_CHECKING:
+    from .file_manager import FileManager
     from .spoolman import SpoolManager, DB_NAMESPACE, ACTIVE_SPOOL_KEY
     from ..common import WebRequest
     from ..common import RequestType
@@ -38,6 +39,7 @@ if TYPE_CHECKING:
     from .klippy_apis import KlippyAPI as APIComp
     from .history import History
     from tornado.websocket import WebSocketClientConnection
+
 
 MMU_NAME_FIELD   = 'printer_name'
 MMU_GATE_FIELD   = 'mmu_gate_map'
@@ -61,7 +63,6 @@ class MmuServer:
         # Full cache of spool_ids and location + key attributes (printer, gate, attr_dict))
         # Example: {2: ('BigRed', 0, {"material": "pla", "color": "ff56e0"}), 3: ('BigRed', 3, {"material": "abs"}), ...
         self.spool_location = {}
-
         self.nb_gates = None             # Set during initialization to the size of the MMU or 1 if standalone
         self.cache_lock = asyncio.Lock() # Lock to serialize a async calls for Happy Hare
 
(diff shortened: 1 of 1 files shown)


components/mmu_server.py
```diff
@@ -77,11 +78,31 @@ def __init__(self, config: ConfigHelper):
...
+            "command": " ".join([
+                sys.executable,
+                os.path.abspath(__file__),
+                "-f",
+                "{gcode_file_path}",
+                "-m" if config.getboolean("enable_file_preprocessor", True) else "",
+                "-n" if config.getboolean("enable_toolchange_next_pos", True) else "",
+            ]),
+            "timeout": 300,
+            "version": config.getfloat('happy_hare_version', 2.2),
```
skyguy94, line 94: I wasn't sure how to handle the version and this doesn't work.

components/mmu_server.py
```diff
@@ -77,11 +78,31 @@ def __init__(self, config: ConfigHelper):
...
-
         # Options
         self.update_location = self.config.getboolean("update_spoolman_location", True)
+        self.proc_config: Dict[str, Any] = {
+            "name": "mmu_server",
+            "command": " ".join([
+                sys.executable,
+                os.path.abspath(__file__),
+                "-f",
+                "{gcode_file_path}",
```
skyguy94, line 89: This template value is baked into the new metadata.py: https://github.com/Arksine/moonraker/blob/master/moonraker/components/file_manager/metadata.py#L1065

Its the only value as far as I can tell.

components/mmu_server.py
```diff
@@ -77,11 +78,31 @@ def __init__(self, config: ConfigHelper):
...
+                "-m" if config.getboolean("enable_file_preprocessor", True) else "",
+                "-n" if config.getboolean("enable_toolchange_next_pos", True) else "",
+            ]),
+            "timeout": 300,
+            "version": config.getfloat('happy_hare_version', 2.2),
+            "ident": {
```
skyguy94, line 95: The "ident" values are used by metadata.py as a way of managing state so the processor only runs once.

moggieuk:
Hi!  I received two PRs to address this and pulled the other one first.  Can you verify that the current latest update works and if so close out this PR?

I really appreciate the help!

skyguy94:
Sure. I'll close this out. But for future reference, the other one does work, but its approach is to add support for the `-c` argument for klipper estimator while keeping the existing "replace and wrap" logic around METADATA.py.

This one embraces the new functionality in moonraker that allows for pluggable metadata 'modules' and does away with the "replace and wrap" logic.

When you feel like you have some time, please revisit this as I think its worthwhile to migrate to this approach long term.

PR #649, opened 2025-02-23
Accommodate fans other than [fan] in Blobifier

ksummers92:
Added a variable (fan_name) to allow restoring fan speed after using Blobifier if one is using a fan definition other than [fan] for a part fan.

For example, I am using the [auxiliary fan functionality](https://github.com/SoftFever/OrcaSlicer/wiki/Auxiliary-fan) in OrcaSlicer which has a different configuration in order to use properly in which the part fan is controlled using a [fan_generic] definition and an override of the M106 command. Therefore, using printer.fan.speed to get the current fan speed and save it as `backup_fan_speed` will result in the "true" part fan being set to 0.

The implementation is compatible with either configuration of part fan with the variable providing a default example which is the suggested configuration provided by the OrcaSlicer wiki.


ksummers92:
@Dendrowen since this is your area of the code. Can you take a look? I've tested it both ways and it works as intended.

Dendrowen (approved):

JonasBouchard:
It seems like the problem has not been fixed in the software only version of blobifier. Can somebody confirm?

PR #651, opened 2025-02-25
Update pin_defs to include MMBv2

Miriax:
Here are updated pin definitions for MMBv2.


moggieuk:
Thanks for the pins.   It won't work as supplied because the comments in the lines but I'll copy into a new definition.  Thanks again.

moggieuk:
Going to close this PR out given that pins_defs were incorporated in another PR.  Thank you :-)

PR #660, opened 2025-03-05
Update mmu.py

Miriax:
Add indentation on line 4400 to bring the `self.log_debug` statement back inside the `for` loop's scope. 
(I think, I don't know much about Python.)


moggieuk:
This was fixed with recent update.  The variable "i" should have been declared outside of the for loop to allow for visibility afterwards.

PR #671, opened 2025-03-15
fix(ServoSelector): initialize selector_bypass_angle with 0

ammmze:
Otherwise during klipper start up the has_bypass method gets called before this get loaded from the saved variables


moggieuk:
Sorry - I pushed a fix before I saw this.  I initialize to '-1 which is the uncalibrated value.  It is necessary to "calibrate" before the bypass can be used with a command similar to:
```
MMU_CALIBRATE_SELECTOR BYPASS=1 ANGLE=0
```

PR #674, opened 2025-03-19
fix: When automatically heating extruder, use gate temperature on gate 0 as well

burkfers:
When changing tools from idle, loading or unloading gate 0 did not use the gate temperature, but the default  temperature.

I believe the `> 0` check was in error because there's two constants for unknown and bypass in the negative, but 0 should already represent a valid gate.

I only tested that is now correctly uses the gate temperature when automatically heating for gate 0, but no other scenario.


norris108 (approved):

moggieuk:
Good catch!  You are correct...  will merge now.

Issue #669, opened 2025-03-13
variable_toolchange_fan_speed Error with tip forming


chrisli12:
Related issue #642 
I have a similar issue where I am using aux fan config in klipper to control my part cooling fan with Orca Slicer. https://github.com/SoftFever/OrcaSlicer/wiki/air-filtration

moggieuk:
Fixed in v3.2 release

PR #677, opened 2025-03-25
Tip forming part cooling fan assist for generic fan

chrisli12:
Added variable_fan_name for using [auxiliary fan](https://github.com/SoftFever/OrcaSlicer/wiki/Auxiliary-fan) in Orcaslicer, which redefines part cooling fan as a generic fan, during tip forming workflow.

See #649 for a similar use case.

#669


moggieuk:
Sorry, I've been distracted, but this is a good idea.  I'll get to it very soon.

moggieuk:
@chrisli12 this has been merged but I changed the variable name to `variable_toolchange_fan_name` to namespace with other related variables.  Thanks!

PR #688, opened 2025-04-17
add a new feature automatically adjust rotation distance value

ntchris:
add a new feature automatically adjust rotation distance value

1 To work with printer has two tension sensors setup - one for compressed and one for expended. 
It completely eliminates the need for calibrating all the gear for each gate.
2 This feature can be disabled or enabled using the new config option **sync_auto_adjust_rotation_dist_enable**.
all related operation is active only when the option is enabled.
3 mmu_status show related message, when the option is enabled, it doesn't ask user to do gear calibration since it's not needed.

**Test:**
1  has been tested using the older Happy Hare 2.7 for about half year and using it to print a few KG filament during the period. Recently I upgraded to happy hare 3.1, so spent some time to rewrite the code to make it work with 3.1.
2 set gate0 has gear calibrated, it uses gate0's value as rotation dist.
3 set all gates has gear calibration -1, it uses default as starting rotation dist value.
4 set a far off value 30 as calibrated rotation dist for gate0 and print using gate5(-1), and started the print, the value was very quickly adjusted and finally stayed as 22.7 in a few minutes and print continued.
see attached test log
[test_log_start_rd_30.log](https://github.com/user-attachments/files/19787467/test_log_start_rd_30.log)

**More detail description:**
 filament sync tension handler, help to adjust rotation dist in real time automatically
 to search and use the perfect rotation_dist value.
 Theory of operation:
 use binary search to get the perfect rotation dist value.
 the upper limit: can slowly trigger expanded sensor (not enough filament)
 the lower limit: can slowly trigger compressed sensor ( too much filament)
 the perfect rotation dist is always within the upper and lower limit.
 when in neutral, get an avg value from upper and lower set it as new rotation_dist, and test it,
 when it triggers another upper or lower limit, and use it as a new upper or lower limit value.
 As it goes, slowly and slowly we have smaller and smaller range from upper to lower value,
 and can have a better and better, more and more precise rota dist value.

Thanks


moggieuk:
Hi!  I'm not going to be able to look at this for 1-2 weeks because of some travel but interesting idea to use the sync feedback to correct rotation distance.

Just curious if you have tried these options:
```yml
# These are auto calibration/tuning settings. Once the gear rotation_distance and encoder are calibrated, enabling these options
# will lessen the initial calibration and will automatically tune bowden length and individual gate rotation_distance differences.
# Note: What can be tuned is based on "variable_rotation_distance" and "variable_bowden_lengths" settings in mmu_hardware.cfg
#       E.g. with fixed bowden and multiple BMG gears and encoder like the ERCF, the bowden length is tuned on gate#0 and
#            rotation_distance (MMU_CALIBRATE_GATE) is tuned for other gates.
#
autotune_bowden_length: 0       # Automated bowden length calibration/tuning. 1=automatic, 0=manual/off
autotune_rotation_distance: 0   # Automated gate calibration/tuning (requires encoder). 1=automatic, 0=manual/off
```
They don't use the sync-feedback but rather then difference in actual gear movement over what is expected and adjust bowden length if the machine has variable bowdens (not many) or the rotation distance if the machine has different per-gate "gears"

ntchris:
Thanks, please take your time,
unfortunately I have not tried not options.

Another thing, If I may make a suggestion, the mmu.py file is getting quite large, currently about 8K lines. 
It makes the mod/edit/merge work quite difficult especially for yourself, all almost aways has merge conflict.  
Would it be possible to break this large file down to multiple smaller files so future work should be a lot easier hopefully.
Thanks again!

moggieuk:
On the `mmu_parameters.cfg` size... quite possibly.  I'm with the work of @wdcocq going to completely redo configuration.  The PR offers a more scabable approach using Kconfig but then, to support multiple dissimilar MMUs on the same printer I'm going to make it more modular.  It is a LOT of work but it is happening for the v3.3 release.

More on the options that are already there:
the idea is that if HH performs a successful load (or unload) it call _autotune().  This will use two bits of info (if available) to adjust bowden_length or rotation_distance.  Repeating comments from code:
```yml
    # Use data from load or unload operation to auto-calibrate / auto-tune
    #                   
    # Data we can use:
    #  - ratio of large bowden move to that measured by encoder (0 if it can't be relied on)
    #  - the amount of unexpected homing necessary to reach endstop. We want some homing
    #    movement but we can't use excessive numbers for tuning (None indicates not available)
    #  - the direction of filament movement
    #
    # Things we could/can tune:
    #  - If gate 0, use the bowden move ratio to update encoder calibration ("encoder calibration"). Dangerous so not done!
    #  - If gate 0, use excess homing move to tune the calibrated bowden length ("bowden calibration")
    #    but only do this if bowden move ratio is reasonable. Can be done in both directions
    #  - If gate >0, use the bowden move ratio to set/tune the gear rotation_distance ("gate calibration")
    #    but only do this if homing movement data tells us we haven't overshot. Can be done in both directions
    #
    def _autotune(self, direction, bowden_move_ratio, homing_movement):
    ```
One of the challenges I have is that many MMU designs have same bowden length per-gate much some don't (where the path from gate homing point to combination point is different).  In these designs, updating the rotation_distance wouldn't be correct, instead it would mean updating the bowden length for that gate. Bottom line is that this isn't easy to so in a generic way and it would be really nice to aggregate all the telemetry and update in one place.

That said, your concept is more of fine tuning the rotation_distance during a print using sync-feedback.  That's actually clever so long as it (IMHO) is limited in adjustment range.  I.e. what if the MMU gear stepper is slipping and thus only appears to be moving too slowly...

**Comment by bf6884** [comment]
Hi guys,

first of all, thank you for the great Happy Hare project and what I think is a great idea in this topic here. 

In my opinion, this topic is a great improvement approach. I'm really looking forward to implementing it and thank you already.
I have tested the code from ntchris. You can see very clearly in TPU 95A that the theme brings significant advantages there. Without adjusting the rotation distance, I keep getting CLOGs because too much filament is being fed by the ERCF. I have been using the adjustment for a few days now and am very pleased that I can now print TPU much better.

I hope that the attachments work, the difference can be clearly seen there

https://github.com/user-attachments/assets/a67c453c-ae1a-4bd2-98d9-cb068fa2dd64

TPU without change of rotation distanz

**Comment by moggieuk** [comment]
Quick update .. I haven't forgotten about this but just been swamped with other things (mainly HH v4!!)

**Comment by igiannakas** [comment]
Just one thing to consider:
**Please make sure this calibration is disabled when loading and when purging filament as well as ignore the sensor state after a purge until it is set to neutral through the sync adjustment**. 

Why? because the sync sensor will almost certainly be in an expanded (compression) state after filament loads and the wipe tower is printed and/or the blobifier executes. That is due to the collision of the filament with the extruder pushing the sensor back to a compressed state. 

So ignoring the sensor value until the sensor has been re-set to neutral (via the sync feedback mechanism) **during a print** - and after the "purge" routines have completed, and then start the search process is essential.

Similarly this feature has to be disabled when unloading filament, as the sensor will most likely end up in a tension/compression state due to the movement of the filament during the cut process.

**Comment by moggieuk** [comment]
@ntchris @igiannakas 

I'm just finishing up "merging this PR".  I think the idea is excellent but I did find some oversights and general cleanup (like supporting python2) was necessary. I hope I don't offend because I always appreciate the help.  Anyway, I'm in final testing of this idea (it's in the `sync_fix_dev` branch -- feel free to look at the code, but don't try to use just yet).

Bullet list changes:
* Complete rewrite of sync-feedback handling.  This is now in a separate Class called `MmuSyncFeedbackManager`
* Integrated this idea with the existing `bang-bang` logic rather than trying to repeat logic.
* Made the logging more consistent with the rest of HH
* Coded so compatible with Python2 (unfortunate but klipper still supports this and its on about 10% of systems)
* Fixed (what I think) was broken logic on the determination of "perfect tuning" - previously the logic to "get of the sensor" occurred before the "is perfect" check.
* Don't force exactly equality of upper and lower bounds but instead set tolerance level.
* Changed the way I monitor extruder movement so I can now give callbacks after a specific movement of filament whilst in a compression/tension state - previously the logic relied on direction changes as a count for when to check.
* Added new calibration/autotune options that is part of another update but applies to this as well.
* Made the logic handle direction changes of the extruder - maybe not a real world case but I wanted the logic to be sound.
* Remove the need to reset on print start or operation only when in print - it is really more of a per-gate operation (namely because rotation_distance of each gate can be different because they use different BMG gears).

If you are reviewing code, then the key is the `rd_clamps` dict. This has an array entry for each gate. The array is initialized on first time gate selection and contains: rotation_distance settings for: slow speed, current, fast speed, tuned (if available), initial value.  The logic works on either sync state change events, extruder direction change events or the new "moved too far in same state" event.

Bottom line is that it is quite different from the PR but I think I have retained the intent. WDTY?

@igiannakas I need to think about the cases you raised, but I want to note that this can only operate when the "gear is synced to extruder".  This is not the case during load and unload.  It's subtle but in those cases "the extruder is synced to gear" -- the direction is important because it determines which stepper is driving.  When in the "extruder is synced to gear" or when the extruder is released completely (after filament is out of the extruder) then rotation_distance of the gear will be the calibrated or default value.

**Comment by igiannakas** [comment]
No offence taken in the slightest - I’m glad it triggered a code review and further improvements :)

the sync feedback I mention is immediately post load and basically during a blobifier purge - I’m unsure how this is handled but in my mind it’s a “special” case as it extrudes significant filament volume with no “break” to allow for rotation distance adjustment. If you’re slightly off with the extruder and Mmu rotation distances it can trigger this.

For example, I’ve noticed on a few occasions issues where the belay was maxing out travel distance and the filament grinding due to the feedback mechanism not acting in time / correctly.

I think even without a blobifier immediately post load the sensor will most likely be in a compression state - as there is slight collision with the extruder till the extruder picks the filament up. Basically the Mmu keeps driving the filament while the extruder hasn’t had enough force exerted to it by the filament to pick it up. At some point it does and then feeds. 

Basically please don’t assume the extruder instantly pulls the filament when loaded and that the sensor will be in a neutral state post load :)

**Comment by moggieuk** [comment]
Understand the immediate post load condition (IF you are using the sync feedback compression to detect the extruder.  In the above branch I have a new parameter `sync_feedback_buffer` with is the distance of travel of the sensor.  I use half of this distance to reduce the calibrated bowden load length.  So in theory if you have toolhead sensor and don't FORCE a home (the default with toolhead sensor would be not to home), then the extruder loading should occur with the sync_feedback in a neutral position...  so even though it doesn't come into play until the actual purge (when the extruder takes over movement from the MMU) it should still be neutral.

The purge is a challenging problem to solve because of the long move and the fact klipper has a "look ahead" that will calculate moves back on old rotation_distance....

I do have a `toolhead_post_load_tighten` move that is designed to tighten the filament before printing to prevent falso clog detection IF NOT syncing -  I move a % of the clog detection length.  I could see using this hook to neutralize the sync_feedback sensor after loading IF SYNCED -- easy move of just the MMU stepper until NEUTRAL.  That and the fact the initial rotation_distance will be good after calibration means that purge shouldn't be a problem. Hmmm.  Let me think about this.  Of course I will need help testing...

I looked at code and I like this idea.  I could rename:
`toolhead_post_load_tighten` to `toolhead_post_load_tension_adjust` and get it to do the appropriate adjustment based on MMU...

**Comment by igiannakas** [comment]
I’ve noticed the sync feedback sensor going to the Extended position even when using an extruder endstop.  

This is because of friction - you need compression to force the filament trough the extruder sensor and onto the gears. 

Loading the Bowden with a reduced length doesn’t change that - you will still encounter compression as you load the extruder :) I can try and film it for you so you see what I mean. 

The post load tighten move is not enabled / available today when using synched steppers (eg type b Mmu). It could solve this by tightening after load till sensor is neutral though (instead of a % of clog length), that could do it!!

**Comment by moggieuk** [comment]
@ntchris  Just letting you know that I have a `v330_beta` branch which includes all the new auto-calibration ideas (both with sync-feeback but also another approach based on delta in homing distance on bowden load/unload.   If you wanted to try.... 

If you do try `v330_beta` then make a copy of `printer_data/config/mmu` to go back to if you have to revert to `main` .. it will not automatically revert back.

Usual to upgrade:
```yml
./install.sh -b v330_beta
```
Here is a quick "what's changed":
1) Support for new KMS MMU!  This has just been launched in China and will be available worldwide soon. It's a really nice MMU with integrated espooler and heated filament chamber.
2) Rework of MMU/extruder synchronization and grip/release of filament on type-A MMU's.  Whilst most should not see any change, this fixes some corner cases that have been reported and enhances similar operation of MMUs. In a nutshell, gear/extruder synchronization is now always dependent on the location of the filament - if the filament is determined not to be in the extruder, the steppers will not be synchronized. Users of 3DChameleon: I am unable to test since I don't have this MMU so please provide feedback.
3) Auto-calibration / Auto-tuning - New options have been added to `mmu_parameters` (see below). These are designed to be set on initial install and ease first-time use on many of the newer MMU's.  For example, on QuattroBox or BoxTurtle or other type-B design fitted with sync-feedback ("buffer") it is now possible to use without initial calibration and have important values tuned over time. Note that I still recommend always accurately calibrating the rotation_distance of gate 0 (with `MMU_CALIBRATE_GEAR` command) because it will greatly reduce the time for other parameters like bowden length and other gate rotation_distances to tune.
```yml
autocal_bowden_length: 1        # Automated bowden length calibration. 1=automatic, 0=manual/off
autotune_bowden_length: 1       # Automated bowden length tuning. 1=on, 0=off
skip_cal_rotation_distance: 0   # Skip rotation distance calibration (MMU_CALIBRATE_GEAR), 1=skip, 0=require
autotune_rotation_distance: 1   # Automated gate calibration/tuning. 1=automatic, 0=manual/off
skip_cal_encoder: 0             # Skip encoder calibration (MMU_CALIBRATE_ENCODER), 1=skip, 0=require
```
4) Thanks to a great idea and PR from @ntchris the `autotune_rotation_distance` functionality is also operational to fine tune synchronization of MMU and extruder

PS. Turing on TRACE log_file_level (or even the STEPPER level) and watching the log with:
```yml
tail -f mmu.log | egrep "MmuSyncFeedbackManager"
```
is a good way to see the actions of the auto-tuning..

All logic is in a new separate class now (called `mmu_sync_feedback_manager.py`).  These are the hardcoded settings:
```yml
    FEEDBACK_INTERVAL     = 0.5     # How often to check extruder movement
    SIGNIFICANT_MOVEMENT  = 5.      # Min extruder movement to trigger direction change (don't want small retracts to trigger)
    MOVEMENT_THRESHOLD    = 50      # Default extruder movement threshold trigger when stuck in one state
    MULTIPLIER_RUNAWAY    = 0.25    # Used to limit range in runaway conditions (25%)
    MULTIPLIER_WHEN_STUCK = 0.01    # Used to "widen" clamp if we are not getting to neutral soon enough (1%)
    MULTIPLIER_WHEN_GOOD  = 0.005   # Used to move off trigger when tuned rotation distance has been found (0.5%)
    AUTOTUNE_TOLERANCE    = 0.0025  # The desired accuracy of autotuned rotation distance (0.25% or 2.5mm per m)
```
There is also a hidden `sync_movement_threshold` parameter which defaults to MOVEMENT_THRESHOLD (i.e. 50mm).  This is how much extrusion to push before the sync manager determines we need to take additional action to change rotation_distance.

Happy to chat with you about this... I think it all works, I just need your help in tuning the default MULTIPLIER values and other tuneables.

PR #703, opened 2025-05-07
Changed PrinterRail to GenericPrinterRail as per the klipper class

trajesus:
Was getting an error after upgrading klipper and Happy-Hare:
 - klipper module 'stepper' has no attribute 'PrinterRail'

After investigating the code, it seems that the class PrinterRail in the klipper stepper.py file is now GenericPrinterRail


trajesus:
will close this as #704 takes into acount older versions of klipper

PR #704, opened 2025-05-07
fix: support klipper v0.13.0-79

ammmze:
In klipper v0.13.0-79 (from Klipper3d/klipper#6815) the printer rail code was refactored and requires some refactoring on our side

I did very basic testing on both klipper v0.13.0-79 and v0.13.0-78 (`git reset --hard 1cc63980747b80516f8fc4f022eedf18ae739086`) and verified klipper could start and then used `MMU_CHECK_GATES` to ensure it could move each of the mmu gear steppers in my Box Turtle.

### DIFF
--- a/extras/mmu_machine.py
+++ b/extras/mmu_machine.py
@@ -820,9 +820,12 @@ class MmuHoming(Homing, object):
             self.toolhead.set_position(homepos)
 
 
+_StepperPrinterRail = stepper.PrinterRail if hasattr(stepper, 'PrinterRail') else stepper.GenericPrinterRail
+
+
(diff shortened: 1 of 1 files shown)


moggieuk:
Thank you very much. Klipper hasn't changed this part of the code for years - I guess they must be cleaning things up!

The PR looks good as far as I can tell without testing.  Given that it will be a few more days before I can deep dive it is probably best to pull the PR so no one is held up.  I can review a second time in a few days.

My only question was around the "Mock" endstop but it might just I need to review the klipper changes in more detail. Pulling now.

PR #705, opened 2025-05-07
Sync feedback corner cases 

CooperGerman:
Hi, here comes the PR that should cover missing cases for dual, single and non sensor configurations.

# theoretical values the tests were build upon
## Compression sensor only scenario

| compression state | tension state | triggered sensor | feedback value |
| ------------------|---------------|------------------|----------------|
| 0                 | 0             | compression      | -1             |
| 0                 | 1             | compression      | -1             |
| 1                 | 0             | compression      | 1              |
| 1                 | 1             | compression      | 1              |
| **0**             | **0**         | **tension**      | **-1**         |
| **0**             | **1**         | **tension**      | **-1**         |
| **1**             | **0**         | **tension**      | **1**          |
| **1**             | **1**         | **tension**      | **1**          |

## Tension sensor only scenario

| compression state | tension state | triggered sensor | feedback value |
| ------------------|---------------|------------------|----------------|
| **0**             | **0**         | **compression**  | **1**          |
| **0**             | **1**         | **compression**  | **-1**         |
| **1**             | **0**         | **compression**  | **1**          |
| **1**             | **1**         | **compression**  | **-1**         |
| 0                 | 0             | tension          | 1              |
| 0                 | 1             | tension          | -1             |
| 1                 | 0             | tension          | 1              |
| 1                 | 1             | tension          | -1             |

## Both sensors scenario

| compression state | tension state | triggered sensor | feedback value |
| ------------------|---------------|------------------|----------------|
| 0                 | 0             | compression      | 0              |
| 0                 | 1             | compression      | -1             |
| 1                 | 0             | compression      | 1              |
| 1                 | 1             | compression      | 0              |
| 0                 | 0             | tension          | 0              |
| 0                 | 1             | tension          | -1             |
| 1                 | 0             | tension          | 1              |
| 1                 | 1             | tension          | 0              |

## No sensors scenario

| compression state | tension state | triggered sensor | feedback value |
| ------------------|---------------|------------------|----------------|
| 0                 | 0             | compression      | 0              |
| 0                 | 1             | compression      | 0              |
| 1                 | 0             | compression      | 0              |
| 1                 | 1             | compression      | 0              |
| 0                 | 0             | tension          | 0              |
| 0                 | 1             | tension          | 0              |
| 1                 | 0             | tension          | 0              |
| 1                 | 1             | tension          | 0              |

# What changes ?
- Single sensor setups will now oscillate between `compressed` and `tensioned` and no longer fallback to neutral when the sensor is not triggerd (which sparked this PR initially).
- Sensors can trigger feedback events with corrected values even though disabled. This adds a little bit of safety in order to ensure that the system always behaves in the way the sensors are configured.
- The stress test suite has been updated to be able to iterate over all "sensor configurations" and randomly generate _realistic_ triggers. (This excludes switching sensors at the same time for instance)
- The `_MMU_TEST SYNC_STATE=loop` command will create missing sensors temporarily so that the tests do not rely on configured sensors.
- The test analysis in the mmu log file has been enhanced in order to facilitate debugging.

# :warning:  Important notes 
The tests and updated feedback logic has been tested by myself (dual sensors setup), and @burkfers with a single "compression" sensor to my knowledge.

I would be happy that somebody could confirm that the temporary sensor creation and removal in the test is done cleanly enough not to let some unwanted python objects in klipper after testing.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -2916,7 +2916,7 @@ class Mmu:
         if not self.is_enabled: return
         if abs(state) <= 1:
             self.sync_feedback_last_state = float(state)
-            self.log_trace("Got sync force feedback update. State: %s" % self._get_sync_feedback_string(detail=True))
+            self.log_trace("Got sync force feedback update. State: %s (%s)" % (self._get_sync_feedback_string(detail=True), float(state)))
             if self.sync_feedback_enable and self.sync_feedback_operational:
                 self._update_sync_multiplier()
         else:
@@ -2978,9 +2978,11 @@ class Mmu:
             if go_slower(self.sync_feedback_last_state, self.sync_feedback_last_direction):
                 # Expanded when extruding or compressed when retracting, so decrease the rotation distance of gear stepper to speed it up
                 multiplier = 1. - (abs(1. - self.sync_multiplier_low) * abs(self.sync_feedback_last_state))
+                self.log_trace("Slowing gear motor down")
             else:
                 # Compressed when extruding or expanded when retracting, so increase the rotation distance of gear stepper to slow it down
                 multiplier = 1. + (abs(1. - self.sync_multiplier_high) * abs(self.sync_feedback_last_state))
+                self.log_trace("Speeding gear motor up")
         self.log_trace("Updated sync multiplier: %.4f" % multiplier)
         self._set_rotation_distance(self._get_rotation_distance(self.gate_selected) / multiplier)
 
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -2914,9 +2914,9 @@ class Mmu:
     # or can be a proportional float value between -1.0 and 1.0
     def _handle_sync_feedback(self, eventtime, state):
         if not self.is_enabled: return
-        self.log_trace("Got sync force feedback update. State: %s" % state)
         if abs(state) <= 1:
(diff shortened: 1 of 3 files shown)


CooperGerman:
Don't mind the commits : https://github.com/moggieuk/Happy-Hare/pull/705/commits/8790c2cc04f4b41cc98ac6281b5f8abe45b71475 https://github.com/moggieuk/Happy-Hare/pull/705/commits/5249fe7a447e2de7ba9ea512f78f4da11b382c79 and https://github.com/moggieuk/Happy-Hare/pull/705/commits/b42276c2b44b3cf99ac1144c2c40887cf37f3f4a. My brain died for 10 minutes ... I had issues with some 1.5 mm diameter filament that cause lots of "clogging issues" as i tried to final test the feedback logic. :relieved: 

The branch is merge ready as far as I'm concerned but i very much would be inclined to see it be a little more tested and a quick review could not hurt :pray: 

I was thinking that the stress test might be run as CI-CD non regression maybe. I'm not really sure how this could be achieved without some virtual printer setup.

moggieuk:
Thanks for the hard work is testing this.  I'd really like to be able to run myself -- I'll be able to do this in 2 days.

BTW You were asking me about the "two triggers at once" case which will currently flag an error in the UI... I was thinking about it and it is actually a really good solution for a minimal movement sync-feedback sensor.  It didn't make sense at first but it does now!

PR #710, opened 2025-05-16
Add an option to suppress the Kalico warning

burkfers:
The warning that Kalico is supported on a best-effort basis only is very useful, but seeing an error every time you boot Klipper gets very old quickly, and it may even mask real errors, which you've now become desensitized to.

I am proposing a new, hidden config option that suppresses the warning. It still emits a message at trace level so if digging into a nitty gritty it's still obvious Kalico is in use.
If the new option is to be documented on the wiki, it should probably come with a big warning to please never forget to mention running Kalico when seeking support.

I've remembered to add the option to `hidden_params` in the installer so an upgrade does not eat the setting.

To suppress the warning, simply set `suppress_kalico_warning: 1` in `mmu_parameters.cfg`.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -484,6 +484,8 @@ class Mmu:
 
         # Turn off splash bling for boring people
         self.serious = config.getint('serious', 0, minval=0, maxval=1)
+        # Suppress the Kalico warning for dangerous people
+        self.suppress_kalico_warning = config.getint('suppress_kalico_warning', 0, minval=0, maxval=1)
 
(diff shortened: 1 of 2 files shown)


ammmze (approved):
This looks reasonable to me 👍🏻

burkfers:
As discussed on Discord:
The Kalico warning might become a regular message instead of an error, solving the annoyance of popups in UIs and the worry of missing real errors.
An option to change it from a red normal message to an uncolored normal message may or may not be added. Fully suppressing it is not viable as that would increase the support burden.

PR #711, opened 2025-05-16
Fix extruder_name for multiextruders printers.

Zeta3D:
Fix extruder_name for multiextruders printers.
This line was not using self.extruder_name and it was hardcoded to "extruder"


moggieuk:
Thanks!

PR #713, opened 2025-05-19
fix: Prevent BLOBIFIER from continuing purge after MMU_PAUSE

bluekiller:
The BLOBIFIER macro continues to execute commands after calling MMU_PAUSE, which can lead to unintended purging at the park position. This fix ensures that the macro properly exits after pausing, preventing any remaining purge operations from being executed.

- Add early return after MMU_PAUSE to prevent continued execution
- Preserve remaining purge length for resumption after pause

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -322,6 +322,7 @@ gcode:
     {% set from_tool = printer.mmu.last_tool %}
     {% set to_tool = printer.mmu.tool %}
     {% set bl_count = printer['gcode_macro _BLOBIFIER_COUNT'] %}
+    {% set bl_pause = printer['gcode_macro _BLOBIFIER_PAUSE_STATE'] %}
     {% set pos = printer.gcode_move.gcode_position %}
     {% set safe = printer['gcode_macro _BLOBIFIER_SAFE_DESCEND'] %}
     {% set ignore_safe = safe.print_height < force_safe_descend_height_until %}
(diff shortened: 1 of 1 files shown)


config/addons/blobifier.cfg
```diff
@@ -445,55 +446,59 @@ gcode:
 
       # Repeat the process until purge_len is reached
       {% for blob in range(blobs) %}
-        RESPOND MSG={"'BLOBIFIER: Blob %d of %d (%.1fmm)'" % (blob + 1, blobs, purge_per_blob)}
+        {% if bl_pause.in_pause %}
```
ammmze, line 449: I don't think this will work as you expect. These template expressions are not evaluated in real-time. Let try to explain with a smaller example example...

Say I have the following:

```
[gcode_macro BLOBIFIER]
variable_prefix: 'Foo'
gcode:
  {% set prefix = printer['gcode_macro BLOBIFIER'].prefix %}
  {% for i in range(5) %}
    {% if i > 3 %}
        SET_GCODE_VARIABLE MACRO=BLOBIFIER VARIABLE=prefix VALUE='Bar'
    {% endif %}
    RESPOND MSG="{prefix} {i}"
  {% endfor %}
```

When I invoke `BLOBIFIER`, what klipper will do BEFORE executing any of the gcode in `BLOBIFIER`, it will first render the template and produce something like this:

```
RESPOND MSG="Test 0"
RESPOND MSG="Test 1"
RESPOND MSG="Test 2"
RESPOND MSG="Test 3"
SET_GCODE_VARIABLE MACRO=BLOBIFIER VARIABLE=prefix VALUE='Bar'
RESPOND MSG="Test 4"
```

Now, klipper starts executing the content, calling that `RESPOND` macro five times and the `SET_GCODE_VARIABLE` once...but there is no real-time expression evaluation while the content is executing.

To work around that limitation, you need to extract the code that is dependent on the gcode variable (and the reading of that gcode variable to another macro). Something like this:

```
[gcode_macro BLOBIFIER]
variable_prefix: 'Foo'
gcode:
  {% for i in range(5) %}
    {% if i > 3 %}
        SET_GCODE_VARIABLE MACRO=BLOBIFIER VARIABLE=prefix VALUE='Bar'
    {% endif %}
    _BLOBIFIER_LOOP ITEM={i}
  {% endfor %}

[gcode_macro _BLOBIFIER_LOOP]
gcode:
  {% set i = params.ITEM|int %}
  {% set prefix = printer['gcode_macro BLOBIFIER'].prefix %}
  RESPOND MSG="{prefix} {i}"
```

What that will do when you call `BLOBIFIER`, it will first render the `BLOBIFIER` template to the following:

```
_BLOBIFIER_LOOP ITEM=0
_BLOBIFIER_LOOP ITEM=1
_BLOBIFIER_LOOP ITEM=2
_BLOBIFIER_LOOP ITEM=3
SET_GCODE_VARIABLE MACRO=BLOBIFIER VARIABLE=prefix VALUE='Bar'
_BLOBIFIER_LOOP ITEM=4
```

Then klipper will execute those commands. When it executes `_BLOBIFIER_LOOP ITEM=0`, it will then render the `_BLOBIFIER_LOOP` template with the given parameters, producing the following and then executing it.

```
RESPOND MSG="Test 0"
```

After it has completed that one, it will then render and execute `_BLOBIFIER_LOOP ITEM=1`, and so on until it has executed all of the commands. Now, the other issue is we can't really break out of the `for` loop early based on things changing within the loop, but that can be worked around by putting conditional stuff in that `_BLOBIFIER_LOOP` macro such that when the conditions are right and we _want_ to break out of the loop, we would want it to effectively render to nothing for that and future loop items.

bluekiller:
Thanks for your insights on the template rendering. I will rework for this. After test done I will raise a new PR.

PR #718, opened 2025-05-22
 Added gate_status_exit Support and New Effects

Batalhoti:
### Description
This PR enhances the Happy Hare MMU LED control system with new functionality and improvements:

### New Features
**`gate_status_exit` mode**  
- New white LED effect variant for gate status indication  
- Provides visual distinction from standard `gate_status` (green/blue/orange)  
- Ideal for exit operation feedback  

**Static `mmu_white` effect**  
- Added to static LED conversion dictionary  
- Pure white RGB (1,1,1) for systems without `led_effects`  

### Improvements
Added `gate_status_exit` support in `_MMU_LED_GATE_MAP_CHANGED`  
Enhanced static LED fallback logic  

### Compatibility
Fully backward compatible  
Requires LED configuration in `mmu_hardware.cfg`  

### Modified Files:
```
mmu_leds.cfg
mmu_macro_vars.cfg
```

### DIFF
--- a/config/base/mmu_leds.cfg
+++ b/config/base/mmu_leds.cfg
@@ -78,9 +78,9 @@ gcode:
     {% set set_led_vars = printer['gcode_macro _MMU_SET_LED'] %}
 
     {% set current = set_led_vars['current_exit_effect'] %}
-    {% set exit_effect=current if current in ["gate_status", "filament_color", "slicer_color"] else "" %}
+    {% set exit_effect=current if current in ["gate_status", "gate_status_exit", "filament_color", "slicer_color"] else "" %}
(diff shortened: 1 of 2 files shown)


moggieuk:
I've completed the much more flexible LED support but not sure which release to include it in yet.  It is really nice because HH doesn't care one bit about how many LEDs, how many pins are used to drive them, the order, etc.  The reason is that will now create "virtual strips" from whatever you configure..  looking forward to what MMU designers do with it ;-)

moggieuk:
I'm going to close this PR as "Wont't do". The reason as explained above is that a new all-python implementation of LED control is about to be released which has much greater flexibility including animation per-gate, etc.

PR #719, opened 2025-05-24
Cast UNLOAD parameter as int, so it can be set as false in MMU_END macro

Luoti:
I noticed that I cannot use `MMU_END` macro's `UNLOAD` parameter as false, because it is not casted as int, and string "0" is considered true in the if statement. So the unload happens even if you call this macro like

    MMU_END UNLOAD=0

Tested the jinja [here](http://jinja.quantprogramming.com/).
Screenshot_20250524_111403


moggieuk:
Thanks. pulling now.

PR #725, opened 2025-05-30
Update values for PicoMMU

AlejandroRivera:
Updated some values to match LH's PicoMMU current standards for Servo and Stepper motor for ease of adoption for people using HH directly or moving from PicoMMU's own gcode:

* [BOM](https://github.com/lhndo/LH-Stinger/tree/main/User_Mods/MMU/Stinger%20Pico%20MMU%20-%20%40LH#bom) uses an EMAX ES3004 Servo, which has 140º of motion and uses a 1.2 seconds wait for servo to reach target values.  See https://github.com/lhndo/LH-Stinger/blob/main/User_Mods/MMU/Stinger%20Pico%20MMU%20-%20%40LH/Klipper/sp_mmu.cfg#L39-L44

* Stepper Motor gear ratio should be 1.25:1 since the input gear has 20 teeth and the output gear has 25 teeth. See image (screenshot taken from [3MF file](https://github.com/lhndo/LH-Stinger/blob/main/User_Mods/MMU/Stinger%20Pico%20MMU%20-%20%40LH/Stinger%20Pico%20MMU.3mf))

* Stepper Motor hold current value in LH's code is 0.4. See [source code](https://github.com/lhndo/LH-Stinger/blob/main/User_Mods/MMU/Stinger%20Pico%20MMU%20-%20%40LH/Klipper/sp_mmu.cfg#L60)

* Pico MMU's default behavior is to only have sync'ing between extruder and MMU gears during extruder loading phase. At all other times, during normal printing, the servo is disengaged (at 0º) and stepper motor is disabled.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1747,10 +1747,10 @@ questionaire() {
             _hw_variable_bowden_lengths=0
             _hw_variable_rotation_distances=0
             _hw_require_bowden_move=1
-            _hw_filament_always_gripped=1
-            _hw_gear_gear_ratio="1.28:1"
+            _hw_filament_always_gripped=0
+            _hw_gear_gear_ratio="1.25:1"
             _hw_gear_run_current=0.7
-            _hw_gear_hold_current=0.1
+            _hw_gear_hold_current=0.4
             _hw_chain_count=4
             _hw_exit_leds="neopixel:mmu_leds (1-4)"
             _hw_entry_leds=""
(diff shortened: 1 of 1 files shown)


moggieuk:
Nice, thank you. Will get to this soon.
The syncing behavior may default one way but it is really a user choice.  Happy to make defaults match what the respective community expects..

AlejandroRivera:
Reverted hold current back to 0.1 as PicoMMU author didn't have a strong opinion on 0.4 being the correct value, as it's expected to be different depending on the stepper motor -- besides, holding current in PicoMMU's default operation isn't important as the stepper is not engaged all the time.

PR #726, opened 2025-06-01
Wrong pinout on ESPOOLER_FWD_2 mmu.cfg

theminjus:
I have corrected pintout on espooler 2 to avoid getting errors.


moggieuk:
Thank you!

PR #730, opened 2025-06-09
Enable purging (blobifier) when run out occurs

igiannakas:
Restoring previous behaviour where after a run out and endless spool triggering, the purge macros (blobifier) where invoked. In previous releases that was done automatically as the blobifier was tied to the post load macro user hook. 

This is particularly helpful to avoid blobs on the print if the toolhead dimensions and ooze reduction are not spot on (which is not as critical when a purge macro is used) and to ensure consistent nozzle pressure when the new filament is loaded and any air pockets in the nozzle caused by filament retraction, loading and inevitable ooze during the process, are purged out.

IMG_6320

Tested with an endless spool operation in latest HH with success.

(also adding .DS_Store files automatically generated by MacOS to the git ignore list to prevent them polluting the repo).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 my_*.cfg
+.DS_Store
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -5009,7 +5009,7 @@ class Mmu:
             # Deal with purging
             if purge == self.PURGE_SLICER and not skip_extruder:
                 self.log_debug("Purging expected to be performed by slicer")
-            elif purge == self.PURGE_STANDALONE and not skip_extruder and not self.is_handling_runout:
+            elif purge == self.PURGE_STANDALONE and not skip_extruder:
                 with self._wrap_track_time('purge'):
                     sync = (self.is_printing() and self.sync_purge) or self._standalone_sync
                     self.sync_gear_to_extruder(sync, current=True)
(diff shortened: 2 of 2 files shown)


igiannakas:
Bumping - @moggieuk any chance for a review please?

Swamp-Ig:
This has caused damage to my build plate - I had a runout in the first layer, and the toolhead didn't move up before re-homing to where it left off making a scratch.

igiannakas:
Probably something in the blobifier did not restore Z correctly?

igiannakas:
Probably something in the blobifier did not restore Z correctly? I have also tested with a runout on first layer and it was Ok. Is your blobifier z below the bed?

Swamp-Ig:
Yeh it is. I think it parked on the tray but then didn't blob or move up to safe height.  I don't really want to try testing it though.

igiannakas:
I’d advise setting the tray above the bed by printing a slightly taller spacer for the blobifier.

this PR doesn’t affect how the blobifier restores z height (which by the sounds of it may have a bug) and you’ll encounter the same issue if you are printing a multi color print that swaps colors on first layer if your z hop is set to 0 in the slicer and no prime tower is used.

also this could be mitigated by selecting the option to resume in the next position and by selecting a z hop value during tool change and run out in the happy hare macro config.

PR #760, opened 2025-08-03
BLOBIFIER: Turn off the part cooling fan before nozzle cleaning

Marek-Dvorny:
Changed blobifier.cfg GCODE, so the part cooling fan turns off before nozzle cleaning with brush for better cleaning.

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -479,6 +479,11 @@ gcode:
             G1 E2 F800
           {% endif %}
         {% endfor %}
+		
+		{% if part_cooling_fan >= 0 %}
+		# Reset part cooling fan if it was changed
+		M106 S{(backup_fan_speed * 255)|int}
+		{% endif %}
 
         # ==================================================================================
         # ==================== DEPOSIT BLOB ================================================
@@ -520,10 +525,6 @@ gcode:
     G90 # absolute positioning
     G1 Z{restore_z} F{travel_spd_z}
     
-    {% if part_cooling_fan >= 0 %}
-      # Reset part cooling fan if it was changed
-      M106 S{(backup_fan_speed * 255)|int}
-    {% endif %}
     
     M220 S{(backup_feedrate * 100)|int}
   {% endif %}


igiannakas:
I have observed something similar but my personal fan speed approach is a bit more elaborate: 
1. Set full fan when the pellet purge is complete but BEFORE retracting the tray. This helps unstick the blob from the nozzle and shrink it a little allowing for a cleaner blob depositing. Like the below:

```
M106 S255
{% if safe.tray or ignore_safe %}
```
2. Reset the fan to purge fan speed immediately after the blobifier servo out in move to prepare for next purge
          
```
          BLOBIFIER_SERVO POS=out
          BLOBIFIER_SERVO POS=in
          BLOBIFIER_SERVO POS=out
          
          # Keep track of the # of blobs
          _BLOBIFIER_COUNT
          M106 S{part_cooling_fan * 255}
```
3. Reset to print fan speed for brushing

```
    {% if part_cooling_fan >= 0 %}
      # Reset part cooling fan if it was changed
      M106 S{(backup_fan_speed * 255)|int}
    {% endif %}
    {% if safe.brush or ignore_safe %}
      BLOBIFIER_CLEAN
    {% else %}
      G1 X{brush_start} F{travel_spd_xy}
    {% endif %}
```

These changes combined help the **pellet unstick better** as with high fan speed the filament shrinks and drops nicely off the nozzle and gets pushed out cleanly from the servo as the filament is then solid (hence snapped off). 

Then the nozzle brush cleans with the correct fan speed for the material, preventing the material from over solidifying if for example its ABS or over melting if its PLA.

**Results:**
IMG_6330

https://youtube.com/shorts/P2c9okgFZ-0

I would recommend that you attempt the above changes and give it a go, then we can look at the PR?

@moggieuk FYI - reviewing this PR.

Marek-Dvorny:
Thank you! I have tested your version and I think it's really good! It's better than mine. I had little problems with pieces of "strings" on sthe blobs and this fixed it. So, I will close the PR and it would be nice if your version would be merged into main. 👍 Once again: Thank you!

igiannakas:
I'll raise a PR hopefully today ;)

igiannakas:
@Marek-Dvorny if you could test this code I would be grateful. There's a few more changes vs. the one just above. Seeking feedback for all :)

PR #763, opened 2025-08-06
Servo move as default position

Marek-Dvorny:
Selector servo on ERCF V2 should be now in move position when not printing, so the servo arm is not stressed.

### DIFF
--- a/extras/mmu/mmu_selector.py
+++ b/extras/mmu/mmu_selector.py
@@ -972,8 +972,10 @@ class LinearSelectorServo:
             self.servo_down()
         elif not self.mmu.selector.is_homed or self.mmu.tool_selected < 0 or self.mmu.gate_selected < 0:
             self.servo_move()
-        else:
+        elif self.mmu.is_printing() and not self.mmu_toolhead.is_gear_synced_to_extruder():
             self.servo_up()
+        else:
+            self.servo_move()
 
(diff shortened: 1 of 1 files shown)


moggieuk:
Thanks for the PR -- just getting back to them.

Hmmm, The servo shouldn't be stressed in 'up' position unless it is always energized and mis-calibrated?    The reason for the "up" position is that that is also the "release filament" position.  The "move" position also applies the filament break..

@igiannakas can you look at the BLOBIFIER change request part...

igiannakas:
@moggieuk Blobifier fan changes are superseded by the blobifier PR I raised as it changes fan sequence and use :)

moggieuk:
@igiannakas can we close this PR then?  If the Blobifer piece is handled elsewhere then I think it can be.. ?

igiannakas:
The blobifier change yes - can be closed. Unsure of the servo changes though!!

moggieuk:
I'm not sure the servo change is correct. @Marek-Dvorny feel free to re-open and explain your thinking, but the default servo position should be "up" which can be used to release filament (it is on the ERCF).  The "move" is an optional alias for "up" than allows repositioning only for selector movement.
Stress should not occur - the "up" position should be a valid position or it suggests some calibration is needed.  Are you using default positions? If so, be sure to calibrate for you exact setup..

PR #765, opened 2025-08-11
feat: Add filament_switch_sensor warnings

3DCoded:
I've had several 3MS users ask for help related to unexpected print pauses, where the solution is to remove all `filament_switch_sensor` sections from the configuration, since they can pause prints unintentionally. 

This PR displays a warning on `MMU_BOOTUP` if there are `filament_switch_sensor`s found in the config that are not created by HH (user-created).


moggieuk:
Good idea. I'll merge after a quick sanity test on the latest v340 release.   Or you can confirm...

Issue #757, opened 2025-07-29
MMU_GATE_MAP Unable edit the filaments in web


Gornerrrrr:
Hi, after upgrading to version 3.3 recently, I found that the _EditFilaments_ function in the web interface is no longer usable.
Klipper: v0.12.0-445
Web UI: Fluidd v1.34.1
happy-hare: v3.3

Open the _EditFilaments_  page on the screen, set some materials and colors, and then click "_Save_". You will find that the save function doesn't work at all.

<img width="798" height="837" alt="Image" src="https://github.com/user-attachments/assets/56322abd-4a4a-4721-8dfb-fb8222e96a1a" />
The comands output after clicking "_Save_":
<code>MMU_GATE_MAP MAP="{0: {'status':1,'spool_id':-1,'material':'PETG','color':'FAFAFAD7','name':'Unknown','temp':'250','speed_override':100},1: {'status':1,'spool_id':-1,'material':'PATG','color':'0D47A1B9','name':'Unknown','temp':'250','speed_override':100},2: {'status':1,'spool_id':-1,'material':'PLA','color':'FF5722FF','name':'Unknown','temp':'210','speed_override':100},3: {'status':1,'spool_id':-1,'material':'PLA','color':'000000FF','name':'Unknown','temp':'210','speed_override':100},4: {'status':1,'spool_id':-1,'material':'Unknown','color':'','name':'Unknown','temp':200,'speed_override':100},5: {'status':1,'spool_id':-1,'material':'Unknown','color':'','name':'Unknown','temp':200,'speed_override':100},6: {'status':1,'spool_id':-1,'material':'Unknown','color':'','name':'Unknown','temp':200,'speed_override':100},7: {'status':0,'spool_id':-1,'material':'Unknown','color':'','name':'Unknown','temp':200,'speed_override':100}}" QUIET=1</code>

So I tried using the direct command setting, and this time it worked:
`MMU_GATE_MAP GATE=0 MATERIAL=PLA COLOR=ff0000 TEMP=205 AVAILABLE=1`

Gornerrrrr:
Now I have checked the code and seem to have identified the cause of the problem:
There is a code segment in the cmd_MMU_GATE_MAP function
<img width="1046" height="199" alt="Image" src="https://github.com/user-attachments/assets/0ecbae4a-c01e-499e-8bf8-2ad079623bb8" />
[...]

PR #767, opened 2025-08-15
fix MMU_GATE_MAP cant update when disable spoolman

Gornerrrrr:
Fixed the issue where opening the ***EditFilaments*** interface on the **_Fluidd_** interface would result in the inability to save the material configuration. 

This is because the code has enabled `spool_id`protection. However, when SpoolMan is not installed or enabled, the `spool_id`defaults to -1, making it impossible to write data normally. 

Modifying the save judgment in the *EditFilaments* interface might also fix this issue. This problem only occurs when the MAP parameter of MMU_GATE_MAP is written.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -7857,7 +7857,7 @@ class Mmu:
                             continue
 
                         # Only update gate attributes if we have a valid spool_id
-                        if fil and fil.get('spool_id', -1) != -1:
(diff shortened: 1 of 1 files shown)


Gornerrrrr:
Fixes #757

moggieuk:
Thanks so much for this!   It just so happens I fixed the problem in the just-released v3.4.0 version.  The logic on that check was indeed incorrect!

Going to close this since already fixed, but re-open if I missed something...

PR #768, opened 2025-08-16
fix: testing things to get stepper management working after the klipp…

ammmze:
…er motion queueing stuff

### DIFF
--- a/extras/mmu_machine.py
+++ b/extras/mmu_machine.py
@@ -293,7 +293,6 @@ class MmuToolHead(toolhead.ToolHead, object):
         self.mmu = mmu
         self.printer = config.get_printer()
         self.reactor = self.printer.get_reactor()
-        self._saved_stepper_kins = {}
         self.all_mcus = [m for n, m in self.printer.lookup_objects(module='mcu')]
         self.mcu = self.all_mcus[0]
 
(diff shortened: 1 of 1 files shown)


extras/mmu_machine.py
```diff
@@ -293,6 +293,7 @@ def __init__(self, config, mmu):
         self.mmu = mmu
         self.printer = config.get_printer()
         self.reactor = self.printer.get_reactor()
+        self._saved_stepper_kins = {}
```
ammmze, line 296: don't need this...left over from one of the earlier failed attempts to get it working

extras/mmu_machine.py
```diff
@@ -552,8 +587,10 @@ def sync(self, new_sync_mode):
             s_kinematics = ffi_main.gc(s_alloc, ffi_lib.free)
             self._prev_sk.append(s.set_stepper_kinematics(s_kinematics))
             self._prev_rd.append(s.get_rotation_distance()[0])
-            following_toolhead.step_generators.remove(s.generate_steps)
-            driving_toolhead.register_step_generator(s.generate_steps)
+            if hasattr(following_toolhead, 'step_generators'):
+                following_toolhead.step_generators.remove(s.generate_steps)
+            if hasattr(driving_toolhead, 'register_step_generator'):
+                driving_toolhead.register_step_generator(s.generate_steps)
```
ammmze, line 593: I suspect we need to do more than this...effectively this is just removing these lines of code for newer klipper.

extras/mmu_machine.py
```diff
@@ -698,7 +738,8 @@ def __init__(self, toolhead, config):
 
         for s in self.get_steppers():
             s.set_trapq(toolhead.get_trapq())
-            toolhead.register_step_generator(s.generate_steps)
+            if hasattr(toolhead, 'register_step_generator'):
+                toolhead.register_step_generator(s.generate_steps)
```
ammmze, line 742: Same as previous comment...probably needs more?

moggieuk:
I don't think this approach is sufficient...  there are two concepts to implement:

1) I want to be able to move a stepper between "rails".  Technically these rails are on different Toolheads but which latest klipper changes they may not need to be (e.g. Manual stepper is now an "additional rail" on the toolhead and appears to support proper drip homing).  The "rail" then drives movement.   This allows the extruder to be thrown on the gear rail to allow the mmu to control the extruder while loading/unloading,  It also allows the gear stepper to be synced with the extruder for in-print syncing. Finally it allows a perhaps strange setup of putting only the extruder on the gear rail. The advantage this last type of syncing is that the filament can be controlled with gear rail control rather than extruder.

2) for type-B mmu's multiple gear steppers are initially defined on the rail and those that are not selected are crippled and re-enabled when needed.

extras/mmu_machine.py
```diff
@@ -458,6 +466,30 @@ def select_gear_stepper(self, gate):
         else:
             self._reconfigure_rail(None)
 
+    def _register(self, toolhead, stepper, trapq = None):
+        # klipper 0.13.0 <= 195
+        # if hasattr(toolhead, 'register_step_generator'):
+        #     if stepper.generate_steps not in toolhead.step_generators:
+        #         toolhead.register_step_generator(stepper.generate_steps)
```
ammmze, line 473: I wonder if we still need to fiddle with the step generators on older versions of klipper or if changing the trapq is sufficient?

PR #771, opened 2025-08-26
Blobifier purge sequence improvements

igiannakas:
Following 8+ months of running the blobifier macros in a semi-custom manner, I am proposing the below changes to the purge sequence and logic:

### Changes:

1. **Retraction in between purge loops (poops)**: Introduced variable to perform a small retraction value in-between the formation of purge poops. This assists un-sticking the blob from the nozzle and reduces stringing when depositing the blob. New variable: variable_retract_between_blobs with 5mm default retraction value.
2. **Fan handling during purging and cutting**: Introduced new variable to set the fan speed to a (higher) value during poop depositing. This helps solidify the blob and deposit it cleanly, while also allowing for a lower fan speed during purging, which helps keep the material soft and adhered to the servo tray. New variable: variable_part_cooling_fan_blob_deposit with 100% fan speed as the default. Now also the purging default fan speed value (variable_part_cooling_fan) is 30%.

### Next steps:

Testers wanted to test these amendments. My tests show results like the below:

IMG_6330

https://youtube.com/shorts/P2c9okgFZ-0

### Further PR's in the pipeline:
1. Implement long retraction and push back to further clean the nozzle when high volume purges are requested
2. Implement pause on purge if runout occurs.

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -208,12 +208,18 @@ variable_eject_hop: 1.0
 # Dwell time (ms) after purging and before cleaning to relieve pressure from the nozzle.
 variable_pressure_release_time: 1000
 
-# Set the part cooling fan speed. Disabling can help prevent the nozzle from cooling down 
-# and stimulate flow, Enabling it can prevent blobs from sticking together. Values range 
-# from 0 .. 1, or -1 if you don't want it changed.
-#variable_part_cooling_fan: -1              # Leave it unchanged
-#variable_part_cooling_fan:  0              # Disable the fan
-variable_part_cooling_fan:  1               # Run it at full speed
+# Amount (mm) to retract between blobs to reduce string formation and allow for a cleaner purge.
+variable_retract_between_blobs: 5
+
+# Set the part cooling fan speed during purging (variable_part_cooling_fan) and
+# during the blob depositing process (variable_part_cooling_fan_blob_deposit)
+# Low fan speeds during purging help ensure the pellets are well formed and reduces
+# the probability of the pellets coming unstuck from tray due to material shrinkage.
+# Setting the blob depositing fan speed to high (100%) helps freeze the blob, unstick
+# the blob from the nozzle and shrink it a little allowing for a cleaner blob depositing
+
+variable_part_cooling_fan:  0.3             # Fan speed to use during purging (-1 is unchanged, 0 is off, 0.1-1.0 for fan speed %)
+variable_part_cooling_fan_blob_deposit: 1   # Fan speed to use during blob depositing (-1 is unchanged, 0 is off, 0.1-1.0 for fan speed %
 
 # Define the part fan name if you are using a fan other than [fan]
 # Applies to [fan_generic] or other fan definitons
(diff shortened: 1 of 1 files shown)


Marek-Dvorny:
I have tested your changes and there are my testing  results: Retraction is in my opinion too high, I use 1mm. Fan speed is very good. However, in my case works better the previous version, because the the retractions make the nozzle "un-primed". I like the 30 % of fan - it's not too fast and slow. I am attaching here my config, I have changed some parameters in the upper config part. For me, it works with smaller blobs and Z raise and higher purge volume and eject hop - when too low, the blob sticks on the brush, when too high, the blob has a "string" and hangs down earlier. If you want to test my upper settings, be careful about the other values because I have another build - 350 mm. I also discovered that setting up Blobifier is very individual, because each user has different build, hotend, nozzle, part cooling fan etc. (about my testing - I have Voron 2.4 350, Stealthburner, Bambulab ObXidian hotend with it's 0.4 mm nozzle, DDCENG radial part cooling fan, Blobifier parts from Seleadlab.
[blobifier(1).zip](https://github.com/user-attachments/files/22025567/blobifier.1.zip)

igiannakas:
Thank you for testing! 

A couple of clarifications please:
1. What do you see when you mean "makes the nozzle un-primed"? There is a de-retraction move before each purge starts, so the nozzle should be fully primed. Do you hear air bubbles basically?
2. Did you test the 100% fan speed during blob depositing? What were your thoughts?

So if I am to summarise, reduce retraction to 2mm (which is the default for HH anyway in when tool changing) as a safe default?

Marek-Dvorny:
Sorry for the response delay,
Better clarification: I would recommend a little (0.3 mm) higher deretraction than the retraction, because when printing materials like PETG with higher temperatures, sometimes the melted filament falls out from the nozzle before brush cleaning (even sometimes when more retracted; when retracted too high, occasionally there is an air bubble - it depends on the filament). I tried the part cooling fan at 100 % during blob depositing, and it works well. I think part cooling fan settings are good. I would suggest a little smaller delay between blob depositing and servo-plate cleaning because of possible filament going out of the nozzle.

igiannakas:
The delay is a bit of a tradeoff. Higher cooling performance fans work with lower delay - I’ll need to test on my setup with a lower delay and see how it works.

in any case the delay is configurable in the parameters.


good idea on the re-priming. Thing is, should this be a variable or a hard coded parameter 🤔

moggieuk:
Is this ready to pull?   The code looks good to me, it's just a question of tuning/defaults?

igiannakas:
Correct - tuning defaults is what is left. I’ll probably add one more option for extra deretract length between blobs to cater for oozing that may happen during the cut process. 

ive also experimented with runout detection as you also had in the template purge routine and works quite well. So will fold that in too.


Give me a couple of days and I’ll get it finished.

Marek-Dvorny:
I think it's good idea to make the deretraction settings configurable, so anyone can configure it for his / her printer.

moggieuk:
@igiannakas is this ready for integration?  Feel free to pull if you want...

igiannakas:
Close. I want to add 1 more feature (pause blobifier on runout) and I’ll merge it.

moggieuk:
Great. I'll leave this for you then...

igiannakas:
@moggieuk lets merge - I'll create a new PR for the runout detection and pause.

PR #779, opened 2025-09-09
Proportional sync feedback sensor integration fixes

igiannakas:
Fixes in HH sync feedback routines to enable proportional sync feedback control via events.

Very very much in test/debug. Raising PR for discussion.

Summary of key fixes:

1. Direction detection – switched to use a configurable sync_movement_threshold so the extruder direction is reliably set even with proportional updates.
2. Watchdog reset – no longer resets on every proportional tick; instead it only resets when the side changes (tension ↔ compression), so extruder movement can accumulate.
3. Applied RD tracking – introduced _rd_applied to track the actual gear rotation distance being applied at runtime (not just the calibrated/default one).
4. Skip redundant RD updates – added an EPS check so the same RD value isn’t re-applied and logged repeatedly.
5. Startup/reset consistency – _rd_applied is now set both during initial gate reset and when resetting rotation distance, so the tracked/live value always matches what’s applied.

### DIFF
--- a/extras/mmu/mmu_sync_feedback_manager.py
+++ b/extras/mmu/mmu_sync_feedback_manager.py
@@ -30,8 +30,7 @@ import logging, time, math
 
 class MmuSyncFeedbackManager:
 
-    FEEDBACK_INTERVAL     = 0.5     # How often to check extruder movement
-    SIGNIFICANT_MOVEMENT  = 5.      # Min extruder movement to trigger direction change (don't want small retracts to trigger)
+    FEEDBACK_INTERVAL     = 0.25     # How often to check extruder movement
     MOVEMENT_THRESHOLD    = 50      # Default extruder movement threshold trigger when stuck in one state
     MULTIPLIER_RUNAWAY    = 0.25    # Used to limit range in runaway conditions (25%)
     MULTIPLIER_WHEN_STUCK = 0.01    # Used to "widen" clamp if we are not getting to neutral soon enough (1%)
@@ -49,6 +48,10 @@ class MmuSyncFeedbackManager:
         self.extruder_direction = 0 # 0 = Extruder not moving
         self.active = False         # Actively operating?
         self.last_recorded_extruder_position = None
+        self._last_state_side = 0   # track sign of proportional state to detect transitions
+        self._last_watchdog_reset = 0.0
+        self._rd_applied = None     # track live applied RD so UI can show true adjustment
+        self._proportional_seen = False  # true once we receive at least one proportional event
 
         # Process config
         self.sync_feedback_enabled = self.mmu.config.getint('sync_feedback_enabled', 0, minval=0, maxval=1)
@@ -56,7 +59,13 @@ class MmuSyncFeedbackManager:
         self.sync_feedback_buffer_maxrange = self.mmu.config.getfloat('sync_feedback_buffer_maxrange', 10., minval=0.)
         self.sync_multiplier_high = self.mmu.config.getfloat('sync_multiplier_high', 1.05, minval=1., maxval=2.)
         self.sync_multiplier_low = self.mmu.config.getfloat('sync_multiplier_low', 0.95, minval=0.5, maxval=1.)
-        self.sync_movement_threshold = self.mmu.config.getfloat('sync_movement_threshold', self.MOVEMENT_THRESHOLD, above=self.SIGNIFICANT_MOVEMENT) # Not yet exposed
+        # Make direction detection threshold configurable and reasonable for proportional updates
+        # Min extruder movement to trigger direction change
+        self.sync_movement_threshold = self.mmu.config.getfloat(
+            'sync_movement_threshold',
+            self.MOVEMENT_THRESHOLD,
+            above=0.5
+        )
 
         # Setup events for managing motor synchronization
         self.mmu.printer.register_event_handler("mmu:synced", self._handle_mmu_synced)
@@ -112,7 +121,9 @@ class MmuSyncFeedbackManager:
 
             # Always set initial rotation distance (may have been previously autotuned)
             if not self._adjust_gear_rotation_distance():
-                self.mmu.set_rotation_distance(self.rd_clamps[gate][1])
+                rd = self.rd_clamps[gate][1]
+                self._rd_applied = rd
+                self.mmu.set_rotation_distance(rd)
         else:
             self._reset_gear_rotation_distance()
 
@@ -154,7 +165,7 @@ class MmuSyncFeedbackManager:
                 self.last_recorded_extruder_position = pos
 
             # Have we changed direction?
-            if abs(pos - self.last_recorded_extruder_position) > self.SIGNIFICANT_MOVEMENT:
+            if abs(pos - self.last_recorded_extruder_position) > self.sync_movement_threshold:
                 prev_direction = self.extruder_direction
                 self.extruder_direction = (
                     self.mmu.DIRECTION_LOAD if pos > self.last_recorded_extruder_position
@@ -212,6 +223,7 @@ class MmuSyncFeedbackManager:
         if abs(state) <= 1:
             old_state = self.state
             self.state = float(state)
+            self._proportional_seen = True
             self.mmu.log_trace(
                 "MmuSyncFeedbackManager(%s): Got sync force feedback update. State: %s (%s)" % (
                     "active" if self.sync_feedback_enabled and self.active else "inactive",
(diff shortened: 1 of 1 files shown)


extras/mmu/mmu_sync_feedback_manager.py
```diff
@@ -212,14 +223,22 @@ def _handle_sync_feedback(self, eventtime, state):
...
                     "active" if self.sync_feedback_enabled and self.active else "inactive",
                     self.get_sync_feedback_string(detail=True),
                     float(state)
                 )
             )
-            self.last_recorded_extruder_position = None # Reset extruder watchdog position
+            # IMPORTANT: Do NOT reset the extruder watchdog every proportional tick.
+            # Only reset on *side* transitions (tension<->compression) so ΔE can accumulate.
+            def _side(v): 
+                return 0 if abs(v) < 1e-3 else (1 if v > 0.0 else -1)
```
ammmze, line 237: I'm curious where the `1e-3` / `0.001` comes from. If I understand correctly that the filament pressure sensor is reporting the amount of pressure seen in either direction, this threshold allows for negligible pressure to count as neutral. In which case it makes sense. Do you foresee a need to have that neutral threshold configurable?

extras/mmu/mmu_sync_feedback_manager.py
```diff
@@ -409,13 +430,18 @@ def _adjust_gear_rotation_distance(self):
...
-                self.mmu.log_trace("MmuSyncFeedbackManager: Slowing gear motor down")
+                self.mmu.log_debug("MmuSyncFeedbackManager: Slowing gear motor down")
             else:
                 # Tension when extruding or compressed when retracting, so decrease the rotation distance of gear stepper to speed it up
                 rd = rd_clamp[2]
-                self.mmu.log_trace("MmuSyncFeedbackManager: Speeding gear motor up")
+                self.mmu.log_debug("MmuSyncFeedbackManager: Speeding gear motor up")
 
-        self.mmu.log_trace(
+        EPS = 1e-4
```
ammmze, line 439: Just curious, what does EPS stand for?

igiannakas:
> This all looks reasonable to me! 🚀 I don't however currently have a sync sensor setup to regression test the basic tension/compression sensors. So hopefully we can get some people to test and confirm.

Ive updated the naming of the threshold parameters and made them "tuneables"

igiannakas:
> I've added a few nitpicky comments, from an "I'm not hugely familiar with the codebase and appreciate readable style" standpoint. Please read without combative tone, the idea is "food for thought" only.

Deeply appreciate the review! Let me work on it over the next few days and I’ll push updates.

burkfers:
I've performed regression testing for a single-switch tension sensor (Annex Belay) in hardware and am happy to report everything still works as expected.

igiannakas:
Updated logic to simplify. Also re-wrote the tension adjustment code as a separate function to not bloat the main extruder load function.

igiannakas:
Revised centring logic post extruder load tested successfully. Significantly faster to centre, cleaner code and more versatile.
image

igiannakas:
Proportional sensor tangle/clog detection feature implemented. To-Do list updated. todo 1), 2), 4) are needed prior to "release" of the feature. ToDo 3) requires further exploration towards its usefulness.

moggieuk:
I'll definitely pull this in the future.. just need to find the right time to do so...

igiannakas:
> I'll definitely pull this in the future.. just need to find the right time to do so...

Not yet please - I need to do far more testing and tuning. Also need to make it “feature complete” as in able to run the full gear rotation distance tuning too. 

Ideally we need more proportional sensors in the wild to test this properly as now it’s just a sample size of 1 and going to sample size of 2 when DW Tas gets his set up.

igiannakas:
PR updated with latest sync feedback file changes and HH main branch. Testing revised RD autotuning and will report back.

igiannakas:
Test feedback time. 

Have completed a few multi color prints over the past couple of weeks and the code is stable and working well. Sync feedback is properly set, keeping the proportional sensor centred, alongside end guard monitoring for clog and tangles. Overall very very pleased with how well that works.

IMG_8219

IMG_8244

Screenshot 2025-10-20 at 16 11 03

**RD Auto tuning:**
Auto calibration is more of a hit/miss. It works, it produces a calibrated RD value as expected but...

The PR uses the existing code to find optimal RD. However **I am getting variations in RD on a print by print basis, depending on what material I print, how fast and so forth**. I would not consider auto cal good as it is for day to day use; however I do wonder whether I am hitting a limitation of the binary search method that happens to oscillate / drift and match the print situation at that time-window.

More thought on RD auto cal is definitely needed. I'd be keen to hear from anyone that is using the auto cal with a dual sensor sync feedback sensor, how accurately it works for them. Test would be to 
1. Deliberately set the RD say 3-4% higher/lower than the calibrated value in the mmu_vars.cfg & restart pointer 
2. Run a large test print (a 200x200mm cube for example) and let it calibrate. 
3. Run this with debug logging to file and console and report back with the MMU.log file to review sensor values and calibrated RD and whether it changes to different values within the same print.

igiannakas:
_(posting also here as discord can have this info drowned out :) )_

### RD Auto tune testing (proportional and switch based)
@moggieuk when you have some time I'd love to hear your thoughts too as well as the community's! I've attached my RD sensor data and graphed them in excel below

Some very interesting observations that may help guide future development i believe.

### Observations:
There are four main phases in the auto tune process

**Phase 1 - the rapid convergence:**
Here there is a pretty "rapid" convergance to a "close enough" RD. This is expected as the RD delta from what the extruder calls for is wide (start) and gradually narrowing down as we bounce through the binary search process.

**Phase 2 - The long dwell....:**
Here a good enough auto tuned value has been found. However the clamps have not converged to the threshold value yet (0.25% delta between high and low).
This means we are not triggering the clamp narrowing as often as the RD is pretty much spot on and there is very very little drift in the system. **In my example I was stuck there for 4 hours while printing, even though the sensor range where compression/tension triggers are approx 3mm!** 

There is very little drift in this phase. **DW Tas also run a test with a dual switch sensor** and, while the system converged rapidly to a good enough RD, the print was not big enough to trigger narrowing of the clamps further due to very, very little drift from neutral. This is a much bigger problem with switch based sensors as, by default, the RD value needs to be "off" to trigger the limit switches, which means the closer to get to the ideal RD the exponentially longer you have to wait for them to hit as the switches have 10+mm of distance from each other.

**Phase 3 - The final convergence and tuned value save:**
Here have finally hit tension/compression after the long dwell. This narrows down the clamps the last remaining % and the tuned value is saved 

**Phase 4 - Post auto tuned oscillation**
Here we start seeing oscillation. The clamps start to diverge and the RD value is out of tune again. After a (long) while it will narrow down and converge again, and then start oscillating again.
Why? because we have narrowed down the fast/slow RD clamp values to such an extent (0.25% of each other) we rely on the "nudger" to slowly open them up again to create a meaningful high/low multiplier to cover slight extrusion variation in the system.

### So what?
1. Phase 1 works well! Great!
2. **Phase 4 - oscillation post tune needs addressing**. Once a tuned value is found, we **should expand the clamps back to the sync multipliers set by the user** but centred around the new tuned value. This would allow the system to "relax" and allow for natural slight deviation to not constantly trigger a clamp convergence, fighting the nudger that tries to open the clamps up. This has been addressed here: https://github.com/moggieuk/Happy-Hare/blob/8c2d74a383e2fcdfee639faa2e7c01ca81ea205c/extras/mmu/mmu_sync_feedback_manager.py#L502
3. **Phase 2 - long dwell needs addressing.** Unless the sync feedback sensor is tiny (proportional or not), there is a big chance that we end up in a long dwell situation that will take hours to get out from. How do we address this? 

- When we trigger an RD adjustment we can start an extruder accumulator. 
- The sensor range is say 10mm.
- The desired precision is 2mm over 1m of filament extruded
- If we have not hit a compression or tension switch after (10/2)*1m = 5m worth of filament, we are pretty confident that this is a tuned RD value within the required tolerance. We could make it even larger - say 10m worth of filament to be even more certain.
- If we do hit a switch, then reset the extruder counter and repeat.

That "soft test" would allow for a good enough RD value to be saved even though the clamps have not formally confirmed this as "correct" by converging down to the defined threshold. **This would prevent the situation where an exponentially larger print is needed the closer the RD converges to ideal in order to confirm it** Why? **The closer to ideal we are, the more extrusion is needed to reach a limit switch and narrow the clamps.**

In my example above the dwell RD was 22.7404 and the tuned RD was 22.7574 -> a difference of 0.07% -> 0.7mm over 1m filament extruded. (or 0.35mm over 1m if we double it).

image
[RD Auto tune.xlsx](https://github.com/user-attachments/files/23144463/RD.Auto.tune.xlsx)

mpsnp:
@igiannakas Have you thought about dynamically increasing multiplier values each time we hit switches to speed up phases 2-4? 

Also, maybe even keep multiplier "turned on" (as it works with single swich design), until we hit opposite swich? 

This way i think tuning can be performed much quicker.

UPD: I mean that with higher multipliers like low:0.2-high:5.0, it will bounce between switches/sensors pretty quickly, and what's important: even small error in RD will add up much quicker.

igiannakas:
Yes I did do that initially. When it crossed a threshold I ramped up RD to attempt to recover. However every time you change RD you flush the klipper queues and that causes TTC errors if they happen too quickly / in fast succession. It also offered zero practical benefit. It wasn’t any more responsive. So net net I removed that ability.

igiannakas:
Closing as a much more advanced method is in testing now.

PR #780, opened 2025-09-09
fix: eSpooler running continuously after EndlessSpool runout occured when burst motor was on

PocketMiner82:
I'm currently modifying my Sunlu S4 filament dryer to have electric respoolers. I've setup Happy-Hare's eSpooler component for that and found a bug because after I had a filament runout and Happy-Hare switched to another Spool in the EndlessSpool group, the eSpooler started unwinding my Spool with full speed:

If one is using in-print burst assist mode, the following can happen:
When there is any MMU operation that uses `self.get_operation(gate)` while the motor is on due to a burst move, the `get_operation` will return the current burst PWM value instead of `0`.
Other parts of code then interpret that value as if the eSpooler isn't set up with **burst assist** but rather with a **constant** in-print speed. Even worse: If one is using a burst_power of 100%, then the get_operation method will return a pwm value of `1.0` indicating full speed.

This is a problem for e. g. a filament runout (which is wrapped by `wrap_sync_gear_to_extruder`) while burst is currently running:
1. The `wrap_sync_gear_to_extruder` wrapper in `mmu/mmu.py`, which uses `get_operation`, caches a non zero pwm (e.g. `1.0` if the assist power is `100%`) with `operation=print`.
2. After the runout handling is complete, the wrapper restores the cached espooler state (e. g. the `1.0` from above) via `set_operation`.
3. Now the espooler unspools filament continuously (with the configured burst power) instead of using burst moves because the pwm used in `set_operation` is no longer zero.

I manged to make the following gcode file, where this issue can be reproduced:
```gcode
; Use eSpooler values that make it easier to show the issue
MMU_TEST_CONFIG espooler_assist_extruder_move_length=10.1 espooler_min_distance=10.1 espooler_assist_burst_power=100 espooler_printing_power=0 espooler_assist_burst_duration=3
; We don't want a clog detection to happen, we know there is no filament loaded
MMU_ENCODER ENABLE=0
; Select Gate 0 and turn on sync
MMU_SELECT GATE=0
MMU_SYNC_GEAR_MOTOR SYNC=1
M83
; make sure the values from MMU_TEST_CONFIG are updated before turning on espooler
G1 E0.01 F300
G1 E0.01 F300
G1 E0.01 F300
G1 E0.01 F300
G1 E0.01 F300
G1 E0.01 F300
G1 E0.01 F300
MMU_ESPOOLER GATE=0 OPERATION=print
G1 E2 F300
G1 E2 F300
G1 E2 F300
G1 E2 F300
G1 E2 F300
G1 E2 F300
G1 E2 F300
G1 E2 F300
G1 E2 F300
G1 E2 F300
G1 E2 F300
G1 E2 F300
G1 E2 F300
; an eSpooler burst move will still be running when this command is called
; this could also be a filament runout instead of this test move command
MMU_TEST_MOVE MOVE=4 SPEED=1
; simulate "print continue"
G1 E11 F300
; Pause for 15 seconds.
G4 P15000
MMU_SYNC_GEAR_MOTOR SYNC=0
```
This gcode uses `MMU_TEST_MOVE` instead of a runout command, because similar to a runout command, `MMU_TEST_MOVE` is also wrapped by `wrap_sync_gear_to_extruder`.

To reproduce the issue (I tested it with an ERCF v2, probably can also be reproduced on a BoxTurtle):
1. Remove all filament from gate 0.
2. Make sure there is no filament roll on the eSpooler of gate 0.
3. Preheat your extruder to the temp of your liking.
4. Copy the above gcode into a `Test_eSpooler_Burst.gcode` file and upload it to the printer.
5. "Print" the gcode file.
6. You will observe that in the 15 second pause, the eSpooler runs continuously. If you try the same steps again with the modifications in this PR, the eSpooler shouldn't run continuously during the 15 seconds.

Note: The gcode file will override some settings to make the error easier to happen. If you want to print again after running the code, restart klipper first.

---

This PR fixes the issue by not changing the pwm value in the `self.operation` list if the requested operation is a burst (`None`) operation.

### DIFF
--- a/extras/mmu_espooler.py
+++ b/extras/mmu_espooler.py
@@ -170,7 +170,7 @@ class MmuESpooler:
                     self.advance(gate)
             elif cur_enabled and not enable:
                 self.burst_trigger_enabled[gate] = False
-                # If motor is running because of burst, disable it and prevent future calls to _reset_burst_trigger turning off the motor unwantedly
+                # If motor is running because of burst, disable it and prevent pending calls to _reset_burst_trigger turning off the motor unwantedly
                 self._reset_burst_trigger(gate, True)
 
 
@@ -300,13 +300,8 @@ class MmuESpooler:
                 _schedule_set_pin('enable_%d' % gate, 1)
 
             # Don't change the operation if it is just an in-print assist burst move.
-            # If we would change operation, the following could happen:
-            # When there is any MMU operation that uses self.get_operation(gate) while the burst move is on, the get_operation would return a non zero value for PWM.
-            # This is a problem for e. g. a filament runout while burst is currently running:
-            # 1. The wrap_sync_gear_to_extruder wrapper in mmu/mmu.py, which uses get_operation, caches a non zero pwm (e.g. 1.0 if the assist power is 100%)
-            #    with operation=print.
-            # 2. After the runout handling is complete, the wrapper restores the cached espooler state via set_operation.
-            # 3. Now the espooler unspools filament continuously (with the configured burst power) instead of using burst moves because the pwm is not zero.
+            # If we would change operation, any MMU operation calling the get_operation(gate) method while the motor is performing a burst move
+            # (e. g. a runout which is wrapped by wrap_sync_gear_to_extruder()) would get the burst pwm value instead of 0
             if operation != None:
                 self.operation[self._key(gate)] = (operation, value)
 


moggieuk:
@PocketMiner82 thanks for the PR (and thanks @ammmze for testing -- I've been on vacation so behind on updates.  I'm pulling now but have not completely verified myself but code review looks fine...

PR #781, opened 2025-09-11
Replace long form grep parameters with short form.

herm:
Some busybox grep versions only accept short form parameters.

```
grep: unrecognized option '--quiet'
BusyBox v1.31.1 (2024-03-15 11:13:52 CST) multi-call binary.
```

Should have no impact on GNU grep as it supports both forms.


moggieuk:
Thanks. I can pull this into 'main' but command line options are always a problem.  Note thatHH v4 will have a completely new installer that we are carefully avoiding unusual options but this could possibly occur again in the future..

PR #791, opened 2025-09-23
Fix led black light variable ignored in certain scenarios

igiannakas:
as per title, the variable black_light was ignored in certain scenarios, despite configured settings by the user. Current code reverted it to off, instead of the black_light value.

### DIFF
--- a/extras/mmu/mmu_led_manager.py
+++ b/extras/mmu/mmu_led_manager.py
@@ -612,9 +612,12 @@ class MmuLedManager:
 
                 elif effect == "filament_color":
                     if gate is not None:
-                        rgb = self.mmu.gate_color_rgb[gate]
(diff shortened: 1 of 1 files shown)


moggieuk:
@igiannakas Thanks .. I pulled your logic but then you inspired me to simplify.  I've already integrated into `main` and it appears to work fine but would you mind reviewing / double checking the original problem your PR fixed. Thanks!!

igiannakas:
> @igiannakas Thanks .. I pulled your logic but then you inspired me to simplify. I've already integrated into `main` and it appears to work fine but would you mind reviewing / double checking the original problem your PR fixed. Thanks!!

Haha for sure! Indeed there was quite a bit of duplication in the code.

Will refresh HH tomorrow morning and let you know if any issues.

PR #794, opened 2025-10-02
Remember correctly the last fully loaded tool if option to retry_tool_change_on_error is set

igiannakas:
Previous behaviour would erroneously set the last tool to be the current tool if a tool change failure happened and the retry_tool_change_on_error was set to 1. This would cause downstream components (blobifier) to purge the wrong filament quantity as to the blobifier it appeared as if no tool change happened.


norris108 (approved):

moggieuk:
Thanks!!

PR #798, opened 2025-10-10
Add condition for Kalico mcu lookup

Crosenhain:
Since Kalico isn't providing the mcu lookup that comes with newer Klipper versions, the latest Happy-Hare update breaks Kalico installations. 

By checking if we're running on Kalico, we can switch back to the old lookup for Kalico installs, and default to the new lookup for Klipper installs.


moggieuk:
Really appreciate the PR, but I fixed before I saw it!  You are 100$ correct on the cause.

PR #803, opened 2025-10-13
Handle 'num_gates' option when MMU backend is not detected

CooperGerman:
Hi,
it's been a while =) I stumbled upon a use case in which i need to define more than 1 gate (being on a idex printer), but i do not have any mmu.

I nonetheless wanted to take advantage of `mmu_server` especially for the spoolman interaction it offers.

I'm proposing a _hidden_ `num_gates` option for the mmu_server module so that one can configure this number when using the mmu_server without the mmu backend.

There might be something easier or cleaner, I'm open to discuss this and would understand if it does not align with Happy-Hare's roadmap.


moggieuk:
Sure, no problem with this.  I don't think I'll document at this point but happy to include it.

moggieuk:
The PR is in draft mode so I can't pull. If you can make final I'll pull right away..

Or you can pull..

CooperGerman:
Just wanted to make sure this aligns with the HH development strategy first. I wasn't sure if the "no mmu" support would be something you wanted to do.

I'll update the branch and merge the PR then if it's ok for you.

PR #810, opened 2025-10-21
Fix led_effect update function

julianschill:
This fixes a compatibility issue with the newest klipper version.

### DIFF
--- a/extras/mmu_leds.py
+++ b/extras/mmu_leds.py
@@ -51,7 +51,8 @@ class VirtualMmuLedChain:
             chain.led_helper.led_state[led] = color
             chains_to_update.add(chain)
         for chain in chains_to_update:
-            chain.led_helper.update_func(chain.led_helper.led_state, None)
+            chain.led_helper.need_transmit = True
+            chain.led_helper._check_transmit()
 
     def get_status(self, eventtime=None):
         state = []


cdedwards:
If I knew more about git, I'd add that mmu_leds.py needs the same fix of of removing the update function and using 
          chain.led_helper.need_transmit = True
            chain.led_helper._check_transmit()

julianschill:
> If I knew more about git, I'd add that mmu_leds.py needs the same fix of of removing the update function and using chain.led_helper.need_transmit = True chain.led_helper._check_transmit()

You are right. I also changed it in mmu_leds.py

julianschill:
Doesn't work with Kalico until they pull the led.py from main Klipper.

burkfers:
> Doesn't work with Kalico until they pull the led.py from main Klipper.

Is there an easy way to make the change conditional? Kalico often takes a while to downstream changes (on purpose), and while this change restores functionality on mainline Klipper, it makes Happy Hare fully unusable on Kalico.

moggieuk:
Awesome. Thank you so much!

moggieuk:
> > Doesn't work with Kalico until they pull the led.py from main Klipper.
> 
> Is there an easy way to make the change conditional? Kalico often takes a while to downstream changes (on purpose), and while this change restores functionality on mainline Klipper, it makes Happy Hare fully unusable on Kalico.

Also, older klipper... all the Crealty folks...  I'm going to look into this switch this weekend.

Issue #633, opened 2025-02-10
Spoolman missing part of the filament consumption


igiannakas:
So while the spoolman spool assignment https://github.com/moggieuk/Happy-Hare/pull/619 fixed part of the issue, I am still missing around 40 grams of filament for the below print:

![Image](https://github.com/user-attachments/assets/1cab1d73-2ed5-4af8-8fae-b3fd72cbdcf2)

![Image](https://github.com/user-attachments/assets/d52d8bed-3829-417a-9a94-746866ed2f25)

![Image](https://github.com/user-attachments/assets/e244cb57-3f5b-4394-8002-c14c855ab3b7)

![Image](https://github.com/user-attachments/assets/84419750-88da-4e62-b260-8817e3db8f16)

Started off with a full spool - 1354grams including desiccant etc. Ended up with 1195 grams, **so consumed 159 grams.**

Slicer had requested a total of 152 grams, with purge being 123 grams and model 11 grams, **so slicer estimate is close enough - 152 vs 159 grams.**

**However spoolman recorded ~119 grams consumed** so a delta of around 40 grams or so, which is larger than the model print, hence purging definately is counted but not all (?). This means approx 13 meters of filament are unaccounted for :)

Any ideas what may be going on here? Anyone else noticed this?

moggieuk:
Hmmm, I thought with PR #619 that the hole was completely closed now.  I.e. Happy Hare is always setting the active spool_id because any printing or purging.  It's possible that there is still an oversight, but are you sure you have the density set correctly in Spoolman?   I honestly haven't looked at how it does the calculation but I thought I remember it using `density x volume extruded`

igiannakas:
Yeah I thought so too - it helps but there is still filament missing :(

The density is set correctly and the same as the slicer - slicer, model and poop reconcile but spoolman doesn’t but only on multi color prints.

I’m thinking whether the unload process is the culprit - maybe the filament should be marked as inactive in spoolman before the unload process begins. I’m planning to run a test on it in a few days when the current print job batch is done and will see. 

Alternatively, we are always leaving a fragment of filament in the hotend that gets purged out but not accounted for in the current spool (the residual filament after cutting) but I’m not sure whether that’s enough to explain the delta. 

Will be testing more and let you know…

moggieuk:
[...]

PR #813, opened 2025-10-24
Fix spoolman error in filament accounting

igiannakas:
Potential fix for https://github.com/moggieuk/Happy-Hare/issues/633

Current code deactivates the filament after the unload process has completed. However this includes the macro performed extruder movements which subtract filament use from spool man.

However observed behaviour is that spool man is under accounting filament consumption, which together with the fact that toolhead unload is performed by the HH "extruder" which doesn't notify spool man, the prime suspect for that incorrect accounting is the cut/form tip macros.

image

This PR is experimenting with moving the filament de-activation before the cut / form tip macro is performed to cross check whether this resolves this slight accounting error.

### DIFF
--- a/components/mmu_server.py
+++ b/components/mmu_server.py
@@ -203,9 +203,11 @@ class MmuServer:
         if not hasattr(self, 'mmu_backend_present'):
             await self._init_mmu_backend()
             if self._mmu_backend_enabled():
+                if self.config.has_option("num_gates"):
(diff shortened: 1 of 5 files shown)


moggieuk:
Cool.  You want precision, hey!  :-)

igiannakas:
Indeed I do haha 😬thank you!

PR #815, opened 2025-10-26
New Type A option that has no servo or espooler

sakitume:
I have a basic mmu system that has no servo or espooler and therefore use option 16 of the installer ("Other / Custom") to install Happy Hare. Without this change I must manually comment out a servo section in `mmu_hardware.cfg` and a servo section in `mmu_parameters.cfg` and  comment out the espooler section in `mmu_hardware.cfg` otherwise I encounter cryptic startup errors.

So in addition to fixing these startup issues I am able to use this new option to develop my custom mmu system without having to edit any of the base config files and can instead define my own custom config files which I include *after* the base config files. This lets me fully customize Happy Hare to work with my system and also easily (re)install Happy Hare on any of my other test machines

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2108,6 +2108,7 @@ questionaire() {
             OPTIONS=() # reset option array
             option TYPE_A_WITH_ENCODER                            'Type-A (selector) with Encoder'
             option TYPE_A_NO_ENCODER                              'Type-A (selector), No Encoder'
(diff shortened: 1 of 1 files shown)


moggieuk:
Note that installation is completely changing in v4 code but happy to add this to v3 for now.

PR #820, opened 2025-10-27
feat(blobifier): allow setting y offset for blobifier and brush

ammmze:
These changes are based on @igiannakas [custom blobifier macros](https://github.com/igiannakas/Voron-backups/blob/fbbabc2590192f0845cb33a478a22adbb719428a/printer_data/config/macros/custom_blobifier.cfg). It allows one to specify an offset subtracted from y max position for purging and brushing. One change from his implementation is the ability to set the blobifier offset independent from the brush offset.

One thing still missing that he's got is the ability to have blobifier clean without changing Z ... at some point i'd like to get integrated in here too along with the option to have the nozzle seal on something other than the blobifier tray...specifically i've got a place next to the brush that is fixed to the gantry, so its always available at any Z position.

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -105,6 +105,9 @@ variable_brush_accel: 0
 # adjustments small though! (0.1mm - 0.5mm) and increase it until it works.
 variable_skew_correction: 0.1
 
+# offset subtracted from the Y max position to position the toolhead during purging.
+variable_y_offset: 0
+
 # These parameters define the size of the brush. Update as necessary. A visual reference
 # is provided below.
 #
@@ -127,6 +130,9 @@ variable_brush_start:           67  # For 300mm build
 
 # width of the brush
 variable_brush_width: 35
+
+# offset subtracted from the y max to perform brushing.
+variable_brush_y_offset: 0
 
 # Location of where to purge. The tray is 15mm in length, so if you assemble it against 
 # the side of the bed (default), 10mm is a good location
@@ -340,7 +346,7 @@ gcode:
     {% set ignore_safe = safe.print_height < force_safe_descend_height_until %}
     {% set restore_z = [printer['gcode_macro BLOBIFIER_PARK'].restore_z,pos.z]|max %}
     {% set pos_max = printer.toolhead.axis_maximum %}
-    {% set position_y = pos_max.y - skew_correction %}
+    {% set position_y = pos_max.y - skew_correction - y_offset %}
 
     # Get purge volumes from the slicer (if set up right. see 
     # https://github.com/moggieuk/Happy-Hare/wiki/Gcode-Preprocessing)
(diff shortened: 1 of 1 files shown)


igiannakas:
Thank you!!!

igiannakas:
For the cleaning part I’m using an “off the blobifier” macro and return the blobifier early - ie before the the brush part.

Debating whether it’s worth having this as a wrap  around, ie rename the blobifier to blobifier core or something like that and then create a warp around where you can add modules - custom park, Mmu adjust tension (for proportional tension compression sensor), clean on bed brush, clean on gantry brush etc.

basically make it slightly more modular as adding more conditionals to the existing blobifier code would make it increasingly difficult to maintain.

thoughts?

ammmze:
> For the cleaning part I’m using an “off the blobifier” macro and return the blobifier early - ie before the the brush part.
> 
> Debating whether it’s worth having this as a wrap  around, ie rename the blobifier to blobifier core or something like that and then create a warp around where you can add modules - custom park, Mmu adjust tension (for proportional tension compression sensor), clean on bed brush, clean on gantry brush etc.
> 
> basically make it slightly more modular as adding more conditionals to the existing blobifier code would make it increasingly difficult to maintain.
> 
> thoughts?

👍🏻 Yea, I'll need to get my head around all of the logic currently in there, but I agree, making it more modular would be nice so you can easily "bring your own [wiper, seal, etc]" macro.

ningpj:
After getting carried away rewriting large chunks of blobifier back in Jan I came to the same conclusion and backed myself into a corner as I ended up taking over all HH / blobifier parking / wiping options to enable left / right hand deployment & chained wiper / nozzle rests (static & gantry)…all because I didn't want to change out klicky so I could install it on the right!!  

I haven’t replicated this PR as yet or perfected safety moves needed to avoid non retractable depressor pins as until recently, I exclusively tip formed and only used EREC (crossbow for the win!!).  That and not being able to test default LH operations. 

Enhancement summary for consideration
- LH & RH deployment
- Static and / or gantry wipers with optional nozzle rests
- Forced HH resync if enabled but inactive before blobbing
- Chained wiping action with straight, zig zag and combo wipe options. Goal was never needing to use a wipe tower: Priority parking (gantry rest > static rest > tray), blob, wipe though static (if fitted & Z > user height threshold), restore Z, wipe and park if gantry & rest before restoring LAST/NEXT pos
- Nozzle shaker to help detach belligerent blobs 
- Prime user definable amount as part of first blob before raising Z (e.g. 5mm) 
- Check if klicky probe randomly attached and bail if gantry wiper/rest fitted (optional)
- Calculate plate exclusion zone for slicer setup
- Convert all speed settings to mm/sec to be consistent with HH
- Other changes to messages, etc

Old video up on github showing concept and general multi wiper / rest operation: https://github.com/ningpj/myVoronMods/tree/main/Blobifier%20Modifications  

Happy to collaborate and share.

PR #821, opened 2025-10-30
correct direction logic in filament movement

CooperGerman:
Litte fix for null range buffers on the autocentering logic.

It seems like in the "null range" case we want to invert direction between the first `homing to compression/tension sensor` move and the final `centering`.

I think that not many people run "null range" buffers so i guess i'll be the the guinea pig for this configuration =)

Let me know if i misunderstood the logic here.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -4672,6 +4672,7 @@ class Mmu:
                 elif tension_active:
                     direction = 1 if use_gear_motor else -1
                     actual,fhomed,_,_ = self.trace_filament_move("Homing to compression sensor", max_move * direction, speed=speed, motor=motor, homing_move=1, endstop_name=self.SENSOR_COMPRESSION)
+                direction = not direction
             else:
                 # Normally configured buffer with neutral (no-trigger) gap
                 direction = 0
@@ -4882,7 +4883,7 @@ class Mmu:
             # Deactivate spool immediately before tip forming/cutting
             # Tip forming/cutting macros use the extruder to execute, hence
             # any retraction / de retraction moves are accounted in Spoolman.
-            # By de-activating early, the retraction performed from the macro 
+            # By de-activating early, the retraction performed from the macro
             # is deliberately not accounted in spoolman
             self._spoolman_activate_spool(0)
 


moggieuk:
H!, long time since we chatted.  The change you propose changes the "direction" after it is used in the `trace_filament_move` command so would only effect the trace log message. Was that intended?

Anyway, I do think that logic is suspect.  JFYI the new `flowguard` branch is a complete rework of syncing feedback - it simplifies (so there is a single reader of combined sensor state) but also allow much more complete support including proportional sensors, autotuning and a new flowguard feature (clog/tangle detection).

This particular method has been moved over so I need to take a look at it. But if you are interested in beta testing the new codebase let me know and I can add you to the test group. That would ensure that everything works for your dual active = neutral setup.

CooperGerman:
> H!, long time since we chatted. The change you propose changes the "direction" after it is used in the `trace_filament_move` command so would only effect the trace log message. Was that intended?
> 
> Anyway, I do think that logic is suspect. JFYI the new `flowguard` branch is a complete rework of syncing feedback - it simplifies (so there is a single reader of combined sensor state) but also allow much more complete support including proportional sensors, autotuning and a new flowguard feature (clog/tangle detection).
> 
> This particular method has been moved over so I need to take a look at it. But if you are interested in beta testing the new codebase let me know and I can add you to the test group. That would ensure that everything works for your dual active = neutral setup.

It has been indeed, my update actually serves to invert the direction after the trace filament move which is in the correct direction but before the `"Centering sync feedback buffer"` buffer move which in the neutral point overlapping buffers seemed inverted.
image

If you want to sign me up for testing of your reworked sync feedback code i'd be glad to help to confirm the directions are correct. 

FYI : i ordered kits to build hall effect proportional buffers so my use case might not be that important anymore (until somebody else builds something similar or uses my current sensor).

moggieuk:
@CooperGerman I haven't pulled this PR because the logic has changed a little and I'd love you to validate.  The "trim" move you are referring to is now only applied if sync_feedback_buffer_range is not 0.  I.e. not the "both sensors active means neutral case".

Please jump over to the `flowguard` branch.  it will rev your version number but I hope to push to main once it has had a few more eyes... currently about 6 active beta testers.

```
cd Happy-Hare
./install.sh -b flowguard
```

CooperGerman:
Yes, no problem as I keep an eye on the discord I figured this was going to be the case. I'll close it as it only served as a reminder. I'll keep you posted if I experience any issues with the flowguard branch !

PR #828, opened 2025-11-10
fix: correct sync feedback status condition

lightoze:
Trivial fix of a typo bug


moggieuk:
Thanks, you are absolutely correct.  I actually had already fixed in the 'flowguard' branch that contains the next update, but I'll pull your request into main now...

PR #831, opened 2025-11-14
Blobifier: implement runout detection and set PA to zero when purging.

igiannakas:
As per title, this PR implements runout detection in a similar manner to the "stock" mmu purge macro. It also sets PA to zero when purging (and resets to previous value at the end) to avoid needless compensation by the extruder when creating blobs at high purge velocity. 

Testers wanted please

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -329,6 +329,12 @@ gcode:
     # Set feedrate to 100% for correct speed purging
     {% set backup_feedrate = printer.gcode_move.speed_factor %}
     M220 S100
+    
+    # Backup and zero Pressure Advance during blobbing
+    {% set backup_pa = printer.extruder.pressure_advance|default(0.0)|float %}
+    RESPOND MSG={"'BLOBIFIER: Setting PA to zero (was %.4f)'" % (backup_pa)}
+    SET_PRESSURE_ADVANCE ADVANCE=0
+
 
     # ======================================================================================
     # ==================== DEFINE BASIC VARIABLES ==========================================
@@ -478,7 +484,7 @@ gcode:
         # Un-retract other than the first purge, to re-prime nozzle
         {% if not loop.first %} 
           M400
-          G1 E{retract_between_blobs} F{sequence_vars.retract_speed * 60}
+          __BLOBIFIER_EXTRUDER_MOVE E={retract_between_blobs} EXTRUDER_SPEED={sequence_vars.retract_speed * 60}
         {% endif %}
 
         # Purge filament in a pulsating motion to purge the filament quicker and better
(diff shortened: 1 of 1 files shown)


igiannakas:
Interesting! I thought that it was necessary to save it as the lender variable would have been overriden by the set PA command? 

Would this restore the slicer dictated value or the default value in the printer setup?

ningpj:
Yes, its what I've been doing in my version. restore feedrate as well like this and whatever value was set as the macro was invoked.   I will simulate a few clogs this morning to double check logic as i havent done so as yet. run out is a bit harder

ningpj:
Hmmm. Interesting. I  dont have pregate sensors (TR setup with Binky) and expected HH would trigger a runout using the gate sensor and/or encoder.  It didnt and kept on feeding.  I did manage to trigger an endguard clog though (still have PR#779 in place) which caused blobifier purge sequence to bypass futher purging, restore PA, etc. and exit.  Since I have blobifier on the right, I cant revert to test mainline code (& conditional logic to resync steppers & report mm³ in progress messages, etc).  PA restored directly using primary printer variable

```
07:34:48 $ BLOBIFIER PURGE_LENGTH=200
07:34:48 // BLOBIFIER: Resynching extruder / gear steppers to assist
07:34:48 // BLOBIFIER: Blob 1 of 2 100.0mm (240.5mm³)
07:34:48 // BLOBIFIER: Blob 2 of 2 100.0mm (240.5mm³)
07:34:48 // probe: TRIGGERED
07:34:50 // pressure_advance: 0.000000
// pressure_advance_smooth_time: 0.025000
07:34:55 // MmuSyncFeedbackManager: EndGuard triggered: Proportional sensor near end of travel during forward feed (accum 7.8mm ≥ 6.0mm; |state|=0.941)
07:34:55 // action:paused
07:35:05 // BLOBIFIER: Blobs in tray: 2/500. Next shake @ 25
07:35:21 // BLOBIFIER: Blobs in tray: 3/500. Next shake @ 25
07:35:23 // pressure_advance: 0.030000
// pressure_advance_smooth_time: 0.025000
07:35:24 !! MMU issue: Endguard detected clog or tangle
```

igiannakas:
Hm interesting indeed - maybe that macro variable is not set when triggering via the encoder?

endguard doesn’t trigger during purge by the way unless you have the blobifier being called in the post load macro hook or running it stand alone  - if you have it in the purge macro hook it gets bypassed to avoid incorrect triggering due to potential under extrusion during purging. It should trigger immediately after the purge.

I’ve also got both enabled on my system, will give another test a go and see what it does.

ningpj:
Called from hh purge macro hook rather than original method.

igiannakas:
The above test was done stand alone purge though, correct? Or in print?

ningpj:
Standalone. Will circle back and test in print in the morning

PR #832, opened 2025-11-14
feat: add REASON parameter to extension macro

vajonam:
allows you to call a macro with reason. This then can be passed on a MQTT message. 

```  
[gcode_macro _SEND_NOTIFICATION]
gcode:
  {% set msg = params.REASON|default("MMU Error") %}
  {action_call_remote_method("publish_mqtt_topic",
                             topic="klipper/print_notifications",
                             payload=msg,
                             qos=0,
                             retain=False,
                             use_prefix=True)}
```

this can be delivered using any number of methods.

image


vajonam:
This is not needed since the reason is part for the `printer.mmu.reason_for_pause` variable and I get this on MQTT.

PR #833, opened 2025-11-15
feat(blobifier): Support gantry mounted brush

ammmze:
When `brush_top` is set to `None`, we won't change Z position (aside from the minimum z clearance) when brushing (i.e. `BLOBIFIER_CLEAN`).

This does NOT bring anything for gantry mounted nozzle seal. I want to keep changes small and manageable.

### DIFF
--- a/config/addons/blobifier.cfg
+++ b/config/addons/blobifier.cfg
@@ -76,8 +76,8 @@ variable_toolhead_y: 50                 # From the nozzle to the front of your t
 variable_force_safe_descend_height_until: 1.0 
 
 # Adjust this so that your nozzle scrubs within the brush. Be careful not to go too low!
-# Start out with a high value (like, 6) and go
-# down from there.
+# Start out with a high value (like, 6) and go down from there.
+# Set this to None if you have a gantry mounted brush and do not want move Z axis for brushing.
 variable_brush_top:             6
 
 # These parameters define your scrubbing, travel speeds, safe z clearance and how many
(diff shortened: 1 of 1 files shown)


igiannakas:
Thank you that's awesome! 

I keep on thinking what is the best way to manage this going forward. One approach is to keep Blobifier "monolithic" ie a single macro that "does it all" and add macro hooks for extensions -> ie park overriding, nozzle clean overriding etc.

Another approach is to focus the macro on blobifier park -> calculate purge volumes -> execute purge. Then wrap this core macro with another that can enable "modules" - park for nozzle seal, or park without one, call the core blobifier, wipe on nozzle brush or wipe on gantry brush etc.

The first option is probably "easier" to code and keep backwards compatible. 

What do you think?

ammmze:
> Thank you that's awesome!
> 
> I keep on thinking what is the best way to manage this going forward. One approach is to keep Blobifier "monolithic" ie a single macro that "does it all" and add macro hooks for extensions -> ie park overriding, nozzle clean overriding etc.
> 
> Another approach is to focus the macro on blobifier park -> calculate purge volumes -> execute purge. Then wrap this core macro with another that can enable "modules" - park for nozzle seal, or park without one, call the core blobifier, wipe on nozzle brush or wipe on gantry brush etc.
> 
> The first option is probably "easier" to code and keep backwards compatible.
> 
> What do you think?

Yea...I think modules would be nice and could make it easy for people to swap out modules if they want.

moggieuk:
Thanks @ammmze .   @igiannakas  has the ability to pull this PR since I've deferred responsibility to him for this module.

igiannakas:
Will do, need to give it a quick test and will pull in.

moggieuk:
@igiannakas  is this ready to merge?

ammmze:
Thanks guys!

Issue #807, opened 2025-10-16
Support [hall_filament_width_sensor] as endstop


vega-d:
Some printers like qidi q1 pro and plus 4 use a hall effect sensor for extruder entry filament sensing, and happyhare simply doesn't comprehend the concept of such a sensor. In theory should be pretty simple - if filament thickness >1mm, then there's filament, if <1mm, then there's no filament. 

I tried implementing a shim using virtual_pins plugin like so

```
[virtual_pins]

[delayed_gcode detect_toolhead_filament_macro]
initial_duration: 0.1
gcode:
  {% if printer.hall_filament_width_sensor.Diameter > 0.5 %}
    SET_VIRTUAL_PIN PIN=toolhead_filament_sensor_pin VALUE=1
  {% else %}
    SET_VIRTUAL_PIN PIN=toolhead_filament_sensor_pin VALUE=0
  {% endif %}
  UPDATE_DELAYED_GCODE ID=detect_toolhead_filament_macro DURATION=0.1

[mmu_sensors]
extruder_switch_pin: virtual_pin:toolhead_filament_sensor_pin
```

but HH crashes when trying to home to a virtual pin so I guess the support has to be added either in klipper to make hall effect sensors to behave more like switches or in happyhare to comprehend this type of sensor.

Wazzup77:
We're working on this (in my fork), implementing an additional sensor class that merges functionality from Klipper's hall_filament_width_sensor with MmuSensors. Somewhat similar to ViVid's analog endstops, except detection happens in adc callbacks instead of adc buttons.

Wazzup77:
@vega-d you can close this

PR #859, opened 2025-12-13
Hall filament width sensor as endstop

Wazzup77:
Adds the ability to use hall sensors as endstops - necessary for Qidi printers, which use a hall sensor in the extruder to detect filament presence.

Developed and tested on the Qidi Plus4, but should work on other printers. Multiple testers on Qidi printers (Plus4 and Q2) are already running it.

Should also work with other setups - the hall sensor can be configured as a toolhead or gate sensor. If anyone is running a separate hall sensor, like it was originally developed (https://www.thingiverse.com/thing:4138933) it might be interesting to remake it into a gate+filament width sensor combo.

Should be safe to pull - it does not run if the hall sensor is not defined in the config.

#807 is related


moggieuk:
Hi!, I'm happy to consider this but have a couple of thoughts:
1) The "klipper-side" endstop detection is a little error prone because it is not supported by the mcu and thus the detection overruns the endstop by a few mm.  It depends on speed but with my "hack" for Vivid I was seeing at least 6mm.
2) I REALLY don't want to duplicate all of the original klipper implementation because it creates a maintenance headache and probable incompatibility (now or future) with Kalico, etc.   I'm pretty sure it isn't necessary anyway -- I think you can just use the minimal endstop implementation logic (as I did with `MmuAdcSwitchSensor`) and just register the pin as one that allows multi-use:
```
ppins.allow_multi_use_pin(pin_name)
```
That way you have a minimal implementation that works alongside the original klipper module.

Along the same thoughts, my `MmuAdcSwitchSensor` was written for the one use case I needed to prove which was for "post_gear" sensor.  It would be great if any endstop sensor could optionally be analog (accepting caveat #1).  That way, the pin setup would be very similar.  If analog then the "allow multi use" is enabled and additional optional parameters would be read.
E.g.  For gate sensor with switches:
```
gate_switch_pin: xxx
```
or for gate sensor with analog/hall effect
```
gate_analog_pin: yyy
gate_analog_pullup_resistor:  (optional)
gate_analog_range: (optional)
...
```

WDYT?

Wazzup77:
I'm ok with this and very much agree with the goal. However:
- your analog implementation is not ideal here - these hall filament width sensors use 2 analog signals to compensate temperature drift. The Qidi Plus4 sensor is placed on the extruder inside a heated chamber. So I'd want to keep a 2 pin implementation somewhere
- I need to investigate the possibility of pin reuse here. Both my implementation and analog buttons you use check the trigger in adc callbacks. The stock hall sensor (which would be kept defined here) also uses these - I need to check whether we can keep this compatible or if the callbacks are overwritten on pin reuse.

I tried to make this as a wrapper for a stock hall_filament_width_sensor defined in printer.cfg, but due to initialization order it did not really work. 

On the overrun - yup, that does happen. If we control the callback implementation then the reporting frequency can be controlled too. In our case it is not too relevant - the sensor is just before extruder gears. I honestly think in the vast majority of cases this will be rather irrelevant, especially if the sensor is placed after an encoder.

tl;dr: I will investigate and try to converge in your suggested direction as much as possible

Wazzup77:
OK after pondering it a bit and I'm pretty conflicted on which direction is best, I see 3 possible ways forward:
- keep the full definition of hall sensor in mmu - it gives us the most control and is probably best performance wise. The interface with Klipper/Kalico is the adc implementation, which of all things is the least likely to be significantly changed I think. It is a large class though and indeed may drift from Klipper/Kalico implementation unless we maintain it with this intention.
- multi use pins to define a minimal (2 adc pin) hall sensor endstop similar to MmuAnalogSwitch sensor - your implementation relies on Klipper's MCU_ADC_buttons class, which does not work for the 2-signal hall sensor setup. Easy to do via ADC callbacks, pretty much just need to strip my proposed implementation. I am a bit concerned about reusing ADC pins due to the callbacks though. Having reviewed related code in Klipper (it's a couple layers so I may be wrong) I think it looks solid and will likely work, so this is more intuition than anything speaking. I feel like if stuff starts going wrong here it will be annoying to debug.
- wrap a hall sensor defined in printer.cfg with a class that looks up an existing hall sensor, lobotomizes any runout/insert gcode and hooks endstop logic into original adc callbacks. It is dependent on knowing the implementation of hall_filament_width_sensor though - we need to know what parameters the class has. I like it as a concept, but it might be a pain to maintain as it's the most dependent on the Klipper hall implementation.

I will test the multi use pins in the next few days. Honestly I think if it works reliably this is more of a question about your vision for HH scope - I think a hall filament sensor could compliment the encoder well, but honestly not a lot of people use it for flow compensation. 

I will for now implement the minimum version you described with multi use as a separate class. If we want to go in this direction it should be pretty simple to handle in one class that just handles slightly differently depending on if 1 or 2 ADC pins are defined. I'd remove the dependency on MCU_ADC_buttons and implement the ADC callbacks directly though. That would also allow to configure the ADC sampling and reporting to mitigate the issue with overrun. If you prefer, I can keep it a separate class to leave the current implementation of analog endstop in.

I defer to you for the decision and will report back when I test your idea.

moggieuk:
Sorry, I'm been really busy with the holidays and "flowguard" release -- haven't forgotten about you ;-)

moggieuk:
I've pulled your PR into a branch `qidi_width_sensor` so we can play with it.  This also merges all the new "flowguard" logic and proportional sync-feedback "buffer" sensor which I didn't want to have to merge again.

I have to put some conditional (kalico) logic:
```
        _kalico = bool(self.printer.lookup_object('danger_options', False))
        # ADC 1
        if self.pin1_name:
            ppins.allow_multi_use_pin(self.pin1_name)
            self.mcu_adc = ppins.setup_pin('adc', self.pin1_name)
            if _kalico:
                self.mcu_adc.setup_minmax(self.sample_time, self.sample_count)
            else:
                self.mcu_adc.setup_adc_sample(self.sample_time, self.sample_count)
            self.mcu_adc.setup_adc_callback(self.report_time, self.adc_callback)
```
 setup_minmax() no longer exists in klipper.  Rather than if `_kalico` maybe an simple `has_attribute` test is safer (but I don't have older klipper to test). Feel free to change.
 
 Also, the is a startup error as a result of not having a "switch_pin" on the RunoutHelper.  I'm not sure how you got past this, but in `mmu_sensor_manager`:
 ```
                 if sensor.__class__.__name__ in ["MmuAdcSwitchSensor", "MmuHallEndstop"]:
                    sensor_pin = sensor.runout_helper.switch_pin
                    mcu_endstop = self.mmu.gear_rail.add_extra_endstop(sensor_pin, name, mcu_endstop=sensor)
```
This adds the "switch" as an endstop to the gear stepper(s). Without this, homing can't work..??

Also, your comments say adc2 is optional but the parameter definition is forced so I doubled up in my quick test setup...

Anyway, can you checkout the `qidi_width_sensor` branch.  You will need to run:
```
./install.sh -b qidi_width_sensor
```
because the HH version number will rev owing to a minor addition to `mmu_parameters.cfg`.

My quick (sanity test) config
```
hall_sensor_endstop: gate
hall_adc1: PC0
hall_adc2: PC0
```

Let me know..

**Comment by Wazzup77** [comment]
Sorry, I am a bit messy with Github. Did a bit more experimenting on this branch to get your multipin approach working - didnt realize the previous commit showed up here.

BTW no pressure on the timeline for this - we have a fork going for our work on the Qidi Box Happy Hare mod, this is just to make it simpler for when we release. And who knows, others may benefit (we already have a use getting an MMX working on a Qidi printer thanks to this).

I also managed to get the multi use pins working. So your approach of making it work very similarly to your analog sensors is on the table and I am testing it right now. So the configs being like:
```
# for simple switch
pre_gear_switch_pin_0: PA0 

# for ViVid analog sensors
pre_gear_switch_pin_0: PA0
pre_gear_analog_range_0: 0, 500 
pre_gear_analog_pullup_resistor: 4700. # optional

# for hall sensors
pre_gear_switch_pin_0: PA0
pre_gear_switch_pin_0: PA1 # if 2nd pin detected, hall sensor is used
pre_gear_analog_range_0: 0, 500 
```

setup_minmax vs setup_adc_sample i solved via hasattr.

The sensor factory had to be updated, as the init grew quite a bit. So while we're here I made analogs and halls work for all sensor positions.

This is still WIP, since I was working remotely (got a VPN to the workshop) and need to test this when I get back to the machine.

**Comment by Wazzup77** [comment]
One more thought - there are now 2 different implementations of ADC triggering in the sensors. Your analog sensors use Klipper/Kalico MCU_ADC_buttons (which handle adc callbacks), my hall sensor implements the ADC callbacks inside the class.

Your approach (adc buttons):
- is simpler
- has less control over working parameters
- is dependent on the adc buttons implementation (which may be good or bad - it is a layer of abstraction for the adc itself)

My approach (adc callbacks):
- lets us have full control over the parameters
- is a bit more complex (need to implement adc callbacks)
- has the setup_minmax issue (i.e. Klipper changed ADC implementation circa 2024 and we need to work in compatibility; looks easy)

I'm thinking if it's a good idea to move analog buttons to the adc callback implementation to make it consistent and work some configuration of ADC parameters in there - to manage the overrun issues. ADC sampling is so fast it's pretty much free, so it's honestly just the callback logic that limits reaction time, and how often you can do that depends on the MCU; ADC button refreshes every 15 ms, so roughly 66 Hz. On the Qidi Box, which is an STM32F401 we're using the hall sensor at 100 Hz (2 callbacks) with no issues, can probably test how far we can go. Less reaction time should mean less overrun.

**Comment by Wazzup77** [comment]
@moggieuk Please review. Overview for your convenience:
 - decided to go rework `MmuAdcSwitchSensor` for consistence - should be faster, with slightly less overrun. Both hall sensor and adc switch use the same approach, with a new `MmuAdcSensorBase` class that handles setup to make maintenance a bit easier (also enables the check in mmu_sensors_manager.py to look for that base class)
 - intentionally - the trigger checking in adc_callbacks is inlined for performance. It could be split into helper methods for more clarity and to avoid repeated code, but that would come at the cost of performance (did a bunch of benchmarks to assess this, probably the most interesting part of this project). Since this is in callbacks I propose implementing it inline. Performance is not super important for the standard user, but...
 - added the option to configure ADC - sample time, sample count and report time. This effectively allows overclocking of the ADC, which can be used to control overrun. At default the config checks threshold 33% more often than adc switch did (10 ms vs 15 ms).  ADC can be overclocked quite a lot it turns out. I tested running at 2 ms reporting time, the overrun seems to be greatly reduced with no consequences (didn't see any "too far in the past" errors despite the Qidi processors being rather poor)
 - hall sensors and analog sensors should work everywhere now. Rewrite the sensor creation methods to also handle different sensor types (if analog range & 2 pins defined - Hall, elif analog range - Analog, elif simple switch)
 - added a "Raw" readable status value - can be read with `MMU_SENSOR_STATUS` (optionally with `SENSOR='X'` to read just one sensor) or with any other macro that reads variables, like the commonly used `SEARCH_VARS`
 - small change in MmuRunoutHelper - it kept crashing for me when it was the first sensor initialized (`TypeError: cannot unpack non-iterable NoneType object` on the `_, prev_values = prev line`). I don't 100% understand how that worked previously with `MmuAnalogSwitchSensor` tbh, which worries me a bit because I may
 - you had a typo in your configuration parameters - "X_analog_pullup_resistEr", fixed that to "...resistor"

Tested on my Qidi Plus4. Config looks like this:
```
extruder_switch_pin: PA2
extruder_switch_pin2: PA3 # if this line was missing, sensor would be registered as MmuAnalogSwitchSensor
extruder_analog_range: 13000, 20000 # a_min, a_max
extruder_adc_settings: 0.001, 4, 0.005 # optional ADC settings! sample_time, sample_count, report time
```

Hopefully didn't mess anything up here! I also am ready to document this (on the wiki?) if you have any guidelines for that.

Would appreciate testing on ViVid. I only have a Qidi here, so I tested the Adc switches on the hall sensor in a rather crude test case.

Wazzup77:
Now that I think about it, the 2nd switch pin can probably be defined cleaner (as a list?). But I'm not gonna change anything - wouldn;t want to make a mess before you get a chance to look through it

moggieuk:
Sorry. Just you let you know I will get back to this as soon as I have time -- haven't forgotten about it ;-)

Wazzup77:
@moggieuk Any change this could make it in with the dryer/heater release? No pressure, just an ask for my convenience.

Wazzup77:
@moggieuk I see a version of the hall sensor in the new release - although it's the first one. Is this ready to close or do you want to also work in the minimal (optimized) version and sensor factory refactor?

Wazzup77:
Closing it with the intent to repost the PRs as 3 discrete changes:
 - hall endstop optimization (more predictable and configurable timing; no unnecessary features from Klipper)
 - common MmuAdcSensorBase class to ease future maintenance (avoidng repeated code)
 - sensor factory rework (mostly just FYI for v4)

PR #860, opened 2025-12-13
Implement has_sensor method, returning self.sync_feedback_enabled to suppress errors as temp workaround

ningpj:
Added method to object to suppress sensor manager errors.

Review broader sensor manager context and required changes to enable ``MMU_STATUS SHOWDETAIL=1`` and other upstream dependencies.


moggieuk:
Created a full fix for this so going to close this "patch" PR..

PR #863, opened 2025-12-18
Adjust MMU_CALIBRATION_PSENSOR compression and tension limit calculations

ningpj:
Hand checked (not tested): Remove ``s_maxrange`` var, replacing with inline calcs. Apply ``MAX_MOVE_MULTIPLIER`` to stepped distance to ensure moves match max attempted distance reported previously. Update summary to use ``c_prev`` & ``t_prev``.


ningpj:
Should we also apply a sane upper limit for move gcode arg distance??? e.g. 

``move = gcmd.get_float('MOVE', self.sync_feedback_manager.sync_feedback_buffer_maxrange, minval=1, maxval=20)``

PR #885, opened 2026-01-25
fix: correct the led spec for per-gate leds with multiple units

ammmze:
In my config i've got a 4 lane BT and currently a 1 lane EMU, here's an excerpt of the led configuration:

```
[mmu_leds unit0] # 4 lane BT ... 1 led per lane
exit_leds: neopixel:mmu_leds (1-4)

[mmu_leds unit1] # 1 lane EMU ... 3 led per lane (2 in button, 1 illuminating spool)
exit_leds: neopixel:emu0_leds (1-2)
entry_leds: neopixel:emu0_leds (3)
```

During initialization, HH would try to set a per-gate effect on the exit_leds which then triggered an internal error while running `_MMU_SET_LED_EFFECT` because it was trying to set led index 4, which doesn't exist on a 3 led chain.

Through some debugging, I found it trying to create `mmu_breathing_red_entry_4` with a spec of `unit1_mmu_entry_leds (5-6)` (which would translate to led index 4-5...both invalid).

Now with this fix it is creating it with a spec of `unit1_mmu_entry_leds (1-2)`.

### DIFF
--- a/extras/mmu_led_effect.py
+++ b/extras/mmu_led_effect.py
@@ -88,7 +88,7 @@ class MmuLedEffect:
                         # Per gate
                         if segment in MmuLeds.PER_GATE_SEGMENTS and (not define_on or 'gates' in define_on):
                             for idx in range(mmu_unit.first_gate, mmu_unit.first_gate + mmu_unit.num_gates):
-                                led0 = idx * leds_per_gate - mmu_unit.first_gate + 1
+                                led0 = (idx - mmu_unit.first_gate) * leds_per_gate + 1
                                 led_spec = str(led0)
                                 if leds_per_gate > 1:
                                     led_spec = "%d-%d" % (led0, led0 + leds_per_gate - 1)


ammmze:
FYI, the error in the log didn't include the stack trace, while debugging, I did update it to print the stack trace to help with debugging. Here is the stack trace I got:

```
Internal error on command:"_MMU_SET_LED_EFFECT"
  File "/home/pi/klipper/klippy/gcode.py", line 223, in _process_commands
    handler(gcmd)
  File "/home/pi/klipper/klippy/gcode.py", line 151, in <lambda>
    func = lambda params: origfunc(self._get_extended_params(params))
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pi/klipper/klippy/gcode.py", line 161, in <lambda>
    handler = lambda gcmd: self._cmd_mux(cmd, gcmd)
                           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pi/klipper/klippy/gcode.py", line 326, in _cmd_mux
    values[key_param](gcmd)
  File "/home/pi/klipper/klippy/extras/mmu_led_effect.py", line 673, in cmd_SET_LED_EFFECT
    self.set_enabled(True)
  File "/home/pi/klipper/klippy/extras/mmu_led_effect.py", line 636, in set_enabled
    self.handler._getFrames(self.handler.reactor.NOW)
  File "/home/pi/klipper/klippy/extras/mmu_led_effect.py", line 337, in _getFrames
    chain.led_helper.led_state[index] = (0.0, 0.0, 0.0, 0.0)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
IndexError: list assignment index out of range
```

Another thing we should consider is looking at why this error causes klipper to shutdown.

PR #888, opened 2026-02-06
bigtreetech VVD gate pin fix

luxflow:
https://github.com/bigtreetech/BIGTREETECH_MMS/blob/aab0edb4bde351f4ac6dff77f2aff9cd3028a469/config/bigtreetech-mms/hardware/mms-slot.cfg#L57

### DIFF
--- a/config/base/mmu_hardware.cfg.vvd
+++ b/config/base/mmu_hardware.cfg.vvd
@@ -340,7 +340,7 @@ effect_gate_empty_sel:     mmu_ready_orange2,       (0.1, 0.04, 0)
 # Define any additional hardware for this MMU unit, e.g. heaters, chamber temperature probes, etc
 #
 [temperature_sensor MMU_vivid_left]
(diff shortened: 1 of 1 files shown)


luxflow:
and it seems after klipper commit "1f43be0b8b55d90753578d06ac06356d1ab9a768" (aht30_patch_id)
it is ok to just use AHT3X, no more patch is needed

https://github.com/bigtreetech/BIGTREETECH_MMS/blob/aab0edb4bde351f4ac6dff77f2aff9cd3028a469/install.sh#L72-L73
https://github.com/bigtreetech/BIGTREETECH_MMS/blob/aab0edb4bde351f4ac6dff77f2aff9cd3028a469/config/bigtreetech-mms/hardware/mms-heater.cfg#L16-L24

moggieuk:
I can pull this, but note that the sensor type change will limit VVD to new klipper installs.   My concern would be for those running Kalico and also a large community that has to use an older klipper with Creality machines....

Perhaps add a comment with the option for AHT10 if AHT3X not available?   (Happy Hare internally does the instruction mapping).

PR #889, opened 2026-02-06
add gate available_status for leds

burkfers:
*ready for review other than a wording issue I'd like feedback on*

I wanted to have the LEDs inside my EMU boxes illuminated based on gate availability, but without the transient statuses (preloading, loading, etc). Just one effect when filament is available, and another when it's not.

`effect_gate_available2` is a silly name, but I can't come up with anything better that's not ludicrously wordy. Draft status purely for that. I'd be happy to have suggestions for a more proper wording!

A branch based on `flowguard` is available at [`burkfers:flowguard_available_status`](https://github.com/burkfers/Happy-Hare/tree/flowguard_available_status) for easier testing; I'm sure we're all loving the beta branch.

### DIFF
--- a/config/base/mmu_hardware.cfg
+++ b/config/base/mmu_hardware.cfg
@@ -425,8 +425,8 @@ frame_rate: 24
 #
 enabled: True				# True = LEDs are enabled at startup (MMU_LED can control), False = Disabled
 animation: True				# True = Use led-animation-effects, False = Static LEDs
-exit_effect: gate_status		#    off|gate_status|filament_color|slicer_color|r,g,b|_effect_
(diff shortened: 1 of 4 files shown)


burkfers:
I misunderstood how the existing `gate_status` already does what I want, and that the loading and unloading animations always play on the exit, regardless of the effect there.

My desired configuration can already be achieved by swapping the definitions of entry and exit LEDs, using `gate_status` for the boxes and a static dim white for the buttons. Loading/unloading/heating animations now play on the buttons which are dim white otherwise, and the boxes are illuminated with filament present and off otherwise.

PR #894, opened 2026-02-11
Form tip standalone temp fix

53Aries:
Fix premature temperature change during form_tip_standalone.

Using form_tip_standalone calls line 5058 of mmu.py which checks extruder temperature. It is wrongly getting the temp of the NEXT filament and updating to it, this happens BEFORE tip forming.
The real issue is that by the time form_tip_standalone is called, the self.gate_selected has already changed to the next gate so the self._ensure_safe_extruder_temperature gets the temp of the next filament, not the one currently loaded.

- Added verify_only parameter to _ensure_safe_extruder_temperature()
- When verify_only=True, checks temp is safe without changing setpoint
- Updated form_tip_standalone() to use verify_only=True (line 5060)
- Prevents changing to next filament's temp before tip forming completes
- Temperature now correctly updates during _load_extruder() for new filament
- Maintains safety check to prevent cold extrusion damage


53Aries:
Do not merge yet. There are issues. Don't have time to fix right now but will get to it soon.

moggieuk:
> The real issue is that by the time form_tip_standalone is called, the self.gate_selected has already changed to the next gate 

Are you sure about this?  It doesn't seem possible....  but if there is something going on there then that is the way to fix.   We need to be really careful about "workarounds" and keep the logic as "sequential" as possible..

moggieuk:

```sh
git remote -v
git fetch upstream
git switch <your-pr-branch>
git rebase upstream/v3
```

```sh
git add <resolved-files>
git rebase --continue
```

```sh
```

```sh
git fetch upstream
```

PR #899, opened 2026-02-15
Same Tool Reload Fix

ModularPrintingSystem:
I had a Problem, where my Filament was already loaded and when starting a Print, it would unload the Filament to load it again. I fixed it and it workes fine for me.

(I use a ERCF V3.0)


moggieuk:
I think this logic makes sense.  I tend not to leave my printer with filament loaded so have never noticed the behavior before.

Would you mind just updating the PR against the latest `main` branch (now the v3.4.2 version) so the merge is correct and won't remove the new MMU_QUERY_SENSOR macro for example...

ModularPrintingSystem:
I'll do it.

I also have to execute `MMU_Recover Loaded=1` after I started the printer. 

I'll also add a function for that, which you have to set a variable to true in the configs.

ModularPrintingSystem:
I think it's now up to date. Can you verify it? I'm fairly new to github with push/pull.

Till now I didn't add the named variable. I only updated it for the MMU_QUERY_SENSOR.

ModularPrintingSystem:
I added and testet the Variable and it works.

moggieuk:
Hi!  There is definitely merit in the detection of a pre-loaded tool, but I'm not confident of the fix....

Firstly there is already startup logic in HH to recover the filament position.  It's in `mmu.py` around line 1307 if you are interested.  HH will set LOADED=1 if a sensor indicates that (extruder entry or toolhead).  If not it will do a "MMU_RECOVER" without heating extruder to try to figure out the position. 

So adding:
```
variable_recover_mmu_at_startup             : False	; True/False, Run MMU_RECOVER LOADED=1 on Klipper startup
```
to force the loaded state isn't a good idea for a general purpose solution.  If in your case you really want to force this you can just have the delayed gcode command or you can hook your macro into `user_print_state_changed_extension` and check for the state of "initialized" which happens only on startup / klipper reset.  See `mmu_state.cfg` `_MMU_PRINT_STATE_CHANGED` macro for how your extension hook will be called.

Also, I don't think forcing a 10mm purge is correct either.  The assumption would be that uses have a pre-print prime line or custom purge (think of Blobifier, Gooseneck, etc).

So I think **the really valuable part of the PR** is in detecting that same initial tool loaded and then force skipping the "check_gates" step...  can you separate that out because that would be an easy pull...

ModularPrintingSystem:
Yeah, the `MMU_RECOVER` will be executed on startup, I don't have any extra toolhead sensor. I always got an error message in the console, that the state couldn't be recovered and I always had to manually execute `MMU_RECOVER LOADED=1` in the console.

I'll change the code to only the same Tool fix.

I'll also remove the 10mm purge for now.
I have no pre-print prime line, because it's out of the print area and I only scan the bed in the area I want to print. Should I add another variable for that? Maybe some users have some kind of setup. (I have a VzBot with a watercooled hotend and cannot automatically cut the Filament off near the nozzle (like Filametrix).

ModularPrintingSystem:
Now the rest of the Code is up to date and I only added the same tool fix.


What to you think about an variable called `start_prime_on_skip` for the purge?

ModularPrintingSystem:
I added a separate uninstall script.

moggieuk:

```sh
git remote -v
git fetch upstream
git switch <your-pr-branch>
git rebase upstream/v3
```

```sh
git add <resolved-files>
git rebase --continue
```

```sh
```

```sh
git fetch upstream
```

PR #900, opened 2026-02-15
feat: push gate info to moonraker lane_data to support new orca feature

ammmze:
The beta version of OrcaSlicer introduced the ability to sync filament information from various systems. Upon first look, it had appeared that they were using something specific to AFC, so I had [implemented support for HH](https://github.com/OrcaSlicer/OrcaSlicer/pull/12307), but after that was merged, I was made aware that the AFC folks are attempting to normalize/standardize on pushing the information to the moonraker database and then Orca will fetch from there. So we will probably revert that Happy Hare specific code from OrcaSlicer in favor of pushing gate info to moonraker.

https://github.com/user-attachments/assets/e77f41d2-f7b3-4310-93c8-e517f26b212b

Screenshot 2026-02-15 at 3 18 29 PM


moggieuk:
Awesome.  I think this looks good so will pull, but don't have a good way to test at the moment.

Issue #908, opened 2026-02-25
[Bug]: Klipper v0.13.0-557 unhandled exception - mmu_sensors.py


ningpj:
### Summary

Latest klipper (v0.13.0-557) unhandled exception on startup

### Steps to reproduce

Upgrade from `v0.13.0-536` to `v0.13.0-557` with HH `v3.4.2-8` & restart klipper

### Expected behavior

HH starts up without unhandled exception / fault

### Actual behavior

Klipper faults and reports 

```
Unhandled exception during connect
Traceback (most recent call last):
  File "/home/pi/klipper/klippy/klippy.py", line 130, in _connect
    self._read_config()
  File "/home/pi/klipper/klippy/klippy.py", line 123, in _read_config
    self.load_object(config, section_config.get_name(), None)
  File "/home/pi/klipper/klippy/klippy.py", line 112, in load_object
    self.objects[section] = init_func(config.getsection(section))
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pi/klipper/klippy/extras/mmu_sensors.py", line 794, in load_config
    return MmuSensors(config)
           ^^^^^^^^^^^^^^^^^^
  File "/home/pi/klipper/klippy/extras/mmu_sensors.py", line 695, in __init__
    self.sensors[Mmu.SENSOR_PROPORTIONAL] = MmuProportionalSensor(config, name=Mmu.SENSOR_PROPORTIONAL)
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pi/klipper/klippy/extras/mmu_sensors.py", line 271, in __init__
    self.adc.setup_adc_callback(self._report_time, self._adc_callback)
TypeError: MCU_adc.setup_adc_callback() takes 2 positional arguments but 3 were given
webhooks client 548481672336: New connection
webhooks client 548481672336: Client info {'program': 'Moonraker', 'version': 'v0.10.0-8-g293a4cf'}
```

### MMU Type

Tradrack

### Happy Hare version in '3.1.4-16' format

3.4.2-8

### Klipper version in '0.13.0-121' format

0.13.0-536

### Relevant logs (`mmu.log`)

```shell

```

### Checklist

- [x] I searched existing issues
- [x] I can reproduce with the latest version of Klipper and Happy Hare

ningpj:
[...]

PR #911, opened 2026-02-28
Fix(ADC sensors): Added support for yet another adc setup API signature (klipper change)

moggieuk:
Also addressed in embedded led-effects ahead of real module update.

This addresses issue #908


ningpj:
ADC sensor isnt returning any values post function call fix.
Klipper: `v0.13.0-557` HH: `v3.4.2-10`

```
$ mmu_query_psensor
 PSENSOR Enabled: True  Value: 0.0  Raw Value: 0.0
```

ningpj:
Issue resolved - ADC positional data looks good now.

PR #913, opened 2026-02-28
Bug fixes

GioeEng:
fix: In New klipper (>= v0.13.0-557) setup_adc_callback accepts a single array parameter

fix: In Spoolman the json material property is nested under filament


moggieuk:
Hi! I just beat you to a complete fix for ADC sampling change, so could you reduce this PR to just the mmu_server / moonraker fix and I can pull..

Issue #856, opened 2025-12-10
Add Espooler speed variable using spoolman


crackmonkeyy:
### Summary

.

### Problem statement

I'm hoping you can add an option for the espoolers to use the weight from spoolman to adjust the speed. For example for a full spool the espooler speed is say 50% of that of an empty spool with the percentage adjusting slowly as the filament is used. Thanks in advance

### Proposed solution / approach

.

### Priority / Impact

not a bug

### Checklist

- [x] Have you running the lastest Happy Hare release
- [x] I searched existing issues
- [x] I have asked on the Happy Hare Discord
- [ ] I’m willing to contribute a PR

PR #915, opened 2026-02-28
fix: In Spoolman the json material property is nested under filament

GioeEng:


moggieuk:
I think I follow this logic but can I request two things:
1) mmu_server.py is fine for python3, but can you ensure the klipper code (mmu.py, espooler.py) works with Python 2.7. I.e. avoid f-strings.
2) Can you give me some doc to better explain the two new parameters for calculating burst duration .. I will need to update this wiki page: https://github.com/moggieuk/Happy-Hare/wiki/Espooler-Support#mmu_parameterscfg

Thanks.

moggieuk:
BTW Keep PR as small and separate as possible.  Because of my complete HHv4 refactor I'm having to hand merge into the v4 codebase and it's quite hard not to miss anything..

GioeEng:
I haven't opened a pull request yet. The code has been pushed to the repository only to track commits.
Hardware context: Type B MMU (BoxTurtle).

The current burst assist uses fixed parameters, which creates an unresolvable conflict between a full spool and an empty spool:

With a full spool (≈1 kg): parameters optimized for an empty spool cause excessive burst → **filament overfeeds and interferes with adjacent spools**.
With an empty spool (reduced weight): parameters optimized for a full spool are insufficient → the spool **falls inside the box and jams the mechanism**.

Proposed solution:
Implement a weight-based dynamic burst assist based on the remaining spool weight (retrieved from Spoolman):

Weight > 500 g: burst disabled. The spool is heavy enough to remain stable by inertia.
Weight ≤ 500 g: burst is calculated dynamically based on the remaining filament diameter on the spool, ensuring a consistent buffer and preventing the spool from falling.

Status: Logic already tested on a previous version of Happy Hare. Current implementation is pending testing. Reference: issue #856.

See the Espooler Print Assist section on the website [].](https://www.armoredturtle.xyz/docs/afc-klipper-add-on/features.html#direct-drive)

GioeEng:
I hadn't considered Python 2.7 compatibility.
By next week, I will refactor the code to remove f-strings and validate it on my MMU.
Unfortunately, the current implementation is also broken — prints consistently stall when using full or nearly empty coils.

Thanks.

moggieuk:
No problem.  I'll wait for your signal that things are stable..

GioeEng:
I ran the tests on my BoxTurtle unit today, including validation with all parameters disabled (set to 0). I applied several software adjustments, and the system now operates correctly under both fully loaded and empty spool conditions.
For reference, I have embedded inline documentation within the mmu_parameters.cfg file to facilitate understanding of the operational logic.
I would suggest considering its integration into the existing wiki, though I leave that decision to your discretion.
Thanks again — you did a great job with this software!

PR #924, opened 2026-03-11
feat: Detect if MMU and printer MCU's have same serial

3DCoded:
Similar to #765 , this aims to make the overall user-experience with HH smoother. Since some users prefer to run their MMU off the printer board instead of an external MCU, they won't need the MMU MCU configuration, but the installer may fill in the existing MCU's serial path in. In this case, Klipper will give a somewhat cryptic error in the klippy.log: "Resource temporarily unavailable". This update will have HH display a more user-friendly message instructing users to either change the [mcu mmu] serial address or to remove it.


3DCoded:
I'm converting this to draft since there are a few kinks I have to work out with this (namely, there are no safetys in place should the `[mcu mmu]` section not exist or if users use CANBUS instead of USB/UART)

moggieuk:
I know this PR is a little old now, but I think we need a slightly different approach for v4...

Current the menuconfig will ask for the MMU's serial or canbus connection.   It should be possible to bypass this and select "printer's mcu".  Then the user should be clearly nofiied to check all the pins in the "pins" section for features selected.

I'd love to support this in v4, but I don't want to continue to evolve v3....

3DCoded:
Good point about allowing users to use the printer's main MCU. But in that case, shouldn't the [mcu mmu] section be omitted entirely? I can take a look at the v4 code and see if it's possible to display a warning instead of an error.

moggieuk:

```sh
git remote -v
git fetch upstream
git switch <your-pr-branch>
git rebase upstream/v3
```

```sh
git add <resolved-files>
git rebase --continue
```

```sh
```

```sh
git fetch upstream
```

PR #936, opened 2026-04-05
v4 ported - Utilise proportional sync sensor for senseless toolhead - extruder homing, bowden calibration, unload verification, filament position recovery.

igiannakas:
## Experimental feature: use proportional sync feedback sensor for sensorless toolhead — extruder homing, bowden calibration, unload verification, and filament position recovery

This PR introduces a virtual endstop driven by the proportional sync feedback (PSF) sensor. It enables filament homing to the extruder without a physical extruder-entry switch.

When filament reaches the extruder entrance and begins to compress the buffer spring, the sensor value rises toward 1.0. When the mapped value crosses a configurable threshold, the virtual endstop "trips" and the gear-rail homing move completes — homing to the extruder entry using only the PSF.

### Implementation details

**Virtual endstop:** `MmuProportionalExtruderEndstop` (`mmu_sensors.py`)
A virtual endstop implementing the Klipper endstop interface (`query_endstop`, `home_start`, `home_wait`, `add_stepper`, `get_steppers`). It registers as a listener on the existing `MmuProportionalSensor` ADC callback chain via `register_endstop_listener()`. When the mapped sensor value crosses the configured threshold the endstop triggers, completing the homing move.

**Listener hook on `MmuProportionalSensor`** (`mmu_sensors.py`)
The proportional sensor now exposes an `_endstop_listeners` list and a `register_endstop_listener(callback)` method. On each ADC sample, after updating its own state, it notifies all registered listeners with `(read_time, mapped_value)`. This avoids duplicating ADC setup and ensures the endstop sees the same processed values as the rest of the system.

**Sensor manager** (`mmu_sensor_manager.py`)
The new `SENSOR_EXTRUDER_ENTRY_PROP` (`"proportional"`) sensor is added to the sensor name display list, the endstop registration list, and the rapid-stop stepper list. At startup, when a proportional sensor is configured, the sensor manager registers the virtual endstop on the gear rail via `gear_rail.add_extra_endstop()`.

**Homing logic** (`mmu.py`)
Proportional homing in `_home_to_extruder()` is implemented as a `trace_filament_move()` with `homing_move=1, endstop_name=SENSOR_EXTRUDER_ENTRY_PROP`. **A 2-second pause precedes the homing move to allow the bowden to settle after the fast approach move.**

**Bowden calibration support** (`mmu.py`, `mmu_calibration_manager.py`)
The proportional sensor is added to the allowable bowden-calibration sensor list. A pre-initiation check raises `MmuError` if the sensor is already above the extruder threshold before homing starts. **The calibrated bowden length is reduced by `sync_feedback_buffer_range`** to account for the buffer compression distance, preventing overshoot on subsequent loads.

**Post-load extruder entry validation** (`mmu.py`)
`_validate_extruder_entry_proportional(max_steps=4)` confirms the extruder has gripped the filament after homing:

1. **Phase 1 — synced grip:** runs a short `gear+extruder` synced feed of `0.5 × buffer_range` to seat filament in the extruder gears and establish grip.
2. **Phase 2 — extruder-only probes:** drives the extruder motor only, in steps of `0.5 × buffer_range`, up to `max_steps` (capped at 4). After each step it checks whether the sensor has dropped below `proportional_extruder_threshold`. A drop confirms the extruder is actively pulling filament through and relieving compression. If the sensor never drops, an `MmuError` is raised and filament position is set to `FILAMENT_POS_EXTRUDER_ENTRY`.
3. **Adaptive budgeting:** rather than skipping wholesale when the worst-case 4-step move won't fit, the caller now sizes `max_steps` to whatever the remaining nozzle-load distance allows: `max_steps = min(4, int((length − grip_distance) // step_size))`. Validation is only skipped when the remaining load distance is less than `grip_distance + step_size` (i.e. one synced grip move plus one probe step). This lets short toolhead geometries still get at least one meaningful probe instead of bypassing validation entirely. The check is gated by the same `toolhead_entry_tension_test` parameter as the switch-based compression-loading validation.

**Post-unload extruder validation** (`mmu.py`)
`_validate_extruder_unload_proportional()` confirms the extruder has fully released the filament after an unload. From the post-unload state it performs a short forward extruder-only spin of `buffer_range` and reads the sensor again after a settle dwell. If the sensor shifts toward compression by more than 0.1 the extruder is judged to still be gripping filament. Currently warns (does not error) on failure to mirror the existing encoder-based validation pattern.

**Filament recovery with proportional sensor** (`mmu.py`)
`recover_filament_pos()` now uses the PSF to determine filament state during `MMU_RECOVER` when no toolhead or extruder-entry switch is fitted. Three additions:

1. **Presence inference:** before any state branch, if no sensor in the path detected filament and a PSF is configured, a sensor reading above −0.9 (i.e. not in deep tension) marks `filament_detected = True`. This recovers the case where filament is present but no in-path switch can see it.
2. **"Probably loaded" verification (persisted LOADED state):** when filament position was persisted as `FILAMENT_POS_LOADED` and `looks_loaded` is true (gate confirmed), but no toolhead or entry sensor can verify extruder state, the proportional two-phase grip test is run. If grip is not confirmed, the state is demoted to `FILAMENT_POS_IN_BOWDEN`.
3. **Active grip detection (cascade):** when filament is detected (gate sensor or sensors in path) but position is not persisted as `LOADED`, the proportional two-phase grip test determines whether the extruder has grip. The gate condition was broadened to `(filament_detected or gs)` because `filament_detected` from `check_any_sensors_in_path()` returns `None` when `gate_selected` is unknown, while the gate sensor can still independently confirm filament presence via direct `check_sensor()`. If grip is confirmed, the state is set to `LOADED`.

The two-phase test `check_filament_in_extruder_by_proportional()`:

- **Phase 1:** engages the gear via `filament_drive()` and attempts to centre the sensor using `adjust_filament_tension(max_move=2 × buffer_range)`. If the sensor remains in deep tension (≤ −0.9) after the attempt, the filament is free in the bowden with nothing to compress against → returns `False`.
- **Phase 2:** performs an extruder-only retract of one full `buffer_range`. When the extruder has grip, the retract pushes filament back into the bowden and compresses the buffer — the sensor shifts toward compression (more positive). A shift > 0.1 confirms grip. Filament is re-fed to its original position after a successful detection.

Debug messaging in `check_filament_in_mmu()` was improved to distinguish "no sensors configured" from "no switch sensors or encoder available, will defer to proportional sensor" when a PSF is present.

### Parameters

**`proportional_homing_speed`** (default **15 mm/s**, range 1–40): governs the gear speed during proportional-sensor extruder homing. **A relatively slow speed is recommended to absorb the virtual portion of the endstop (ADC polling cadence and listener dispatch latency).** Not exposed in the default config files but accepts overrides; can also be tuned at runtime via `MMU_TEST_CONFIG PROPORTIONAL_HOMING_SPEED=…`.

**`proportional_extruder_threshold`** (default **0.9**, range 0.1–1.0): the mapped-sensor compression threshold at which the virtual endstop trips. Not exposed in the default config files; runtime-tunable via `MMU_TEST_CONFIG PROPORTIONAL_EXTRUDER_THRESHOLD=…`, which also pushes the new threshold into the registered endstop instance.

**`extruder_homing_endstop: proportional`** — set this to enable proportional-sensor extruder homing.

**`toolhead_entry_tension_test: 1`** — strongly recommended; gates the post-load entry validation.

### Tested and functional

1. Extruder homing using the proportional sync sensor only.
2. Bowden calibration using the PSF only.
3. Post-load extruder entry validation — adaptive probe count, drives the extruder in incremental steps and checks the sensor drops below threshold to confirm grip.
4. Post-unload extruder validation — forward gear spin checks for residual extruder grip.
5. Filament recovery (`MMU_RECOVER`) — two-phase proportional grip test correctly distinguishes loaded vs. bowden-only state, and the −0.9 presence heuristic recovers cases with no in-path switch detection.

### To be validated

1. Further field testing of post-load validation reliability under high-compression scenarios where the PSF may remain above threshold even after successful entry.
2. Recovery reliability under varied buffer spring tensions and bowden path conditions.
3. Behaviour of the adaptive probe budgeting on toolheads where only a single probe step fits (typical of short bowden-to-nozzle geometries with moderate `toolchange_retract`).

#### Bowden tube calibration results
**Extruder endstop vs. proportional sensor acting as virtual endstop. Variance is ~2 mm.**
image

**Run-to-run variance ~1 mm**
image

#### Loading process including load validation
image

#### Deliberately triggered post-load validation error by holding the sensor in a compressed state
image

#### Recovery when filament in bowden but not loaded
image

#### Recovery when filament in bowden and loaded at the extruder
image
image

#### Filament unload verification at the extruder — currently just warns if not released
image

### Reliability considerations

1. Losing the toolhead/extruder sensor means we rely on the buffer spring being stiff enough to overcome natural bowden tube friction.
2. Without a sensor inside the toolhead we are never strictly certain a load has reached the nozzle — we infer it from PSF readings. The post-load validation step mitigates this by confirming extruder grip through incremental movement, but field testing is needed under varied bowden path conditions.
3. Recovery accuracy depends on the buffer having sufficient range for the retract test to produce a measurable sensor shift. Short buffer ranges or very long/loose bowden paths may reduce detection reliability.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -6028,7 +6028,10 @@ class Mmu:
             detected = self.buzz_gear_motor()
             self.log_debug("Filament %s in encoder after buzzing gear motor" % ("detected" if detected else "not detected"))
         if detected is None:
-            self.log_debug("No sensors configured!")
+            if self.sensor_manager.has_sensor(self.SENSOR_PROPORTIONAL):
+                self.log_debug("No switch sensors or encoder available, will defer to proportional sensor")
+            else:
+                self.log_debug("No sensors configured!")
         return detected
 
     # Check for filament at currently selected gate
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -4859,7 +4859,7 @@ class Mmu:
     # threshold, confirming the extruder is actively pulling filament through.
     # Returns: total distance moved (mm) on success
     # Raises: MmuError if the sensor never drops below threshold
-    def _validate_extruder_entry_proportional(self):
+    def _validate_extruder_entry_proportional(self, max_steps=4):
         prop_sensor = self.sensor_manager.sensors.get(self.SENSOR_PROPORTIONAL)
         prop_threshold = self.proportional_extruder_threshold
         buffer_range = self.sync_feedback_manager.sync_feedback_buffer_maxrange
@@ -4874,7 +4874,6 @@ class Mmu:
 
         # Phase 2: Extruder-only steps to confirm grip by checking sensor drops below threshold
         step_size = buffer_range / 2.
-        max_steps = 4
         self.selector.filament_release()
         self.log_info("Proportional post-load validation: driving extruder-only in %.1fmm steps (up to %d) to confirm grip..." % (step_size, max_steps))
 
@@ -4987,7 +4986,9 @@ class Mmu:
                         raise MmuError("Failed to load filament passed the extruder entrance (sync-feedback buffer didn't detect neutral tension)")
 
             # Proportional sensor post-load validation: drive the extruder motor only to pull filament through then
-            # check the proportional sensor has dropped below the threshold, confirming extruder grip
+            # check the proportional sensor has dropped below the threshold, confirming extruder grip.
+            # Adaptive to available travel: always runs a synced grip-establishing feed, then fits as many
+            # extruder-only probe steps as the remaining load distance allows (up to a hard cap).
             elif (
                 self.gate_selected != self.TOOL_GATE_BYPASS
                 and self.toolhead_entry_tension_test
@@ -4996,12 +4997,21 @@ class Mmu:
                 and has_proportional
                 and self.extruder_homing_endstop == self.SENSOR_EXTRUDER_ENTRY_PROP
             ):
-                max_range = self.sync_feedback_manager.sync_feedback_buffer_maxrange * 2.5 # 0.5 synced grip + 4 * 0.5 extruder-only steps
-                if length > max_range:
-                    moved = self._validate_extruder_entry_proportional()
-                    length -= moved
+                buffer_range  = self.sync_feedback_manager.sync_feedback_buffer_maxrange
+                grip_distance = buffer_range * 0.50   # synced feed to establish extruder grip
+                step_size     = buffer_range / 2.     # extruder-only probe step
+                min_range     = grip_distance + step_size   # minimum budget for a meaningful validation (grip + 1 probe)
+                max_cap       = 4                     # hard cap on probe steps to avoid wasted travel when budget is large
+
+                if length >= min_range:
+                    max_steps = min(max_cap, int((length - grip_distance) // step_size))
+                    self.log_debug("Proportional post-load validation: grip %.1fmm + up to %d x %.1fmm probe step(s) (budget %.1fmm of %.1fmm available)"
+                                   % (grip_distance, max_steps, step_size, grip_distance + max_steps * step_size, length))
+                    moved = self._validate_extruder_entry_proportional(max_steps=max_steps)
+                    length = max(length - moved, 0)
                 else:
-                    self.log_info("Proportional post-load validation: skipped - remaining load distance (%.1fmm) is less than required validation move (%.1fmm)" % (length, max_range))
+                    self.log_info("Proportional post-load validation: skipped - remaining load distance (%.1fmm) is less than minimum validation move (%.1fmm = %.1fmm grip + %.1fmm probe)"
+                                  % (length, min_range, grip_distance, step_size))
 
             self.log_debug("Loading last %.1fmm to the nozzle..." % length)
             _,_,measured,delta = self.trace_filament_move("Loading filament to nozzle", length, speed=speed, motor=motor, wait=True)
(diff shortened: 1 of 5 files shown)


JohnBaumb:
I just wanted to chime in and say I'm excited for this PR to get merged and to try it out.

One thing I noticed about my PFS, when first loading from the MMU, the filament may cause the PFS to move as it pushes through, I've observed mine go full limit and then snap back; hopefully that behavior is considered in this PR.

igiannakas:
Yes there’s a settling period of 2 seconds before attempting load to ensure PSF is indeed decompressed, among other checks.

crackmonkeyy:
I copied over all 5 files to my printer (v13.0.464 klipper version and modded ercf v2) then added the extra bits to mmu_hardware.cfg including proportional_extruder_threshold, which i set to 0.2 as i found the spring included in the psf kit very strong and takes alot of force to trigger the stock 0.9 setting and could cause filament grinding some times. your code works very well, i hope its added to mainline soon :) it would be kl if possible to have the virtual endstop show in mainsail ui as the extruder entry endstop switch. thanks

igiannakas:
Glad you’re liking it!!

btw 0.2 threshold is risky. With slight Bowden friction it can trigger early. The PSF included spring ought to have sufficiently low tension for the ERCF to still be able to push through it.

crackmonkeyy:
I have my psf mounted very close to the hotend with a 900mm bowden tube from the ercf. it replaced a tension and compression microswitch setup with a homing seitch for extruder entry so i just put the psf in its place to test.

igiannakas:
Ported to v4

PR #942, opened 2026-04-13
v4 ported - Blobifier beta feature release

igiannakas:
# Blobifier beta

## Overview
This update is a stepping stone towards more a modular blobifier, that is able to operate with both bed mounted brushes, gantry brushes and gantry mounted nozzle leak stops and brush. 

It also brings finer purge process control by exposing some of the hard coded variables aiming to better match the tuning of the purge process to the melt zone length.

In addition, the configuration file has been reorganized into better labeled sections for easier navigation and tuning.

### Stepper blobifier support
Enable the use of the new stepper blobifier, aiming to mitigate servo dying from heat issues. https://github.com/Carrot-collective/Blobifier/tree/main/User%20mods/Stepper%20Blobifier

### User Extension Hooks
Three new hooks allow custom gcode macros to be called during stages of the blobifier sequence:

- `variable_user_pre_blobifier_extension` — Called at the very start, after state save.
- `variable_user_post_purge_extension` — Called after purging but before Z is restored.
- `variable_user_post_blobifier_extension` — Called after Z is restored and state is returned.

### Modular Clean Macro (`variable_clean_macro`)
The nozzle cleaning step is now performed through a configurable macro name (default: `BLOBIFIER_CLEAN`). This allows users to swap in their own cleaning routine — gantry-mounted brush, nozzle stop, or any custom macro — without modifying the main blobifier gcode. The clean macro is now responsible for its own safe-descend checks.

### BLOBIFIER_CLEAN Now Self-Contained
The `BLOBIFIER_CLEAN` macro now handles its own safe-descend check internally rather than relying on the caller. 

### Configurable Mid-Purge Pulse Retract
The periodic retract/de-retract during purging (pulsatile flushing) is now fully tunable:
- `variable_pulse_retract_length` — Length (mm) of each periodic retract (was hardcoded to 2).
- `variable_pulse_retract_speed` — Retract speed in mm/min (was hardcoded to 1800).
- `variable_pulse_deretract_speed` — De-retract speed in mm/min (was hardcoded to 800).
- `variable_purge_retract_interval` — How often (mm of extrusion) the mid-pulse retract fires (was hardcoded to 40).

Larger melt zone toolheads should aim to retract more and faster to allow improved purging efficiency (eg. 5mm-10mm)

### Multi-Step Retract
Retraction during purging is now split into steps (`variable_retract_step`). The final steps ramp down to half speed to ease through the cold-zone transition.

### Configurable Purge Pulse Size
`variable_purge_pulse_size` controls the amount of filament (mm) extruded per Z-raise increment within a blob. Tune this to be approximately equal to your purge zone length. Previously hardcoded to 5.

### Extruder Residual Filament Toggle
`variable_include_extruder_residual` (default: 1) controls whether `extruder_filament_remaining` is added to the purge length calculation. The slicer purge matrix usually already accounts for residual filament in the nozzle. If so, set to 0 to avoid over-purging.

### Purge Length Calculation
`park_vars.retracted_length` is now added **after** the minimum purge length floor is applied, rather than before. This correctly treats the retracted filament as mechanical overhead rather than purge volume, preventing the minimum purge from being inflated by the retraction distance.

### Default Fan Name
`variable_fan_name` now defaults to `"fan"` instead of being commented out

### Sync Feedback Tension Adjustment Relocated
`MMU_SYNC_FEEDBACK ADJUST_TENSION=1` is now called **inside** the blob loop immediately after pressure release, rather than after the entire purge sequence. This centres the sync feedback sensor before the next blob begins, reducing the risk of accidentally triggering FlowGuard.

## Configuration Reorganization
The variable block has been reorganized into the below sections for improved configuration flow:

1. **Hooks & Extensions** — User extension macros and clean macro selection.
2. **Hardware & General** — Toolhead offsets, travel speeds, skew correction, Y offset.
3. **Tray Positions** — Purge location, tray height, servo/stepper tray settings.
4. **Brush / Nozzle Cleaning** — Brush geometry, wipe parameters, acceleration.
5. **Purge Length Tuning** — Min/max purge, modifier, addition, residual toggle.
6. **Purge Sequence Tuning** — Iteration heights, speeds, pulse size (visual guide retained).
7. **Retraction Tuning** — Between-blob retract, step size, periodic pulse retract parameters.
8. **Fan Control** — Part cooling fan speeds and fan name.

## Notes
- All existing variable names are preserved — no breaking changes to variable names.
- Users with custom cleaning routines should set `variable_clean_macro` to their macro name instead of editing `BLOBIFIER` directly.
- This PR contains the stepper blobifier changes.

### DIFF
--- a/config/addons/blobifier_beta.cfg
+++ b/config/addons/blobifier_beta.cfg
@@ -1,0 +1,1248 @@
+# Include servo / stepper hardware definition separately to allow for automatic upgrade
+[include blobifier_hw.cfg]
+
+##########################################################################################
+
+# Sample config to be used in conjunction with Blobifier Purge Tray, Bucket & Nozzle 
+# Scrubber mod. Created by Dendrowen (dendrowen on Discord). The Macro is based on a 
+# version, and Nozzle Scrubber is made by Hernsl (hernsl#8860 on Discord). The device is 
+# designed around a Voron V2.4 300mm, but should work for 250mm and 350mm too. This 
+# version only supports the assembly on the rear-left of the bed. If you decide to change 
+# that, please consider contributing to the project by creating a pull request with the 
+# needed changes.
+
+# IMPORTANT: The rear-left part of your bed becomes unusable by this mod because the 
+#            toolhead needs to lower down to 0. Be sure not to use the left-rear 130x35mm.
+
+# The goals of this combination of devices is to dispose of purged filament during a 
+# multicolored print without the need of a purge block and without the flurries of
+# filament poops consuming your entire 3D printer room. The Blobifier achieves that by
+# purging onto a retractable tray which causes the filament to turn into a tiny blob 
+# rather then a large spiral. This keeps the waste relatively small. The bucket should be
+# able to account for up to 200 filament swaps (for the 300mm V2).
+
+# The Blobifier uses some room at the back-left side of your printer, depending on your
+# printer limits and positions. (usually max_pos.y - toolhead_y and brush_start + 
+# brush_width + toolhead_x). If you do place objects within this region, Blobifier will
+# skip purging automatically. It does this by extending the EXCLUDE_OBJECT_* macro's, so
+# make sure you have exclude objects enabled in your slicer.
+
+# If your using Blobifier in conjunction with the filament cutter on the stealthburner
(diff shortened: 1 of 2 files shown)


igiannakas:
Converting to draft and then will close - code ported in V4 dev branch

PR #944, opened 2026-04-15
Purge_On_Load Function

ModularPrintingSystem:
I added a Purge feature when forcing purge standalone. 

It automatically purges when loading a new filament.

There is a Temperature Logic in the background, which can automatically heat the extruder to a set unloading temperature.

When using purge_on_load, first the extruder gets heated to the set Filament Temperature (from the Gate-Map/Spoolman) of the previous loaded Filament and then extrudes filament as long as the Filament Temperature of the currently loading Filament is not reached. You can set a minimum length, which must always be extruded.

#899 Is also included in here.


igiannakas:
I'm not sure how this differs from using the `purge_macro: "your purge macro"` and `force_purge_standalone: 0`

The system automatically heats up the nozzle already to either the spool man defined value or to the minimum defined in mmu_params. Once a load is performed, it calls the purge macro hook above which handles the purging logic.

ModularPrintingSystem:
> I'm not sure how this differs from using the `purge_macro: "your purge macro"` and `force_purge_standalone: 0`
> 
> The system automatically heats up the nozzle already to either the spool man defined value or to the minimum defined in mmu_params. Once a load is performed, it calls the purge macro hook above which handles the purging logic.

What I want is more specific than the existing purge_macro hook.

With the current logic, Happy Hare heats to a generic safe load temperature, performs the load, and then hands off everything else to purge_macro. That means the purge macro is responsible for both the thermal transition and the purge behavior, but it does not natively know about the previous filament temperature vs. the target temperature of the newly loaded filament.

My requested behavior is:

Before load, heat to the gate-map temperature of the previously loaded filament.
Perform the load at that previous filament temperature.
Then start heating toward the gate-map temperature of the newly loaded filament.
While that target temperature is not yet reached, keep extruding.
If the target temperature is reached before the configured minimum purge length, continue extruding until that minimum length is reached.
Only then continue with any optional post-load macro logic.
So the difference is that this is not just “run a purge macro after load”. It is a temperature-aware transition purge built into the load sequence, using both the previous filament’s configured temperature and the new filament’s configured temperature from the gate map.

igiannakas:
Can't the same be achieved through the use of the currently exposed HH variables to a vast extent? Together with a revised either blobifier, mmu_purge or new purge macro under the add ons folder?

**HH holds the below:**
1. printer.mmu.last_tool
2. printer.mmu.ttg_map : {list} defined gate for each tool
3. printer.mmu.gate_temperature : {list} of filament temperatures, one per gate
4. printer.mmu.gate_material : {list} of material names, one per gate

Also the slicer tool maps as below:
```slicer_tool_map dictionary:
  .tools.<tool_num>
    .name : {string} name of filament
    .color : {string} color of filament in "RRGGBB" or "RRGGBBAA" form (no leading #)
    .material : {string} name of material type
    .temp : {int} filament print temperature
```

**Approach:**
1. Look up the last active tool, 
2. Identify its corresponding gate,
3. Look up the filament temperature for that gate. 
4. Then use that for purging.

**Alternatively:**
1.  Store the last filament print temperature  in the mmu_vars.cfg
2. Reference that new variable in the purge macro (as you are doing now).

**During the purge the purge macro can:**
1. Set the hot-end to the previous current tool temperature from the gate map/new variable 
2. Purge a configurable % from the purge matrix map, bound by a floor value
3. Set the temperature to the new filament temperature 
4. Continue with the last part of the purge. 
5. The macro can have localised variables that govern this process (taper length, min purge temp, min purge volume etc)

This approach keeps python changes limited (a singular new variable set on filament load at worst) and transfers the responsibility to the macro. This makes portability into V4 a lot easier, as the core mmu python code has undergone a lot of change for V4.

The blobifier currently demonstrates how to read these printer variables - for example:
{% set from_tool = printer.mmu.last_tool %}
{% set to_tool = printer.mmu.tool %}
{% set pv = printer.mmu.slicer_tool_map.purge_volumes %}
etc

Just throwing my 2 cents on how I'd approach this challenge :) Hope this helps!

ModularPrintingSystem:
> Can't the same be achieved through the use of the currently exposed HH variables to a vast extent? Together with a revised either blobifier, mmu_purge or new purge macro under the add ons folder?
> 
> **HH holds the below:**
> 
> 1. printer.mmu.last_tool
> 2. printer.mmu.ttg_map : {list} defined gate for each tool
> 3. printer.mmu.gate_temperature : {list} of filament temperatures, one per gate
> 4. printer.mmu.gate_material : {list} of material names, one per gate
> 
> Also the slicer tool maps as below:
> 
> ```
>   .tools.<tool_num>
>     .name : {string} name of filament
>     .color : {string} color of filament in "RRGGBB" or "RRGGBBAA" form (no leading #)
>     .material : {string} name of material type
>     .temp : {int} filament print temperature
> ```
> 
> **Approach:**
> 
> 1. Look up the last active tool,
> 2. Identify its corresponding gate,
> 3. Look up the filament temperature for that gate.
> 4. Then use that for purging.
> 
> **Alternatively:**
> 
> 1. Store the last filament print temperature  in the mmu_vars.cfg
> 2. Reference that new variable in the purge macro (as you are doing now).
> 
> **During the purge the purge macro can:**
> 
> 1. Set the hot-end to the previous current tool temperature from the gate map/new variable
> 2. Purge a configurable % from the purge matrix map, bound by a floor value
> 3. Set the temperature to the new filament temperature
> 4. Continue with the last part of the purge.
> 5. The macro can have localised variables that govern this process (taper length, min purge temp, min purge volume etc)
> 
> This approach keeps python changes limited (a singular new variable set on filament load at worst) and transfers the responsibility to the macro. This makes portability into V4 a lot easier, as the core mmu python code has undergone a lot of change for V4.
> 
> The blobifier currently demonstrates how to read these printer variables - for example: {% set from_tool = printer.mmu.last_tool %} {% set to_tool = printer.mmu.tool %} {% set pv = printer.mmu.slicer_tool_map.purge_volumes %} etc
> 
> Just throwing my 2 cents on how I'd approach this challenge :) Hope this helps!

Thanks, that does help, and I agree that a lot of this can be approximated in a macro.

The main reason I was leaning toward core HH logic instead of only a purge macro is that I wanted this behavior to be part of the load sequence itself, not an implementation detail of one specific purge macro. In my case the desired behavior is:

Heat to the previous filament’s gate-map temperature before loading.
Perform the actual load at that temperature.
Then transition toward the new filament’s gate-map temperature while extruding.
Keep extruding until the new target temperature is reached and a minimum transition length is satisfied.
That is slightly different from “load first, then run a purge macro”, because the pre-load temperature choice and the transition timing are part of the loading workflow, not just the purge routine.

That said, your suggested macro-based approach is valid, especially from a V4 portability and maintenance perspective. If the same behavior can be achieved cleanly through exposed HH state plus a dedicated purge macro/add-on, that is probably the better long-term design than pushing more logic into core Python.

So I think the real question is not whether it is possible in a macro, but whether HH should own that sequencing natively or just expose enough state for macros to implement it. If the exposed variables are sufficient, I’m happy with the macro route too, as long as it can reproduce the full sequence reliably for both manual loads and toolchanges.

moggieuk:

```sh
git remote -v
git fetch upstream
git switch <your-pr-branch>
git rebase upstream/v3
```

```sh
git add <resolved-files>
git rebase --continue
```

```sh
```

```sh
git fetch upstream
```

PR #945, opened 2026-04-16
v4 ported - Gear stepper current adjustment to mitigate tangle

igiannakas:
Adjusts gear current in proportional sync feedback sensor setups when detecting progress towards a tangle scenario.

This can be due to:
1. Momentary additional filament friction due to spool winding
2. End of spool scenario where the filament "hook" or tape are causing the gear stepper to skip

This feature triggers when the proportional sensor is at 30% it bumps the gear stepper current to 100% of the configured value, similarly to when in a load bowden state.

Once sync sensor reading drops below 20%, it is considered safe to restore gate current back to the user configured value (sync_gear_current) as the tangle scenario has been mitigated.

Parameters set conservatively due to very limited downside of operating with a tight prevention margin.

This allows the use of a **low sync current** to prevent stepper overheating/filament deformation at the gate during print while also **providing enough torque to the stepper when needed to overcome spool friction**.

image

**Implementation details:**

1. Three new config parameters: `tangle_prevention_enabled` (default on), `tangle_prevention_threshold` and `tangle_prevention_release` for the trigger and release points respectively. These are hidden by default and feature is always enabled.
2. Feature is only armed after a successful load sequence completes and disarmed before unload sequence begins so it does not interfere with load/unload operations
3. Current changes from the timer callback path are dispatched via reactor.register_callback to avoid lookahead flush issues during printing
4. All parameters are exposed via _MMU_TEST for runtime tuning and visible in MMU_SYNC_FEEDBACK status output

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -420,8 +420,8 @@ class Mmu:
         # Tangle prevention - gear current boost based on proportional sensor tension. Hidden config parameters, feature enabled by default
         # Defaults to 50% tension meaning we are heading towards a tangle. Release at 30% tension to prevent "thrashing"
         self.tangle_prevention_enabled = config.getint('tangle_prevention_enabled', 1, minval=0, maxval=1)
-        self.tangle_prevention_threshold = config.getfloat('tangle_prevention_threshold', 0.5, minval=0.3, maxval=0.9)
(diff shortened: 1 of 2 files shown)


igiannakas:
Ported to v4 dev

PR #946, opened 2026-04-19
v4 merged - Fix cut tip not respecting safe margin post cut

igiannakas:
Currently there is the potential for a cutter pin crash after the cut move is complete due to the do cut move macro not travelling away from the cutter pin by the safe margin to clear the cutter.

This PR addresses this by ensuring the toolhead is distanced from the cutter pin by the user configured safe margins in the mmu cut macro variables.

This PR also fixes a spurious Klipper reported out of range gcode position error.


igiannakas:
Incorporated in v4

PR #948, opened 2026-04-21
v4 ported - Fix MMU_CALIBRATE_PSENSOR sensor saturation grinding

igiannakas:
Adds plateau detection to `_seek_limit()` in `MMU_CALIBRATE_PSENSOR`. Previously, when the sensor hit its physical limit (e.g. ADC saturated at 1.0), the seek loop continued stepping for the entire remaining move distance because the stop condition only checked for a value reversal — which never comes when the sensor is saturated.

Now, if the value changes by less than 0.005 for 2 consecutive steps, the limit is treated as found and the seek stops immediately. This is in addition to the existing value reversal logic.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -2866,6 +2866,9 @@ class Mmu:
             return (avg)
 
         def _seek_limit(msg, steps, step_size, prev_val, ramp, log_label):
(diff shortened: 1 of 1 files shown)


igiannakas:
Ported to v4 dev

PR #949, opened 2026-04-23
v4 ported - Reduce klipper load to mitigate potential TTCs during EKF RD adjustments for proportional sync sensors.

igiannakas:
Proportional sync feedback was occasionally causing Timer-too-close stalls during prints. This PR tackles three sources of load on the reactor thread and adds a dead-band so we don't write to the stepper every controller tick.

### **What changed**
**The gear stepper's rotation_distance is no longer altered directly from the reactor timer body.** Instead the update is registered as a lookahead callback on the printer toolhead, so it lands between moves rather than racing an active step-generation window. 

**Multiple pending rd changes get coalesced to the latest target** with a single in-flight callback.

**The sync controller's JSONL telemetry log now keeps one file handle open per session with a periodic flush**, instead of opening and closing the file on every tick. The handle is closed on reset, disconnect, and gate change.

The extruder-movement callback used to run the full EKF + FlowGuard + rd-update pipeline inline. **That's now split** The timer body captures a move, sensor_state pair and defers the heavy work to its own reactor callback resulting in shorter timer body, but same outcome.

**A 0.3% dead-band was added on top of the scheduler so sub-threshold EKF RD adjustments (hundredths of a percent) doesn't schedule a lookahead callback**. This avoids trivial RD adjustments from being applied. These happen due to sensor chatter/ slow movement during print. Sustained drift accumulates across ticks and crosses the band normally; the controller's envelope clamp still bounds both sides. **This may also mean a deviation up to 0.3% in auto tune, which IMHO is resonable and within general loading/unloading tolerances.**

**Validated against a real print log (5628 baseline schedules → 53 with the dead-band, max single-tick residual 2.5%).**
The dead-band tracker is invalidated at every point where the stepper's rd is set outside the scheduler: unsync, gate swap on type-B MMUs, and controller reset. That keeps the controller's model and the stepper coherent across activation/deactivation.

**Logging around rd changes is split.** Recalculated fires when the controller's internal rd moves; Altered only fires when the dead-band actually commits a stepper-level change, and shows the accumulated delta against the last committed rd rather than the tick-to-tick one, so the numbers in the log match what the gate compared against.

### DIFF
--- a/extras/mmu/mmu_sync_controller.py
+++ b/extras/mmu/mmu_sync_controller.py
@@ -1012,6 +1012,14 @@ class SyncController(object):
         self._last_time_s = None
         self._log_ready = False
 
+        # Cached jsonl log state (used only when cfg.log_sync is True).
+        # We keep the file handle open across ticks and flush periodically to
+        # avoid per-tick open()+close() I/O on the reactor thread, which was
+        # observed to contribute to host stalls / Timer-too-close during print.
+        self._log_fh = None
+        self._log_write_counter = 0
+        self._log_flush_every = 20
+
         self.K = 2.0 / cfg.buffer_range_mm   # mm => normalized delta in x
         self.state = EKFState()
         self.state.c = max(cfg.c_min, min(cfg.c_max, c0))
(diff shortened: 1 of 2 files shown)


igiannakas:
Graph showing RD adjustment filtering on an overnight print. EKF noise suppressed reducing tiny RD adjustment triggering while maintaining responsiveness and convergence to a calibrated RD value.

image

Red dots are the application points of the new RD. Blue is the requested RD from EKF

moggieuk:
finally had time
to look at this.  all makes a lot of sense. Would be great to prepare PR against v4

igiannakas:
Yeap will do it’s in the plan to adapt all these when I’m setup on V4. Expecting to start playing with V4 in a week or so :)

igiannakas:
Ported to v4 dev codebase.

PR #951, opened 2026-04-25
v4 not relevant - Fix wrong-gate unload during cross-gate MMU_EJECT

igiannakas:
`cmd_MMU_EJECT` calls `select_gate(target)` before running the unload, so the unload of the currently loaded tool runs with `gate_selected` already pointing at the eject destination instead of the gate the filament is actually on.

For example: T0 loaded on gate 0, user runs `MMU_EJECT GATE=2`.
- The bowden sensor check evaluates gate 2's sensors (correctly Empty) and trips a spurious `Possible sensor malfunction - mmu_gear_2, ...` warning.
- On a multigear MMU, the bowden retract drives gate 2's empty gear stepper for the full bowden length with the unload reporting success while leaving gate 0 still physically loaded.

This PR aims to fix this by moving `select_gate(gate)` to immediately before `_eject_from_gate`, so the unload runs on the actually loaded gate.

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -7032,12 +7032,17 @@ class Mmu:
                 with self._wrap_suspend_filament_monitoring(): # Don't want runout accidently triggering during filament eject
 
                     current_gate = self.gate_selected
-                    if gate != current_gate:
-                        self.select_gate(gate)
 
                     try:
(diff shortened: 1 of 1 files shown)


igiannakas:
Not relevant for v4. Closing PR.

Issue #954, opened 2026-05-04
[Docs] Document cut tip pushback is a push then a retract


wizhippo:
Depsite the name in cut tip "pushback" is actually 2 moves.  An extrude then a retract. 

This caught me when setting up where my cutter is before the extruder and no sensor (not ideal but what I have to work with), I want to push the filment back past the gears but but then optout of the retract after push..

Just adding a comment where referenced could have made that obvious isntead of having to debug it.

https://github.com/moggieuk/Happy-Hare/blob/a880ac0adccf532cf98a20567c47e10efdee5576/config/base/mmu_cut_tip.cfg#L100
https://github.com/moggieuk/Happy-Hare/blob/a880ac0adccf532cf98a20567c47e10efdee5576/config/base/mmu_macro_vars.cfg#L326

A "variable_pushback_pushonly" option would be handy in my case.

PR #955, opened 2026-05-04
[Feature] Add pushback_push_only variable to control whether to only extrude th…

wizhippo:
Add pushback_push_only variable to control whether to only extrude the pushback length after the cut, or to extrude then retract it. This is useful when the cutter is before the extruder.

Allows extrude only on pushback. See https://github.com/moggieuk/Happy-Hare/issues/954


wizhippo:
more is needed this is not enough

PR #956, opened 2026-05-10
Allow parameters for customization macros

mk-maddin:
Allow the macros purge_macro, mmu_event_macro, print_state_changed_macro & action_changed_macro
 set via parameters.cfg to contain parameters.

e.g. it is possible now to define:
`purge_macro:  MY_PURGE_MACRO CLEAN_NOZZLE=true`

whereas in the past it was not possible to define parameters.


moggieuk (commented):
I'll pull this into v3, but note that parameters are already supported in Happy Hare v4
(currently in early access)

PR #960, opened 2026-05-16
v4 ported - Fix scenario where mmu unload failure on print end would leave the heaters running, ignoring MMU defined timeout.

igiannakas:
If MMU_UNLOAD fails inside the MMU_END macro, the trailing MMU_PRINT_END STATE=complete line still runs (it's already in the gcode queue). That cancels the hotend-off safety timer and transitions straight to COMPLETE — but pause_resume has stopped the SD reader, so the user's PRINT_END never runs either. 

This results in the heaters staying at print temp indefinitely.

This PR aims to address that:
1. cmd_MMU_PRINT_END now early-returns if the MMU is paused, leaving the safety net intact. Normal recovery (MMU_UNLOCK - MMU_RECOVER - RESUME) then lets PRINT_END run from SD as intended.
2. _on_print_end calls TURN_OFF_HEATERS when transitioning into an end-state directly from a paused state — covers CANCEL_PRINT and any other path..

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -3148,10 +3148,19 @@ class Mmu:
     def _on_print_end(self, state="complete"):
         if not self.is_in_endstate():
             self.log_trace("_on_print_end(%s)" % state)
+            # If we're leaving a paused state directly into an end-state, the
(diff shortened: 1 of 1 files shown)


igiannakas:
Ported to v4 dev

PR #963, opened 2026-06-01
Add support for K1-GuppyMod and fix a mispelled gear sensor pin

marcelolcosta:
the gear_sensor_10_pin was duplicated and the gear_sensor_11_pin was missing

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -24,13 +24,18 @@ ARGS=( "$@" )
 
 OS_CREALITY_K1="creality-k1"
 OS_FLYOS_FAST="flyos-fast"
+OS_GUPPY="guppy-k1"
 OS_TYPE=""
+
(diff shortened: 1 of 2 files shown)


marcelolcosta:
pull updated to add support for K1-GuppyMod, an K1 alternative firmware

moggieuk (approved):
I can merge this but note that I'm going to be switching to v4 soon -- v3 will become a legacy branch.

Currently v4 works with more modern versions of klipper so I've no idea if it works with Creality but it would be great to have somebody try...    The installer is also radically different and uses menuconfig.

marcelolcosta:
I will try the v4 as soon as you publish it. but it will be using the GuppyMod. the original fw uses a very old klipper. but, my short-term plan is to replace the electronics of my k1 to a octopus Ez Max + raspberry (or a pad7)..  :)

PR #965, opened 2026-06-05
Improve mmu_cut_tip safety moves to avoid static depressor pin

ningpj:
Current logic wasnt effective and continued to crash into depressor pin.
Uses `safe_margin_x` & `safe_margin_y` to maneuver around static depressor pin with up to 3 moves, bypassing if servo fitted. Post cut and return to park pos, move inboard by `safe_margin_x`


ningpj:
@moggieuk updated mmu_cut_tip macros for v4

moggieuk (commented):
Will merge then delete the duplicate file...

PR #968, opened 2026-06-09
Chore: fix exception when print_time isnt passed

ningpj:
Default to `None`


ningpj:
Implemented directly.

PR #970, opened 2026-06-09
fix(mmu): skip unload of already-loaded gate on print start check (v4)

czechbol:
v4 forward-port of #969.

On print start with the correct filament already loaded, `MMU_CHECK_GATE` ejects
it before verifying the gate, then it gets reloaded.

- `commands/mmu_check_gate.py`: mark an already-loaded gate `AVAILABLE` and skip
  it instead of unloading to re-verify; return early if it was the only gate.
- `config/macros/mmu_software.cfg`: skip `MMU_CHECK_GATE` for a single-tool print
  whose initial tool is already loaded.

Isolated core of #899 (credit @ModularPrintingSystem).


moggieuk (approved):
Thanks and welcome to the project!   I'm going to merge into v4 branch now then will look at v3 later..

PR #972, opened 2026-06-16
Implement NEO intensity parameter V4 feature

ningpj:
Tame neo output from blowing out colors when set to `filament_color` or `slicer_color`. Defaults to 50%.
Currently untested as I don't have LEDs on current HH v4 test rig and wanted to reduce NEO intensity for EMU spool.

Reject if not appropreiate.

### DIFF
--- a/extras/mmu/mmu_filament_movement.py
+++ b/extras/mmu/mmu_filament_movement.py
@@ -1142,9 +1142,10 @@ class MmuFilamentMovement:
 
             moved += step_size
 
-            self.movequeue_dwell(0.2)  # Let ADC sensor catch up
+            self.movequeue_dwell(0.2) # Let ADC sensor catch up
             self.movequeue_wait()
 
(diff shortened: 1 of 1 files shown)


moggieuk:
No problem with this concept but a couple of questions:
1) Probably wants to be applied to "slicer_color" as well
2) Doesn't the color shift with a raw intensity change?  Maybe not particularly important for this application.

ningpj:
It should already be applied to both filament and slicer colour options.  Honestly i expect it will improve visual representation across a broader range of colours as they wont be so blown out. Unfortunately I cant verify it at the moment or tune the intensity default until my EMU testbed is ready to go once PSF homing is pushed.  Sadly i dont have gate neos on my TR

moggieuk:
I see the application to slicer_color now. Sorry I should have expanded the PR a bit.    I can pull this tomorrow...

PR #976, opened 2026-06-20
V4 kconfig Installer updates - generic & EMU specific

ningpj:
* Dont write type-CTD tension/compression pins settings to config when PSF selected
* Add option to edit neopixel pins in Pins & LED sections. Pins for 2nd chain only visible when defined in mmu kconfig e.g `kconfig.emu`
* Uncommented and refined ooze settings in toolhead to enable users to manage related settings. Capped 0->5mm
* Purge refinements - added option for "Slicer Controlled" to make more intuitive
* Tip Forming/Cutting refinements - added option for "Slicer controlled" to make more intuitive
* Refined extruder Homing to enable `filament_compression` selection when compression & PSF configured
* Added conditional HW block for MMU fans - single & multi-mcu setups. Delete pins to disable config block for specific gate

More EMU work needed - fan macros, custom neo effects, additional env sensor HW block. Pushing for feedback and validation

### DIFF
--- a/config/base/mmu.cfg
+++ b/config/base/mmu.cfg
@@ -464,7 +464,7 @@ toolhead_residual_filament: [[PARAM_TOOLHEAD_RESIDUAL_FILAMENT]]		# Reduction in
 # Similarly a value >+5mm also suggests the four settings above are not correct. Also see 'retract' setting in
 # 'mmu_macro_vars.cfg' for final in-print ooze tuning.
 #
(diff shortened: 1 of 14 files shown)


ningpj:
@igiannakas will need to review ebb pin defaults to make sure they are consistent. slb pins mirror my setup and will become the new standard unless cross checked as well ;-)

moggieuk (commented):
Going to merge.  Then will make a couple of minor edits -- for consistency on where types are defined, etc.

PR #983, opened 2026-06-29
V4 Fix proportional sensor unload test

ningpj:
test activated extruder rather than gear to check if the filament had exited the extruder. Changed to "gear" and reduced speed to a sane level (virtual_sensor_homing_speed) for users with fast extruder_load_speed's (e.g. 100mm/s) to prevent filament tip from getting caught in extruder gears


ningpj:
I needed to correct proportional unload logic as it was using the extruder rather than gear stepper and capturing the filament tip again and failing. Also dropped speed to virtual homing speed to prevent driving tip hard enough to get caught again.  Tested and works again.

PR #984, opened 2026-07-05
v4 detune led intensity for two other filament/slicer colour use cases

ningpj:
Missed in orginal change


moggieuk (commented):
Looks good. Thanks.

PR #986, opened 2026-07-06
Proportional post load validation and unload verification code fixes

igiannakas:
Addressing review comments.

Code subject to real world testing (to be performed today).

**Key changes:**

**Load side validation**
Reworked the proportional-sensor validation in the extruder load and unload paths, and cleaned up the two spots that were flagged as unfinished.

The grip check used to wait for the compression sensor to release, which could fail whenever the buffer wasn't compressed to begin with. Unlikely scenario however new approach is more robust. It now watches the sensor shift toward tension as the extruder pulls filament through, which is the real signal that the extruder has grabbed regardless of where the buffer started. 

**Unload side validation**
 It's now built around the potential failure modes. 

First it just looks at where the sensor is resting: if it's sitting in compression, the MMU gear never pulled the filament back, so we flag that and stop. 

Otherwise it does extruder-only retract in steps, watching for the sensor to move toward compression. If it does, the extruder is still holding the filament. If it never does across the whole move, the filament's out. 

We retract rather than push so we can't accidentally shove filament back toward the nozzle, and we step-and-check so we stop the moment we see a grip instead of over-driving the buffer into its end.

**Unload safety margin**
There's one physical dependency we need to take into account. The probe only works if a released tip has been pulled far enough back to cross the tension/compression threshold in the tunables. Also it needs to give a strong enough signal taking into account things like bowden slack. 

In the unload move itself this PR now makes sure the safety margin is at least one buffer range whenever the proportional sensor is doing the validating. If its not a log info is thrown notifying the user that their configured value has been overridden.

This max value is applied so the filament-position accounting still balances, and it only kicks in on the proportional path. 

**The tunables all live as named constants:** 
The shift that counts as a grip, the resting level that means "MMU didn't retract" (0.8), and the bowden slack over-move factor.

### DIFF
--- a/config/base/mmu.cfg
+++ b/config/base/mmu.cfg
@@ -41,8 +41,8 @@
 #
 [mmu_machine]
 
-happy_hare_version: [[HAPPY_HARE_VERSION]]		# Don't mess, used for upgrade detection
-units:              [[MMU_UNITS]]		# Comma separated list of mmu unit config names. E.g unit0, unit1
+happy_hare_version        : [[HAPPY_HARE_VERSION]]		# Don't mess, used for upgrade detection
+units                     : [[MMU_UNITS]]		# Comma separated list of mmu unit config names. E.g unit0, unit1
 
 
 # ┌───────────────────────────────────────────────────────────────────────────────────────────────────────┐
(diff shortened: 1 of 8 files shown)


moggieuk:
How is this going?

igiannakas:
In Athens this week :( fingers crossed all will be well and I’ll be back on Saturday

moggieuk:
Thanks! This is the last KNOWN change.  I didn't want to jump in and thought is better to leave to you to wrap up.

moggieuk:
Any progress?   I have already incorporated your check in mmu_buffer.py, so if you rebase there shouldn't be much left.  Happy to help get this over the finish line.

moggieuk:
Please retarget this PR to the `development` branch. The `v4` branch has been merged into `main`, and ongoing development should target `development`. Thank you.

PR #987, opened 2026-07-06
Fix embedded led-effect shutdown on recent Klipper (set_color now private)

triqi:
Recent Klipper renamed `LEDHelper.set_color` to `_set_color`, already handled for `check_transmit` in this module. The embedded led-effect `_handle_shutdown` still calls `set_color`, so a `klippy:shutdown` raises `AttributeError`. Adds the matching `_set_color`/`set_color` fallback.

Note: not exercised on a real `klippy:shutdown` (its only trigger) — it's a direct mirror of the existing `_check_transmit`/`check_transmit` fallback just above, and the embedded led-effects otherwise render correctly on current Klipper.

### DIFF
--- a/extras/mmu_led_effect.py
+++ b/extras/mmu_led_effect.py
@@ -260,7 +260,10 @@ class ledFrameHandler:
         for effect in self.effects:
             if not effect.runOnShutown:
                 for chain in self.ledChains:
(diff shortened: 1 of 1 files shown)


moggieuk:
Thanks!  I guess that because this is on shutdown no-one cared... but good to have the LED's turned off..

PR #992, opened 2026-07-15
v4: refine last [include..] insertion logic

ningpj:
* Factor inline & newline comments when inserting new include block.
* Make eject button config conditional on pin being defined


ningpj:
Parsing is fiddly!  Wasnt handling inline / following comments properly and was inserting new `[include...]` in between block and comment <sigh>

PR #999, opened 2026-07-27
Initial fixes to get NFC testing running against live EMU for V4

cwiegert:
## RFID2: V3 compatibility, NFC startup, and diagnostics

This PR brings the RFID2 integration into line with the working V3 behavior and adds enough trace logging to diagnose UID-to-Spoolman resolution.

### Changes

- Add per-reader `i2c_mcu` installer configuration.
  - Defaults to the MCU assigned to that gate for per-gate readers.
  - `i2c_mcu` configured immediately above `i2c_address` in the generated NFC reader configuration.
  - Prevents I2C bus lookup failures when readers are attached to gate MCUs rather than the main MMU MCU.

- Preserve V3 Spoolman compatibility.
  - RFID UID lookup, registration, cache building, and auto-created spools now use the Spoolman extra-field key `rfid_tag`.   The RFID2 branch used RFID, but that would render all V3 created spools and UID's obsolete.   With the V3 tag, the users could create their linked database over time, and maintaining resolution to that field will enable a seamless upgrade with their spoolman database.
  - The Happy Hare-visible label remains `RFID`.

- Correct NFC reader initialization timing.
  - Disable the reader’s duplicate `klippy:connect` initialization.   if initialized here, the readers are unstable, and will have ACK responses stuck and unread on the i2c bus.   
  - Initialize all readers once from `mmu:bootup`, scheduled for 2 seconds later.   this is a replica of the stable working model in V3
  - The delay is a non-blocking one-shot reactor timer—there is no `sleep()` or `reactor.pause()`.
  - This matches the V3 startup behavior and gives I2C devices time to settle before PN532 initialization.

- Add level-4/debug NFC lookup tracing.
  - Klipper logs the UID read and its dispatch to Spoolman.
  - Moonraker logs the raw and normalized UID, cache hit/miss, cache-refresh result, and the exact `MMU_GATE_MAP` callback G-code returned to Klipper.
  - This makes “UID not found” diagnosable from `klippy.log` and `moonraker.log`.

- Make installer parsing tolerant of multiline Jinja blocks used by current MMU configuration files.
  - Addresses installer parse failures in multiline template expressions and configuration sections.
  - Solves the installer crash on exist of menuconfig.

### Validation

- `git diff --check` passes.
- Updated Python modules compile successfully with `py_compile`.
- live testing on Klipper with 5 lanes configured on an EMU

## Summary -- 07/27/2026 updates

This change improves the stability of automated NFC jog scanning with per-gate PN532 readers. It prevents I²C communication from starting synchronously while Klipper initializes the NFC virtual endstop and reduces aggressive polling that could leave the PN532 busy between requests.

These changes were live-tested on Klipper with PN532 per-lane readers during automated preload and NFC jog scanning.

### `extras/mmu/unit/nfc/mmu_nfc_reader.py`

Updated `MmuNfcReader.clear_uid()` so it only clears the cached reader state:

```
self.last_uid = None
self.last_target_info = None
self.present = False
```

Removed the hardware release from `clear_uid()`:

```
self.release(reason="nfc_homing_clear")
```

`clear_uid()` is called while the NFC virtual endstop is being prepared for homing. Calling `release()` at this point initiates synchronous I²C communication during `home_start()`. On the tested Klipper/PN532 setup, this caused Klipper to shut down as the automated jog scan started.

PN532 `read_tag()` already releases a successfully selected target after reading it, so issuing another release while clearing the cached UID is unnecessary for the normal scan path.

### `extras/mmu/unit/mmu_nfc_manager.py`

Changed the NFC homing polling values from:

```
NFC_HOMING_POLL_INTERVAL = 0.010
NFC_HOMING_POLL_TIMEOUT  = 0.020
```

to:

```
NFC_HOMING_POLL_INTERVAL = 0.050
NFC_HOMING_POLL_TIMEOUT  = 0.250
```

The previous 20 ms timeout could expire while the PN532 was still processing an `InListPassiveTarget` request. A subsequent polling attempt could then access the reader before it was ready, resulting in:

```
I2C request to addr 36 reports error START_READ_NACK
```

Increasing the timeout allows the PN532 transaction to complete. Increasing the polling interval reduces unnecessary I²C traffic during the NFC homing jog.

## Live Test Results

Tested on a live Klipper installation using PN532 per-lane NFC readers.

The test confirmed:

- Automated NFC jog scanning starts without the previous Klipper shutdown.
- NFC virtual-endstop initialization no longer issues an immediate I²C release.
- The longer timeout prevents the PN532 from being polled again while a previous command is still processing.
- The reduced polling frequency improves I²C stability.
- Per-gate NFC tags continue to be detected during automated jog scanning.
- ### NFC jog scan: home at `gear_homing_speed` instead of `virtual_sensor_homing_speed`
__________________________________________________________________________________
**What changed** in commits 164db75, cae5a59

`_jog_scan()` in `extras/mmu/mmu_filament_movement.py` now passes an explicit
`speed=u.p.gear_homing_speed` to `move_filament()` for the homing move against the
gate's NFC reader.

_home_gate_with_nfc () - added the speed parameter to the method self.move_filament() 

**Why**

The NFC reader-as-endstop (`MmuNfcEndstop`) subclasses `MmuVirtualEndstopSensor`, so
`_resolve_filament_move_speed()` matched its first branch:

    if homing_move != 0 and virtual_endstop:
        speed = speed or u.p.virtual_sensor_homing_speed

That parameter defaults to 15 mm/s (capped at 40) and exists for ADC-polled sensors
such as the proportional sync-feedback buffer — it is declared with
`guard=_guard_has_buffer`, i.e. it is not intended as an NFC tuning knob and is not
even surfaced to users on a bufferless unit. The result was that a jog scan sweeping
up to a full spool rotation (~480 mm) crawled at 15 mm/s.

`gear_homing_speed` (150 mm/s) is the speed already used for every other gear-stepper
homing move, is proven in live testing on this hardware, and is user-tunable. Passing
it explicitly wins over the virtual-endstop default because every branch in the
resolver is `speed = speed or ...`.

**Scope / behaviour notes**

- Acceleration is unchanged: still `min(gear_load_accel, gear_unload_accel)` from the
  `motor == "gear"` branch.
- The per-gate `gate_speed_override` multiplier still applies on top, as with any
  other homing move.
- No new configuration parameter; nothing to change in `mmu_parameters.cfg` or the
  installer.
- The gate preload path is **not** touched. Its compound home against
  `[gate switch, NFC reader]` already ran at `gear_homing_speed` — `MmuCompoundEndstop`
  is not a `MmuVirtualEndstopSensor` subclass, so the virtual-endstop branch never
  applied there. The preload tag-chase / re-home block is left as-is by design.

**Testing note**
 - Tested with live 5 lane reader, all 5 lanes with changing spools through the 5 lanes
 _______________________________________________________________________________________
## Summary

Fixes Creality CFS/K1/K2 RFID tags failing to decode on PN532 readers — a regression
from the V3 NFC implementation. Adds diagnostic tracing to the NFC reader that made the
root cause findable, and brings the PN532/RC522 drivers in line with the PN7160.

Also includes an unrelated preload-jog speed fix for NFC gate homing.

## Root cause

`MmuNfcReader._capture_mifare()` tries three key families in order — Bambu (HKDF Key A),
factory default Key A, then Creality (UID-derived Key B, sector 1). Each attempt calls
`mifare_read_authenticated_blocks()`, which **releases the RF target in a `finally` block**.

The PN532 primitives refuse to transmit without a selected target:

```python
def mifare_authenticate(...):
    if self.current_target is None:
        return False        # returns before _transceive — no RF traffic at all
```

So attempts 2 and 3 reported "auth failed" on every sector **without ever talking to the
tag**. Since Creality is attempt 3, no Creality tag could ever be read on a PN532.

The PN7160 already self-healed via `_ensure_active_target()`; the PN532 had no equivalent.
V3 solved this at the caller (`tag_handler.read_current_tag()` → `_retarget_same_uid()`),
and that logic was lost in the V4 port.

Re-selection is required by the protocol, not just by the release above: a failed MIFARE
authentication also drops the tag out of its RF session.

### Evidence

klippy.log for a Creality tag (`A254BC31`, SAK 0x08, 4-byte UID). Attempt 1 produced five
real `mifare_authenticate` exchanges; attempts 2 and 3 produced **none**:

```
deep read uid=A254BC31 strategy=mifare_classic SAK=0x08 ATQA=0x0004 uid_len=4
mifare_authenticate: block=3 key_A status=0x14      <- real RF exchange
mifare_authenticate: block=7 key_A status=0x14
... (5 total)
MIFARE attempt 'bambu':       usable=False blocks=0 auth_failed_sectors=[0,1,2,3,4]
sector 0 auth failed — skipping                     <- NO authenticate line: no RF at all
... (5 total)
MIFARE attempt 'default_key': usable=False blocks=0 auth_failed_sectors=[0,1,2,3,4]
sector 1 auth failed — skipping                     <- NO authenticate line
MIFARE attempt 'creality':    usable=False blocks=0 auth_failed_sectors=[1]
all MIFARE key attempts failed - no metadata
```

### `extras/mmu/unit/nfc/pn532_driver.py` — the fix

Adds `_ensure_active_target(timeout, expect_uid_bytes)`, structurally mirroring
`PN7160Driver._ensure_active_target()`: return the cached target if one is selected,
otherwise re-select, raising if the tag has gone. Called at the top of
`mifare_read_authenticated_blocks()` inside the existing `try`, so the `finally` release
still runs.

`expect_uid_bytes` additionally rejects a re-selection that picked up a *different* tag,
so blocks can never be attributed to the wrong spool. (The PN7160 has no such check — it
is free here because the UID is already a parameter, and matters because the filament is
being driven during a preload jog.)

### `extras/mmu/unit/nfc/rc522_driver.py` — parity

The RC522 already re-selected via `_ensure_selected_target()`, so it did **not** have the
bug. Adds the same `expect_uid_bytes` guard for consistency. Keeps its existing
`return None` contract rather than raising, because its four call sites test for `None`.

### `extras/mmu/unit/nfc/mmu_nfc_reader.py` — diagnostics

A failed deep read was previously indistinguishable from one that never ran —
`_read_tag_metadata()` had four identical silent `return None` exits, and
`_capture_mifare()` swallowed key-derivation errors with a bare `except Exception`.

- Logs strategy / SAK / ATQA / uid_len on entry (restores a V3 line dropped in the port)
- A distinct message per failure path: unsupported target, empty capture, unrecognised
  format, parse error — plus tag_format/material on success
- Per-attempt MIFARE result: `usable`, block count, `auth_failed_sectors`,
  `read_failed_blocks`. This is what exposed the missing RF traffic.
- Reports the exception when Bambu/Creality key derivation is unavailable instead of
  discarding it (names a missing `pycryptodome` explicitly)
- Passes `tag_parser`'s existing `trace=` callback, which was never wired up. Surfaces
  the Creality AES chain — encrypted blocks 4-6, decrypted payload, ASCII preview —
  at `debug: 3`.

## Impact

- **Creality CFS/K1/K2 tags now decode on PN532.** Confirmed on hardware: a scan
  populated the gate map from tag data and auto-created a new Spoolman spool (the
  previous ID had been deleted). The `| RFID` marker now appears on the gate, which is
  only set via `set_gate_filament_from_tag()` — i.e. proof the deep read produced metadata.
- Also unblocks the factory-default-key path (QIDI Box) as attempt 2, which was equally
  dead on PN532.
- Bambu tags are unaffected — attempt 1 always ran with a live target.
- Behaviour change: the PN532 now raises if the tag leaves the field mid-sequence.
  Handled by the existing `read_tag_data()` handler, which logs
  `deep tag read failed: ...` and returns no metadata — same outcome as before, now with
  a stated reason. Matches PN7160 semantics.
- Deep reads only occur with `nfc_deep_read: 1`

## Verification

tested with reality, Qidi and Bambu tags, after deleting the spools and filament from spoolman

| Case | Before | After |
|---|---|---|
| Creality tag, normal | `None` | `[4, 5, 6]` |
| Tag removed mid-sequence | `None` | raises `no active PN532 target` |
| Different tag swaps in | `None` | raises `re-selected a different tag` |
| `pycryptodome` absent (attempt 1 skipped) | — | `[4, 5, 6]`, 1 re-selection (no wasted call) |

Confirmed on hardware by a live gate scan.

## Follow-ups (not in this PR)

1. **`pycryptodome` is never installed by the V4 installer.** `installer/requirements.txt`
   is `jinja2`/`dill` only, and `make python_deps` targets the *installer's* interpreter,
   not Klipper's venv. Without it, Bambu and Creality key derivation are skipped entirely.
   V3's `install.sh` installed it into the Klipper env as a hard requirement. It was
   present on the test machine (inherited from a V3 install), so it did not mask this bug
   — but a clean V4 install has no crypto. `cbor2` has the same gap.
2. **Deleted Spoolman spools keep resolving.** `get_spool_by_uid()` answers a
   `uid_to_spool_id` cache hit with no revalidation, and `_send_gate_map_update()` treats
   a 404 from `_fetch_spool_info()` the same as a comms timeout — keeping and re-pushing
   the stale entry. Observed as a deleted spool being rewritten into the gate map after
   an eject.
3. **`mifare_read_authenticated_blocks()` continues across sectors after a failed auth**
   without re-selecting. Since a failed auth ends the RF session, later sectors in the
   same call may be doomed regardless of key — so a multi-sector read can under-report.
   Unchanged from V3 and not triggered by the single-sector Creality path.
4. `mmu_gate_map.py` has a comment stating RFID is "always allowed through regardless of
   spoolman origin" directly above a `if not from_spoolman:` guard doing the opposite.
   One of the two is wrong.


extras/mmu/mmu_filament_movement.py
```diff
@@ -371,8 +371,11 @@ def _home_gate_with_nfc(self, gate, gate_es_name, nfc_es_name, nfc_mgr, compound
         else:
             # Gate first: chase the tag forward within the +ve window, then re-home to gate
             if pos > 0:
+                # Home at gear_homing_speed: the reader is a virtual (host-polled)
+                # endstop so it would otherwise default to the much slower
+                # virtual_sensor_homing_speed intended for ADC-backed sensors
                 chase, homed2, _, _ = self.move_filament(
-                    "NFC: Preload scan", pos,
+                    "NFC: Preload scan", pos, speed=u.p.gear_homing_speed,
```
moggieuk, line 378: My speed was intentional because the nfc_poll operates like a virtual sensor.  However if this can work at high speed then great!   What max speeds are you seeing?

extras/mmu/unit/mmu_nfc_manager.py
```diff
@@ -386,13 +388,21 @@ def _handle_mmu_disabled(self):
 
     def _handle_mmu_bootup(self):
         """
-        Delayed event fired once after MMU bootup. Initialize every reader we
-        control and arm shared-reader polling.
+        MMU bootup event. Schedule initialization after the I2C bus settles.
```
moggieuk, line 391: Ok, so this is a delayed - delay.  The "bootup event" occurs 2s after klipper is ready.  So this would be an additional delay.

I'm just curious what the activity is because 2 seconds is already a long time.. ?

extras/mmu/mmu_filament_movement.py
cwiegert: on V3, I am up to 250 mm/s reliably, with an access at 2500mm^3.   current testing with the gate_homing_speed at 150 mm/s.   it's well within the safety zone and I have missed a read

extras/mmu/unit/mmu_nfc_manager.py
cwiegert: will test without the 2nd delay.   it may have been the Klipper:connect that was masking and causing the crash.   I can test this tonight.

cwiegert: no need for the 2nd delay, I will add the depreciation of the second delay on the next PR I push up

PR #1007, opened 2026-08-02
save_variables: keep writes off the ready dispatch and never lose one

### LINKED ISSUE #1006: [Bug]:  Recent klipper save_variables threaded i/o breaking change
### Summary

 klipper commits to save_variables for threaded i/o causes klipper to fault when happy-hare initialises and attempts to write to save_variables (klipper version v13.0-719 and later) 

### Steps to reproduce

* Upgrade to klipper version >= v13.0-719
* Restart klipper and fault occurs

Relevant klipper commits
* https://github.com/Klipper3d/klipper/commit/332fbf2363ff2bee2762e5dffe59f458fc4c03ff
* https://github.com/Klipper3d/klipper/commit/a557234732c867f797b559303c267be0bcb1bec7
* https://github.com/Klipper3d/klipper/commit/cb02056659c5c754787f95fb0c228b4e15dd7027

Startup fault appears to be triggered when Happy Hare issues `SAVE_VARIABLES` from `klippy:ready` callback in `mmu_utils.py`. Before Klipper’s threaded save_variables changes, that was tollerated. `SAVE_VARIABLES` now waits on async executor completion, which internally uses reactor pause. During `klippy:ready`, reactor pause is disabled, so startup flush in `mmu_utils.py` triggers `ReactorError: Internal error - reactor pause disabled`.

### Expected behavior

Klipper and Happy-Hare start up and `SAVE_VARIABLES` updates/writes occur without errors. 

### Actual behavior

Klipper faults when Happy-hare initialises

### MMU Type

Tradrack

### Happy Hare version in '3.1.4-16' format

3.4.x & 4.0.x beta

### Klipper version in '0.13.0-121' format

v13.0-719 and later

### Relevant logs (`mmu.log`)

```shell
Starting heater checks for heater_bed
Starting heater checks for extruder
Unable to save variable
  File "/home/pi/klipper/klippy/extras/save_variables.py", line 59, in cmd_SAVE_VARIABLE
    self.executor.submit(write_out)
  File "/home/pi/klipper/klippy/extras/aio_executor.py", line 55, in submit
    result, exc = completion.wait()
                  ^^^^^^^^^^^^^^^^^
  File "/home/pi/klipper/klippy/reactor.py", line 38, in wait
    self.reactor.pause(waketime)
  File "/home/pi/klipper/klippy/reactor.py", line 240, in pause
    self.verify_can_pause()
  File "/home/pi/klipper/klippy/reactor.py", line 275, in verify_can_pause
    raise ReactorError("Internal error - reactor pause disabled")
reactor.ReactorError: Internal error - reactor pause disabled
Unable to save variable
Unhandled exception during ready callback
  File "/home/pi/klipper/klippy/extras/save_variables.py", line 59, in cmd_SAVE_VARIABLE
    self.executor.submit(write_out)
  File "/home/pi/klipper/klippy/extras/aio_executor.py", line 55, in submit
    result, exc = completion.wait()
                  ^^^^^^^^^^^^^^^^^
  File "/home/pi/klipper/klippy/reactor.py", line 38, in wait
    self.reactor.pause(waketime)
  File "/home/pi/klipper/klippy/reactor.py", line 240, in pause
    self.verify_can_pause()
  File "/home/pi/klipper/klippy/reactor.py", line 275, in verify_can_pause
    raise ReactorError("Internal error - reactor pause disabled")
reactor.ReactorError: Internal error - reactor pause disabled
```

### Checklist

- [x] I searched existing issues
- [x] I can reproduce with the latest version of Klipper and Happy Hare

moggieuk:
## Why

Klipper's `SAVE_VARIABLE` now performs threaded io (`aio_executor`), so it **pauses the calling greenlet twice** — once for the write, once for the re-read that follows it. Two bugs fall out of that, and users are hitting the first one at startup:

```
Unable to save variable
...
reactor.ReactorError: Internal error - reactor pause disabled
Unhandled exception during ready callback
```

**Startup crash.** `handle_ready` enabled writes as its very first statement and then ran ~60 lines — `_load_persisted_state`, every sub-manager's `handle_ready`, selector setup. Any write reached from there issues `SAVE_VARIABLE` inside Klipper's `klippy:ready` dispatch, which runs inside `reactor.assert_no_pause()`. Klipper's bare `except:` swallows the `ReactorError` and re-raises `Unable to save variable`, which is why the real cause only shows up in the traceback.

**Lost writes.** `cmd_SAVE_VARIABLE` snapshots `dict(allVariables)` and renders every `repr()` *before* writing, then discards the live dict and re-reads the file. Anything set during either pause misses the snapshot and is then erased by the re-read — from disk **and** from memory, with no error. The pause doesn't have to be ours: any `SAVE_VARIABLE` opens the window, including the one `config/addons/blobifier.cfg` issues.

## What changed

- **`handle_ready` defers enabling writes** to a reactor callback. Moving the *flag* rather than just the write is the point — it makes it structurally impossible for anything in that long handler to write. Everything stages into the journal and lands in one write once clear of the dispatch.
- **New `SaveVariableJournal`** (`mmu_utils.py`). Every `save_variable`/`delete_variable` is recorded and dropped **only once a post-write re-read proves it reached disk**. Nothing else clears it, so a lost update is simply written again by the next flush.
- **Writes move to a coalescing reactor timer.** No caller has to know whether it holds the gcode mutex, and bursts collapse into one file write. The timer retries rather than blocking when a command holds the mutex — parking a reactor callback mid-flight buys nothing and a state save has no deadline.
- **`klippy:disconnect` flush** closes the resulting set-to-disk window. It deliberately ignores `_can_write_variables`: that flag means "not yet", but at disconnect there is no later write, and a `_wrap_suspendwrite_variables` batch is the largest thing that could be dropped.
- **The 16 direct `allVariables` mutations** now go through the wrapper, so the journal can't be bypassed by a later edit in those paths. Only the `mmu__revision` seed stays direct (it feeds the sanity check immediately below it).

## Reviewer notes

**Old Klipper is unaffected.** Every API used predates 2020 — `register_callback` (2018-06), `run_script` (2018-01), `is_shutdown` (2020-04), `get_mutex` (present in both generations). Pre-`aio_executor` Klipper never pauses, so the journal is inert there: exactly one write attempt, always.

**Where the journal lives is deliberate.** `mmu_utils.py` imports only `copy` and `math`, so it loads standalone and the subtle logic — reconciliation, delete semantics, snapshot isolation — is unit-testable without Klipper. `mmu.py` can't be imported outside Klipper, which is why the core was extracted rather than written inline.

**Deep-copied snapshots.** Callers pass live objects they keep mutating (gate maps, stats). Klipper renders `repr()` *before* pausing, so if the journal held those references it could never distinguish "the write dropped this" from "the caller changed it again". `set` is deliberately excluded — `repr(set())` is `"set()"`, which `ast.literal_eval` rejects, so Klipper cannot round-trip one.

**Retry is gated on a mutation counter**, not on a reconcile mismatch. Retrying on any mismatch would be a write-amplification trap: a value Klipper can't round-trip would mismatch forever and treble every flush.

### One behaviour change

`mmu_selector.py:701` and `mmu_calibration_manager.py:86,106` log "...has been saved" immediately after requesting a write. The disk write now lands a reactor pass later, so the message very briefly precedes the file changing. Left alone — milliseconds, and the disconnect flush guarantees durability.

## Testing

11 new unit tests for the journal, all passing (`python3 -m unittest test.test_mmu_journal`). All touched files compile.

> Note: `python3 -m unittest` reports 9 errors from the pre-existing `test_mmu_server.py`, which needs Moonraker's `components.file_manager`. Those fail identically on `main` and are unrelated.

**Not verified on hardware, and the gap is worth knowing about.** This repo has no Klipper test harness, so only the journal logic is actually exercised. The reactor wiring around it — deferred ready flush, the timer, the mutex retry, the disconnect flush — is compile-checked and reviewed by eye only. The equivalent v4 change was developed against a fake-Klipper harness, which caught three real mistakes that nothing here would have caught. Suggested before merge:

- Boot on modern Klipper: no `reactor pause disabled` / `Unable to save variable` in `klippy.log`, one `mmu__revision` bump at startup
- Boot on pre-`aio_executor` Klipper: behaviour unchanged
- Calibration round-trip + restart — exercises the selector paths whose mutations were rerouted

### DIFF
--- a/extras/mmu/mmu.py
+++ b/extras/mmu/mmu.py
@@ -30,7 +30,7 @@ from .mmu_shared                import *
 from .mmu_logger                import MmuLogger
 from .mmu_selector              import *
 from .mmu_test                  import MmuTest
-from .mmu_utils                 import DebugStepperMovement, PurgeVolCalculator
+from .mmu_utils                 import DebugStepperMovement, PurgeVolCalculator, SaveVariableJournal
 from .mmu_sensor_manager        import MmuSensorManager
 from .mmu_led_manager           import MmuLedManager
 from .mmu_sync_feedback_manager import MmuSyncFeedbackManager
@@ -159,6 +159,10 @@ class Mmu:
     FILAMENT_DRIVE_STATE   = 1
     FILAMENT_HOLD_STATE    = 2
 
+    # save_variable write behavior
+    VARS_FLUSH_ATTEMPTS               = 3     # Write retries when an update raced with the write
+    VARS_FLUSH_RETRY_DELAY            = 0.05  # Re-check interval while a command holds the gcode mutex
+
     # mmu_vars.cfg variables
     VARS_MMU_REVISION                 = "mmu__revision"
     VARS_MMU_CALIB_CLOG_LENGTH        = "mmu_calibration_clog_length"
@@ -265,6 +269,9 @@ class Mmu:
         self._next_gate = None
         self.toolchange_retract = 0.          # Set from mmu_macro_vars
         self._can_write_variables = False
+        self._var_journal = None              # Unpersisted save_variable updates. Setup on connect
+        self._var_flush_timer = None          # Non-None while a coalesced write is outstanding
+        self._writing_variables = False       # Guards against overlapping writes
         self.toolchange_purge_volume = 0.
         self.mmu_logger = None                # Setup on connect
         self._standalone_sync = False         # Used to indicate synced extruder intention whilst out of print
@@ -799,6 +806,10 @@ class Mmu:
         if not self.save_variables or (rd_var is None and revision_var is None):
             raise self.config.error("Calibration settings file (mmu_vars.cfg) not found. Check [save_variables] section in mmu_macro_vars.cfg\nAlso ensure you only have a single [save_variables] section defined in your printer config and it contains the line: mmu__revision = 0. If not, add this line and restart")
 
+        # Must exist before anything calls save_variable(), i.e. before the upgrade block
+        # below and before the managers connect
+        self._var_journal = SaveVariableJournal(self.save_variables)
+
         # Create autotune manager to oversee calibration updates based on available telemetry
         self.calibration_manager = MmuCalibrationManager(self)
 
@@ -807,10 +818,10 @@ class Mmu:
         if bowden_length:
             self.log_debug("Upgrading %s variable" % (self.VARS_MMU_CALIB_BOWDEN_LENGTH))
             bowden_lengths = self._ensure_list_size([round(bowden_length, 1)], self.num_gates)
-            self.save_variables.allVariables.pop(self.VARS_MMU_CALIB_BOWDEN_LENGTH, None)
+            self.delete_variable(self.VARS_MMU_CALIB_BOWDEN_LENGTH)
             # Can't write file now so we let this occur naturally on next write
-            self.save_variables.allVariables[self.VARS_MMU_CALIB_BOWDEN_LENGTHS] = bowden_lengths
-            self.save_variables.allVariables[self.VARS_MMU_CALIB_BOWDEN_HOME] = self.gate_homing_endstop
+            self.save_variable(self.VARS_MMU_CALIB_BOWDEN_LENGTHS, bowden_lengths)
+            self.save_variable(self.VARS_MMU_CALIB_BOWDEN_HOME, self.gate_homing_endstop)
 
         rotation_distance = self.save_variables.allVariables.get(self.VARS_MMU_GEAR_ROTATION_DISTANCE, None)
         if rotation_distance:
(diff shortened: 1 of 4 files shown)


extras/mmu/mmu.py
```diff
@@ -3886,21 +3920,68 @@ def _disable_mmu(self):
...
-            self.gcode.run_script_from_command("SAVE_VARIABLE VARIABLE=%s VALUE=%d" % (self.VARS_MMU_REVISION, mmu_vars_revision))
+        if not self._can_write_variables or self._var_flush_timer is not None:
+            return
+        self._var_flush_timer = self.reactor.register_timer(self._flush_variables, self.reactor.NOW)
+
+    # Retries rather than blocking when a command holds the gcode mutex: run_script would park
+    # this callback mid-flight for the length of that command, and a state save has no deadline.
+    def _flush_variables(self, eventtime):
+        if self._can_write_variables and not self.printer.is_shutdown():
+            if self.gcode.get_mutex().test():
```
nefelim4ag, line 3951: jfyi
It does not make a lot of sense to do so.
During normal printing, it will be blocked 99% of the time, as klippy executes lines from virtual_sdcard.

It is cheaper and more robust to simply do:
```
with self.gcode.get_mutex():
    self._var_journal.apply()
    mutations = self._var_journal.mutations
    mmu_vars_revision = self.save_variables.allVariables.get(self.VARS_MMU_REVISION, 0) + 1
# Blocks on mutex again
self.gcode.run_script("SAVE_VARIABLE VARIABLE=%s VALUE=%d" % (self.VARS_MMU_REVISION, mmu_vars_revision))
```
So, it will block any other SAVE_VARIABLES and ensure that your write happens.

As long as everyone blocks around `gcode.mutex()`, the in-memory state should be consistent.

-Timofey

P.S.
Technically, the `self.gcode.run_script()` already serializes every access.
So, by scheduling several `SAVE_VARIABLES` writes already serialized automatically.
So the journal is not necessary unless you really have to manage tens of variables, updated at the same time.
(where in that specific case, the journal is just an optimization to batch writes).

PR #1028, opened 2026-08-07
Handle rewind when using a shared_lane_endstop (BTT)

cwiegert:
### Feedback from jacksky6

When we finish the NFC read we rewind and re-park. The re-park homes back to `gate_homing_endstop` then pulls back `gate_parking_distance`. Fine with `mmu_exit` — the sensor is at the gate and the filament is sitting on it. Not fine with `mmu_shared_exit` (BT hub): the scan happens up at the gate and park is 100mm behind the hub, so the filament never gets near that sensor.

The hub reads clear, which is exactly what a reverse home is looking for, so `MmuVirtualEndstopSensor.home_start` completes instantly at zero distance and reports success. `_park_from_gate` then applies the parking retraction anyway. Every scan walks the filament back another `gate_parking_distance` until it drops out of the gate.

 When the sensor reads clear, `_park_after_scan` calls `_home_to_gate`, which drives forward up to `gate_homing_max` to reach the hub — into the merged path.

Affects `mmu_shared_exit` and, on no-bowden designs, the extruder entry sensor registered as shared exit. `mmu_exit` and encoder machines are unaffected.

### Fix

Park is `datum + gate_parking_distance`, and the scan already tracks `off` — its offset from the datum. So on a shared endstop we don't home at all, we move `gate_parking_distance - off`. Same net displacement the homing path produces (`-off` back to the datum, then `+parking_distance`), without touching a sensor we can't reach.

`off` is accumulated from **halt** positions, not trigger positions (`MmuDrive.move` takes `halt_pos` and passes `probe_pos=True`), so the deceleration ramps are already in it. It's also the same open-loop trust the `-100mm` parking move already relies on.

`_park_after_scan` now branches on whether the datum is privately owned by the gate:

- **`mmu_exit` / encoder** — unchanged. Homes to the datum, parks off the switch release point, keeps the encoder-overshoot handling. No behavioural change on any machine that works today.
- **shared** — open-loop rewind as above.

Also added: after the rewind, if a per-gate `mmu_exit` exists but isn't the homing endstop (BT `imply`s both), it's read and warns if filament is still detected. Read-only, no motion, catches slip where the excursion is longest.

### Scope

`_park_after_scan` is the only method changed, plus a comment note in `_validate_nfc_gate_jog_scan_window`. No new parameters, constants, config or installer changes.

### Testing

- Harness suite: 911 tests, 0 failures — unchanged from baseline.
- Verified on hardware for the `mmu_exit` path.
- **Shared-datum branch is unverified.** Not on hardware, and no harness coverage either — every shipped NFC profile lands on `gate_homing_endstop=mmu_exit`. BT validation outstanding (@jacksky6).


moggieuk:
This was merged and then reverted on \`v4\` (revert commit 4b8cb98) so it could go back through review from a clean pending state — GitHub doesn't allow un-merging a PR back to open.

Re-opened as #1032 with the same effective change.

PR #1032, opened 2026-08-08
Handle rewind when using a shared_lane_endstop (BTT)

moggieuk:
Re-opens #1028, which was merged and then reverted on `v4` (GitHub doesn't allow un-merging a PR back to pending) so the change can go through review again from a clean pending state.

Same effective change as #1028 - a revert-of-the-revert on top of current `v4`, rather than the original (now stale relative to `v4`) `ww-V4-Development` branch, so it doesn't regress unrelated work that landed on `v4` since #1028 was opened.

### DIFF
--- a/extras/mmu/commands/mmu_nfc_scan.py
+++ b/extras/mmu/commands/mmu_nfc_scan.py
@@ -83,11 +83,15 @@ class MmuNfcScanCommand(BaseCommand):
         filament_pos = mmu.filament_pos
         is_unloaded = filament_pos == FILAMENT_POS_UNLOADED
 
-        # Selecting a different gate with filament loaded requires a crossload-capable MMU
+        # Selecting a different gate with filament loaded requires a crossload-capable MMU -
+        # and, if the gate endstop is a per-unit SHARED resource (mmu_shared_exit, extruder
+        # entry, or the encoder), can_crossload alone isn't enough: it says the selector won't
+        # jam, not that the shared path is clear. Check both.
         can_continue = (
             is_unloaded
             or scan_unit is not active_unit
-            or active_unit.can_crossload
+            or (active_unit.can_crossload
+                and not mmu._shared_gate_path_occupied(scan_unit.p.gate_homing_endstop, gate))
         )
         if not can_continue:
             if self.check_if_loaded(): return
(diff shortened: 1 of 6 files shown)


moggieuk:
Pushed 3 commits on top of the original rewind fix, closing the gap this PR's own validator comment admitted ("Nothing checks it at scan time either - a sweep forward of a shared datum can meet another gate's filament"):

**`Guard shared gate endstops against cross-gate occupancy`**
Adds `_shared_gate_path_occupied()` and wires it into:
- The command-level `can_continue` check in both `MMU_NFC_SCAN` and `MMU_PRELOAD` - `can_crossload` alone used to be a terminal override, but it only says the *selector* won't jam moving between gates, not that a *shared* gate endstop's path (mmu_shared_exit / extruder_entry / encoder) is clear of another gate's filament. Now both are required.
- `_jog_scan` / `_preload_gate` directly, for the two switch-based endstops (mmu_shared_exit, extruder_entry), replacing `_preload_gate`'s old `SENSOR_SHARED_EXIT`-only inline check and widening it to also cover `extruder_entry`.
- The encoder case has no presence sensor to read, so it can only be checked before this operation's own gate selection moves `gate_selected` onto the target - that's why it's guarded at the command level only (documented on the helper).

**`Re-validate gate parking distance when the homing endstop changes live`**
`gate_parking_distance`'s legal sign depends on `gate_homing_endstop`. `MMU_TEST_CONFIG` already validates correctly within one combined command, but two separate commands left a hole: set a positive parking distance while on `mmu_exit` (legal), then switch to a shared endstop later, and the now-unsafe value was never rechecked. `_on_gate_homing_endstop` (and a new `_on_gate_preload_endstop` hook) now re-run that validator against the new endstop.

**`Add test coverage for the shared-gate occupancy guard and rewind branch`**
Covers the previously-untested shared-datum rewind branch itself (repeated `MMU_NFC_SCAN` must settle at one park position, not drift), the new occupancy guard (refusal when a sibling gate occupies `mmu_shared_exit`, and when the encoder's actively-fed filament is on a different gate of a crossload-capable unit), and the `MMU_TEST_CONFIG` validation-gap fix. Each new guard test was confirmed to fail cleanly against the pre-guard code before the guard was added, so they're real regression guards, not vacuous passes.

Full suite: 986 tests, 0 failures.

Not covered: the `extruder_entry` variant of the occupancy guard has no dedicated test - no existing test profile renders `gate_homing_endstop=extruder` (would need a new Kconfig profile with `MMU_HAS_SENSOR_EXTRUDER` + `!BOOL_REQUIRE_BOWDEN_MOVE`). The code path is identical to `mmu_shared_exit` (both take the live-sensor branch), just unexercised by name.

PR #1039, opened 2026-08-10
v4 Improve installer venv detection; add support for special OS use c…

ningpj:
Updates `install.sh` to support platform-specific Klipper/Moonraker defaults (Creality K1, FlyOS-Fast, Guppy). Add a new `-r` flag to explicitly allow root runs, and enforce root/non-root behavior based on OS. Move Python environment activation until after config path resolution & expand venv search paths to include target klipper set with -k. Improve Python 3/venv error handling, and keep PEP 668 fallback to repo venv. Makefile output text updated to clarify dependency check now resolves missing Python deps.

Support for  Repetier-Server <stub> handling not ported.


ningpj:
Still need to update `build.py` to stop/restart services for special OS use cases without systemd (e.g. MIPS)

ningpj:
Added MIPS OS detection and non-systemd service stop/start initd script call outs.  Cant test other than usual breaking change regression tests for standard OS use case

moggieuk:
I did pull locally and fixed the "stashed changes" but I couldn't run:
`./install.sh -z -i -t`
on my mac.  make test, make console worked.

I didn't debug...

ningpj:
revised venv logic and command error handling if not present. 
with PEP 668 "externally managed" handler, also now reverts to system python3 if valid.

moggieuk:
Thanks for this — the platform-detection and venv-search widening are a nice improvement. A few things I think need addressing before merge, all in the new root-policy / ordering logic:

**1. `sudo ./install.sh` on a normal (non-embedded) box now leaves a root-owned venv behind, then errors out.**
The PEP 668 auto-venv-build block (`make installer_venv` / `. "${VENV}/bin/activate"`) now runs *before* the root-policy check. On Debian Bookworm+ (PEP 668 is the default there), a user who runs the installer with `sudo` will have `./venv` created and populated as root, and *then* get told "don't run as root." Re-running without `sudo` afterward fails because they can no longer write to their own repo's venv. Previously this block ran with no root gate at all, so this is a new failure mode on a very common platform. Suggest moving the root check earlier, before anything writes to disk.

**2. `-r` disables the *entire* root policy, not just the "don't run as root" prohibition.**
```sh
if [ -z "${F_ALLOW_ROOT+x}" ]; then
    case "${os_type}" in
    creality-k1 | flyos-fast | guppy-k1)
        if [ "$(id -u)" -ne 0 ]; then
            ...
```
Because the embedded "must be root" check is inside the same `F_ALLOW_ROOT`-gated block, passing `-r` on an embedded platform as a *non-root* user skips that check too — it silently proceeds without root and fails later trying to write into e.g. `/usr/data`. `-r` should only affect the generic "don't run as root by default" branch; the embedded "must be root for this OS" requirement shouldn't be skippable.

**3. The MIPS fallback exit has no escape hatch, and its message is wrong when already root.**
```sh
elif [ "${machine_arch}" = "mips" ]; then
    echo "...Unable to detect embedded OS layout without root access. Re-run ... as root...${C_OFF}" >&2
    exit 1
fi
```
If arch is `mips` but none of the three known layouts match, this exits unconditionally — no `-o` or `-r` override exists for it. And if the user is *already* root but the directory just doesn't exist, the message still tells them to "re-run as root," which isn't actionable. Any existing non-embedded MIPS install would be hard-blocked here where it worked before.

**Minor:**
- `usage()` now has `-t` documented twice (the new line was added but the old one wasn't removed).
- The `-e` help line dropped the `--emu` long-form mention, even though `--emu` is still parsed elsewhere — just a doc regression.
- The description mentions a Makefile text update, but this PR only touches `install.sh` (`changedFiles: 1`) — did that part not make it into the branch?

Not blocking, but worth a look: the venv-search loop sources each candidate's `activate` before validating its Python version, and doesn't explicitly deactivate a rejected one before moving on or falling through to the "system default" check. Probably fine since most `activate` scripts self-clean via `deactivate nondestructive`, but flagging in case it's not universal.

ningpj:
@moggieuk 

Refactored OS detection, handling of special MIPS system layouts and moved root check earlier to prevent venv ownership issue. Also echoes Klipper/Moonraker directory overrides if specified.  "most" python environments other than conda/miniconda do not include explicit `deactivate` command so will need to rely on `activate` unsetting the previous venv before activating the new one.
 
`MakeFile` pohibits sudo command use so special OS install use cases will need to run as root user rather than using sudo wrapper.

moggieuk:
I still have issues introduced with this PR so I incorporated into this PR (which has been merged):
https://github.com/moggieuk/Happy-Hare/pull/1083

Please let me know if anything is missing.

PR #1050, opened 2026-08-11
Spoolman: add printer context to spool list requests

WeLizard:
## Motivation

Happy Hare fetches `GET /v1/spool` while building its spool-location cache. The request currently does not identify the originating printer.

For Spoolman-compatible backends serving multiple Happy Hare installations, this creates a bootstrap problem: the backend has no request context for selecting the appropriate printer-specific spool assignments on the initial fetch.

## Change

Include the current Happy Hare printer hostname in the spool-list request using the optional header:

`X-Printer-Name: <printer_hostname>`

The change is limited to `GET /v1/spool` and does not modify request bodies or Spoolman response formats.

The header provides client context only. Authentication and authorization remain the responsibility of the configured service.

Standard Spoolman ignores the optional header, preserving existing compatibility.

## Testing

- Added a regression test verifying that the spool-list request contains the current printer hostname.
- `python -m unittest test.test_mmu_moonraker` — 55 tests passed.
- `make ALL=1 test` — 993 tests passed, 1 skipped, 4 expected failures.

### DIFF
--- a/components/mmu_server.py
+++ b/components/mmu_server.py
@@ -461,7 +461,9 @@ class MmuServer:
             errors = ""
             assignments = {}
             sids_to_fix = []
-            response = await self.http_client.get(url=f'{self.spoolman.spoolman_url}/v1/spool')
+            response = await self.http_client.get(
+                url=f'{self.spoolman.spoolman_url}/v1/spool',
+                headers={'X-Printer-Name': self.printer_hostname})
             if response.has_error():
                 raise RuntimeError(self.spoolman._get_response_error(response))
             for spool_info in response.json():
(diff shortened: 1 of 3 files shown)


moggieuk:
Thank you for this — great catch on the bootstrap problem, and a clean minimal fix.

While reviewing, we realized the same issue applies to every Spoolman request mmu_server.py makes, not just this one, so we generalized it in #1052: a small wrapper around the shared http_client adds `X-Printer-Name` to any request targeting the configured Spoolman server, without touching the individual call sites.

Your commit (`41f8209`) is cherry-picked into #1052 with your authorship preserved, along with your regression test (adapted to that PR's request-log shape), so this contribution is fully part of the history there. Closing this one in favor of #1052 - thanks again for raising it!

PR #1064, opened 2026-08-13
Skip startup selector homing on unresolved filament state; fix LinearMultiGearSelector Kconfig typo

moggieuk:
## Summary
- `startup_home_selector`'s boot-time skip-check exempted `FILAMENT_POS_UNKNOWN` and short-circuited whenever no gate had ever been selected, so an unresolved filament state at boot (a sensor read failure, a fresh install) could fall through into `PhysicalSelector.home()`'s automatic `unload_sequence()` — which can heat the extruder and run tip-forming — based on nothing more than an ambiguous read. Any state other than a confirmed `FILAMENT_POS_UNLOADED` now skips the autohome with a warning instead.
- Toolhead XYZ homing (`_MMU_AUTO_HOME`) was confirmed to be a separate, unaffected path (ships disabled by default) — not touched here.
- Also fixes an unrelated typo'd Kconfig symbol (`LINEAR_MUTLI_GEAR_SELECTOR`) that made the `LinearMultiGearSelector` installer menu choice silently fall back to `VirtualSelector` instead of ever taking effect.

## Test plan
- [x] New regression test in `test/test_mmu_selector.py` (`test_startup_home_selector_skips_rather_than_guess_on_unresolved_filament_state`) — confirmed it fails on the old code and passes on the fix.
- [x] New regression test in `test/test_mmu_config.py` (`test_choosing_linear_multi_gear_selector_sets_the_matching_param`) — confirmed it fails on the old code and passes on the fix.
- [x] Existing `test_startup_home_selector_homes_first_and_then_reselects_the_gate` still passes unmodified (no regression on the happy path).
- [x] Full suite: `make ALL=1 test` → `OK (skipped=1, expected failures=4)`, matching the documented clean baseline.

### DIFF
--- a/config/base/mmu_hardware.cfg
+++ b/config/base/mmu_hardware.cfg
@@ -552,30 +552,9 @@ sensor_mcu            : [[MCU_NAME]]_buffer
 # │ ╚═╝  ╚═══╝╚═╝      ╚═════╝    ╚═╝  ╚═╝╚══════╝╚═╝  ╚═╝╚═════╝ ╚══════╝╚═╝  ╚═╝╚══════╝ │
 # └────────────────────────────────────────────────────────────────────────────────────────┘
 #
-# NFC READER(S) for spool RFID sensing. Either one common reader that filament is presented to
-# before preload, or one reader per gate. Supported chips and the transports each can use:
-#
-#   reader_type   interface (default first)   notes
-#   rc522         spi                         cs_pin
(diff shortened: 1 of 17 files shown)


moggieuk:
Closing in favor of #1065, which is cherry-picked from a clean `v4` base so it contains only the startup fix and the selector_type fix, without the in-flight NFC "noisy neighbor" commits that had accumulated on `private_v4` in the meantime.

PR #1103, opened 2026-08-18
Adding caching of multiple rfid_tags into klipper

cwiegert:
added asynch push of spool aliases defined by multiple RFID_TAGS from Moonraker to klipper.    the tipple is stored in the gate_map structure, but the tuple is never exposed to the end user in gate_map commands.


moggieuk:
Overlap with PR1104 so merged with it and pushed.
Therefore closing this PR.

PR #1112, opened 2026-08-21
Reduce menuconfig help pane height to 7 and make always visible

ningpj:
Shrinks `_SHOW_HELP_HEIGHT` from 8 to 7 lines and simplifies related visual layout changes by always reserving the full help height regardless of whether help is defined.

Woodworker found the visual resizing and redrawing of the screen distracting when stepping over comments and menu items without help.

This is now fixed in height and have updated spoolman help to fit. Given the bottom 2 key help lines are bold, it shouldnt look too cramped when all 7 help lines are used.

### DIFF
--- a/doc_tools/capture.py
+++ b/doc_tools/capture.py
@@ -517,7 +517,7 @@ class Menuconfig:
         Height that can be given back without crowding the help text against the menu.
 
         The layout is: breadcrumb, title bar, menu window, separator bar, a help pane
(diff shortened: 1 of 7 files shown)


ningpj:
Always show (fixed height)
image
Spoolman help refactor to fit in 7 lines
image

ningpj:
Added `_ALWAYS_SHOW_HELP_WINDOW = True | False` flag to toggle behaviour.

PR #1119, opened 2026-08-25
Type-B: support gate operations without a gate sensor

BuildLabDIY:
### Problem

Type-B MMUs have physically independent gear motors per gate, but not all of them
have a post-gear gate sensor. The QIDI Q2 box is one example: 4 independent
motors, pre-gate sensors only, no gate sensors.

Several places assume a gate sensor is present, which blocks operations that are
safe on this hardware:

1. **`can_crossload`** (`MMU_EJECT`, `MMU_PRELOAD`) requires `has_gate_sensor()`,
   so any operation on a non-selected gate fails with *"Operation not possible.
   Filament is loaded"* while the toolhead is loaded.

2. **`MMU_EJECT`** unloads through the extruder unconditionally. Ejecting a gate
   that is *not* the loaded one heats the nozzle, homes and pulls on the **active**
   filament.

Once those are unblocked, three follow-on issues appear because the state model
assumes a single filament path:

3. **`_eject_from_gate()`** clears `filament_pos` even when a different gate was
   ejected, so the toolhead is reported as unloaded while it is not.

4. **`_load_gate()`** marks the whole path unloaded when a pickup fails. On type-B
   a failed pickup in one gate says nothing about the others.

5. **`MMU_PRELOAD` / pre-gate autoload** drive the gear motor in a retry loop that
   cannot succeed when the hub is already occupied by another gate's filament.

The stock QIDI firmware allows independent load/eject per slot, so this is a
regression for users migrating to Happy Hare.

### Solution

* `multigear` alone satisfies the crossload check (independent motors make it safe
  without a gate sensor).
* The extruder unload in `MMU_EJECT` runs only when the ejected gate is the one
  loaded in the toolhead; otherwise `_eject_from_gate()` moves just that slot's
  gear motor.
* `_eject_from_gate()` keeps `filament_pos` when a different gate is ejected.
* `_load_gate()` does not clear the path state on type-B.
* `MMU_PRELOAD` and pre-gate autoload refuse with a clear message when the hub /
  toolhead is occupied, instead of spinning the motor.

### Compatibility

Every change is gated behind `mmu_machine.multigear` or an existing loaded state.
Type-A and sensor-equipped type-B setups take exactly the same code path as before.

### Testing

Tested on a QIDI Q2 box (type-B, 4 gates, pre-gate sensors only) with Klipper
mainline:

* `MMU_EJECT GATE=n` / `MMU_PRELOAD GATE=n` on a non-selected gate — no longer
  blocked, gear motor of that slot runs.
* `MMU_EJECT` of an idle gate with the toolhead loaded — extruder position
  unchanged, nozzle stayed cold, only the target slot emptied.
* Filament state stays consistent with the sensors across eject / failed preload.
* `MMU_PRELOAD` on an occupied hub — immediate error with a hint, no motor
  movement.


moggieuk:
Hi!  Can you be specific about what operations you want to work without gate sensor?  Is it just "preload" and "eject"? 

The is no way that without a reference homing point near the MMU that subsequent operations can be reliable but it might be ok to allow the preload gate sensor to be NONE.

Note that Happy Hare v4 is about to launch and that formalizes the different "preload gate" and "normal" gate so this might work more naturally with that...

BuildLabDIY:
Thanks — and let me correct my own description first, because I under-specified the hardware.

QIDI Q2 box, type-B, 4 gates:

* `pre_gate_switch_pin_0..3` — one per gate, but sitting **immediately upstream of the gear motors**. They fire when filament is inserted (and trigger autoload); they don't confirm the motor fed anything through.
* **`gate_switch_pin` — present**: a single shared switch downstream of all four gates.
* `mmu_gear_0..3` — **absent**. No per-gate sensor *after* the motor.

So there is a reference homing point near the MMU — the shared hub switch. Bowden calibration homes on it and is stable (760.9 mm, repeatable). What's missing is only the per-gate *post-motor* confirmation, which is what `has_gate_sensor(SENSOR_GEAR_PREFIX, gate)` actually requires. Sorry for making it sound like there was no reference at all.

To your question: `MMU_PRELOAD` and `MMU_EJECT` are the only two operations — both take `GATE=` and both gate on `can_crossload`. `MMU_LOAD` / `MMU_UNLOAD` are untouched. Everything else in the PR is fallout from unblocking those two (eject of an idle gate ran the extruder unload and tugged the loaded filament; `filament_pos` got cleared globally when the gate acted on wasn't the loaded one).

And you're right about v4 — I went and read the branch. `commands/mmu_eject.py` already has `if gate == current_gate and filament_pos != FILAMENT_POS_UNLOADED: self._handle_unload(gcmd)`, which is exactly the condition this PR adds, and `_shared_gate_path_occupied()` / `SHARED_GATE_ENDSTOPS` covers the shared-hub case properly. It's a better model than patching the v3 checks.

So: happy for you to close this if v4 lands soon — no need to carry a v3 patch for it.

Two things, if you have a moment:

1. One case may still be worth a look in v4 — though you'll know better than me whether it's deliberate. `_home_to_gate()` in `mmu_filament_movement.py` ends its failure path with an unconditional `set_gate_status(gate, GATE_EMPTY)` + `set_filament_pos_state(FILAMENT_POS_UNLOADED)` before raising "Couldn't pick up filament at gate".

   For a normal load that's obviously right — the pickup failed, the path is empty. But the same primitive also serves preload (the `label="preload"` call), and on a type-B unit a failed pickup in gate 1 says nothing about filament that's loaded from gate 0. In v3 that's what bit hardest here: the toolhead state got cleared while the toolhead was still loaded, so the *next* operation behaved as if the path were empty.

   If v4's command layer already gates this before reaching the primitive, ignore me. If not, I'm happy to send a focused PR against v4. (`_eject_from_gate()` looks clean there already — it doesn't touch state at all any more.)

2. Rough sense of v4 timing — weeks or months? Asking for my own planning, not to hurry you: I maintain an install kit for this printer and need to decide whether to ship it against v3 now or hold for v4. Either answer is useful, and "no idea yet" is a fine answer too.

Hardware's on the bench either way — glad to test v4 on a type-B box with a shared hub sensor when you want that exercised.

moggieuk:
I've travelling until the weekend but will look closer at the case you raised in v4.  There is an UNKNOWN gate load state that could be employed -- the state means "we couldn't be certain of filament availability but don't assume it isn't there"

v4 timing ... a few days :shushing_face: :shushing_face:

BuildLabDIY:
Sounds good — enjoy the trip, no rush.

`UNKNOWN` sounds like the right fit: "couldn't confirm, don't assume empty" is exactly the situation on a type-B unit when the pickup fails on a gate that isn't the loaded one.

And a few days for v4 is great news. I'll start testing as soon as it's available — the Q2 box (type-B, 4 gates, shared hub sensor, no per-gate post-motor sensors) is on the bench and I reinstall from scratch regularly, so a fresh v4 setup is no trouble. Happy to report back on anything specific you want exercised.

moggieuk:
One thing you can help me with ... I'd like to add the QIDI Q2 box to the installer so defaults are good (v4).  Can you give me a list of configuration changes you made to make it work well.  E.g.
endstops, parking distances, max homing movement, stepper motor currents, etc.

Very easy for me to add with this info.

I also have a PR created for a new "none" preload endstop setting -- this just moves the raw distance with no parking.  This will work for MMU's without individual "exit" sensors even if another gate is loaded.   The logic is also altered so that even with preload_endstop="shared_exit" (the hub), it should still work so long as filament is not currently loaded.

I you want to jump to v4 now you can (you would simply need to move back to the "main" branch when it is launched). To do so:
```
cd ~Happy-Hare
./install.sh -b v4
```
NOTE: v4 requires reconfiguration... easy but necessary.  Read full docs here: https://moggieuk.github.io/Happy-Hare-Doc/

If you migrate now, you could validate preload and the new (preload_gate_endstop=none) PR...

After launch you would simply run: `./install.sh -b main` to switch back.

I may be pushing v4 live this weekend and if so, I can address this Qidi formal support after release as a patch..

BuildLabDIY:
Happy to help — that "none" preload endstop sounds exactly right for this hardware, and the `shared_exit` relaxation covers the rest of it.

Here's the full QIDI Q2 box configuration as it runs here. Everything below is from a working machine (mainline Klipper, stock QIDI box hardware, printing daily).

### Machine

```ini
[mmu_machine]
num_gates: 4
mmu_vendor: QIDI                    # currently 'Other' here only because there's no QIDI entry yet
mmu_version: 1.0
selector_type: VirtualSelector      # type-B, one gear motor per gate, no physical selector
variable_bowden_lengths: 0
variable_rotation_distances: 0
require_bowden_move: 1
filament_always_gripped: 1          # type-B: filament is always gripped by the gate motor
has_bypass: 0
homing_extruder: 1
environment_sensor: temperature_sensor box1_env
filament_heater: heater_generic box1_heater   # the box has a built-in dryer
max_concurrent_heaters: 1
```

### Sensors

```ini
[mmu_sensors]
pre_gate_switch_pin_0..3            # one per gate, immediately UPSTREAM of the gear motors
gate_switch_pin                     # single shared hub switch, downstream of all four gates
sync_feedback_tension_pin           # buffer tension sensor
extruder_switch_pin                 # toolhead sensor (on the THR MCU over UART)
```

No per-gate exit sensors, no encoder (stock). Some users add an encoder on the QIDI optical disc, but it's not standard.

### Key parameters

```ini
gate_homing_endstop: mmu_gate
gate_homing_max: 1500
gate_parking_distance: 80
gate_final_eject_distance: 1500

gear_from_spool_speed: 120
gear_from_buffer_speed: 120
gear_short_move_speed: 80
extruder_load_speed: 16
extruder_unload_speed: 16

bowden_apply_correction: 0

sync_feedback_enabled: 1
sync_feedback_buffer_range: 8
sync_feedback_buffer_maxrange: 12
sync_feedback_speed_multiplier: 5

extruder_homing_endstop: extruder
extruder_force_homing: 0
toolhead_extruder_to_nozzle: 72
toolhead_sensor_to_nozzle: 62
toolhead_entry_to_extruder: 8
```

### Steppers

`rotation_distance: 13.6`, `microsteps: 16`, gear ratio 1:1 (calibrated value, consistent across all four gates on two different boxes).

Note on currents: the QIDI box driver is **not** TMC-UART controlled — motor current is set in hardware on the box board, so there's no `run_current` to configure. Only step/dir/enable pins.

### Calibration values from this machine

* `mmu_gear_rotation_distances`: **13.6** for all gates — consistent across calibration runs and across two different boxes, so this looks like a real hardware constant and would make a good default.
* `mmu_calibration_bowden_lengths`: lands around **750–775 mm**. Important caveat: on this machine that number is only the **hub → extruder** leg (`gate_homing_endstop: mmu_gate`, so the hub switch is the reference point). The **pre-gate → hub** leg inside the box is separate (I haven't measured it precisely — a couple of hundred mm) and is covered by `gate_homing_max` instead — the total path from a gate to the nozzle is close to **2 metres**. Worth keeping in mind when picking defaults, since the two legs are configured in different places.

  I'd treat the bowden number as an order-of-magnitude hint only, not a default: with `autocal_bowden_length: 1` / `autotune_bowden_length: 1` it measures itself and keeps tuning, and the PTFE tube length isn't guaranteed to be identical between units (QIDI could change it between production runs). Mine is slightly shorter than the stock QIDI firmware assumes.

### Things that needed changing from Happy Hare defaults

1. `gate_homing_max` and `gate_final_eject_distance` both at **1500** — this covers the pre-gate → hub leg inside the box. The internal path is long enough that the stock defaults were far too short — 1500 is generous but reliable; I didn't try to find the minimum.
2. `gate_parking_distance: 80`.
3. `filament_always_gripped: 1` plus `VirtualSelector` — without it the gate motors fight each other.
4. `bowden_apply_correction: 0` — no encoder, so correction has nothing to work from.
5. The box has a **built-in filament dryer** (`heater_generic` + environment sensor), which is why `filament_heater` / `environment_sensor` / `max_concurrent_heaters` are set.

Ask if you want anything else — pin assignments, the toolhead cutter setup (the Q2 has a blade in the toolhead, not on the MMU), or the runout/endless-spool behaviour.

I'll move to v4 on a spare SD card and report back.

moggieuk:
Great. I'm  going to add a (beta) support for Qidi to the installer now.

moggieuk:
Actually, what is the mcu used? I'm assume it is fixed.  Maybe I do need pin assignments...

BuildLabDIY:
Yes — fixed hardware, the box has its own MCU: **STM32F401xC**, USB serial, firmware offset **0x8000** (`CONFIG_STM32_FLASH_START_8000` — the stock MKS bootloader takes the first 32K). Stock QIDI firmware enumerates as `usb-Klipper_QIDI_BOX_V2_*`; on mainline it's `usb-Klipper_stm32f401xc_*`.

The toolhead sensor lives on a **separate** MCU (the head board, `THR`, connected over GPIO UART), not on the box.

Full pinout as configured here:

```
Gear steppers (4, one per gate):
  gate 0:  step mmu:PC14   dir mmu:PC13   enable !mmu:PC15
  gate 1:  step mmu:PB9    dir mmu:PB8    enable !mmu:PC0
  gate 2:  step mmu:PC12   dir mmu:PC11   enable !mmu:PD2
  gate 3:  step mmu:PC8    dir mmu:PB2    enable !mmu:PC10

Sensors:
  entry (pre-gate) 0..3:  !mmu:PA0   !mmu:PB3   !mmu:PA13   !mmu:PA7
  shared exit (hub):      ^!mmu:PB1
  buffer tension:         ^mmu:PB0
  toolhead:               !THR:PA1        (on the head MCU, not the box)
```

No UART/diag pins — the drivers aren't TMC-UART, so there's no `run_current` to set.

Also worth noting for the installer: the box has a **built-in filament dryer** — a `heater_generic` plus an environment temperature sensor, both on the box MCU.

BuildLabDIY:
A correction to what I wrote earlier about the box drivers, plus a heads-up from trying the v4 install.

I got stuck installing v4 on this hardware: `mmu_unit.py` requires a TMC section per gear stepper, and the TMC section in turn requires `uart_pin` — but the QIDI box has no UART wired to its drivers. Klipper refuses either way (`Gear stepper TMC configuration not found` if the section is removed, `Option 'uart_pin' must be specified` if it stays). Looks like hardware like this will need a "no TMC / standalone driver" option in the driver choice — right now there are six chip types and they all expect UART.

For reference, stock QIDI firmware drives the box through its own `[box_stepper slotN]` sections — just step/dir/enable plus microsteps, no TMC sections and no `uart_pin` anywhere:

```ini
[box_stepper slot0]
runout_pin: mcu_box1:PA0
step_pin:   mcu_box1:PC14
dir_pin:    mcu_box1:PC13
enable_pin: !mcu_box1:PC15
microsteps: 16
rotation_distance: 13.8
```

And the correction: earlier I said the drivers "are not TMC-UART controlled". What I actually know is only that **Klipper drives them without UART** — I never verified which chips they physically are. The printer's own axes are TMC2209/TMC2240 with UART wired, so the box may well use TMC parts in standalone mode; I shouldn't have stated it as fact.

Happy to open the box and pull the heatsink to read the markings if that's useful for the profile — just say the word.

moggieuk:
I'm creating a profile that doesn't care about TMC. Also making sure Happy Hare doesn't complain and just skips current control logic.

I've never done a config with out TMC before so I might put this in a custom branch so you can try and see how close I get...   Should have something to push within the hour..

moggieuk:
The great thing is I built a pretty good simulator (make console) so can try this out without kipper or real printer...

moggieuk:
Question: So the default setup for preloading should be:
gate_preload_endstop=none
gate_preload_homing_max=100 ?? (or more) This isn't a homing distance if no endstop it is an absolute move distance. So filament would sit approximately that distance in after preload and then when loading for the first time it would home to the "shared_exit" hub sensor..

BuildLabDIY:
`gate_preload_endstop=none` — yes, that's the only option that can work here.

But 100 mm looks too short for this box. With no endstop that becomes a fixed forward move, and your own help text says it's "usually the same as gate homing distance" — here that's `gate_homing_max: 1500`, because the path from a gate through the box and down the bowden to the hub is long (roughly 300 mm just to get out of the box, then the bowden run on top). A 100 mm move would leave the filament sitting somewhere inside the box, nowhere near the hub.

Also worth flagging: `Preload parking distance` is hidden when `endstop=none`, so there's no back-off. In normal operation here the filament homes forward to the `shared_exit` hub sensor and then retracts by `gate_parking_distance: 80` — that retract is what puts it back into the section where the gates are still separate, before the shared path. With a fixed forward move and no parking retract, the filament just stops wherever the move ends, which matters if that lands at or past the hub.

On grip: the entry sensor on this box sits immediately before the gear motor, so the sensor firing does mean the filament is already at the gears — a fixed move is a reasonable assumption there.

moggieuk:
So branch `codex/qidi-box-support` contains tentitive support for Qidi..

To answer your questions:
1) I set `gate_peload_homing_distance: 200` (again this is a fixed movement when no homing endstop). Can change, just let me know.
2) There is no preload "parking distance" with endstop of "none".  There is an INDEPENDENT parking distance for the regular gate unload.  So first time: preload to +200mm.   Then real load... then unload to "shared_exist" park with "gate_parking_distance".  Need to make sure that eject_homing_distance is sufficient to eject from the gate parking position
3) Understand your "entry" sensor position.  Still on preload, marking the gate and "unknown" is still technically correct because Happy Hare is not not sure that a load is really possible...  it renders fine in the UI and console....
(simulator)
```
(\_/)
( *,*)
(")_(") Happy Hare v4.0.0 Ready...
Selecting last gate used (0)...
Gate : | 0 | 1 | 2 | 3 |
Tools: |T0 |T1 |T2 |T3 |
Avail: |■■■|■■■|■■■|■■■|
Selct: |\▼/|~~~~~~~~~~~~ T0
[T0] ■◉■■■■■◯┈┈┈┈┈┈┈┈ [◁ ▷] ┈┈┈┈◯┈┈Ex┈┈┈┈┈┈┈┤Nz UNLOADED 0.0mm

# Happy Hare console  profile=qidi  gates=4
# All 4 gates preloaded, extruder at 220 C.
# Log: /tmp/mmu.log
# /help for meta-commands, MMU_HELP for Happy Hare commands, Ctrl-D to quit.
# /s (or /scroll) scrolls back through the log.
# Clock is live: it runs while you sit here, so timers fire on their own. /live off freezes it.
> mmu_eject
18:24:36 Ejecting gate 0...
18:24:39 The filament in gate 0 can be removed
> mmu_preload
18:24:42 Preloading gate 0 with fixed move...
18:24:43 Fixed preload move complete for gate 0; filament presence was not verified
> mmu_status
18:24:48 MMU: Happy Hare v4.0.0 controlling 1 units:
         ● QIDI-0, a QIDI v1.0 (gates 0-3)
         └ Connected to extruder: extruder
         └ Selector Type: VirtualSelector.

         Print state is INITIALIZED. Tool T0 selected on gate #0
         MMU gear stepper at 100% current and is NOT SYNCED to extruder
         Sync feedback indicates filament in bowden is: TENSION (not currently active)
         Filament positon believed to be: UNLOADED AND PARKED

         For details on TTG and EndlessSpool groups add 'DETAIL=1', for configuration summary add 'SHOWCONFIG=1'

         Gate : | 0 | 1 | 2 | 3 |
         Tools: |T0 |T1 |T2 |T3 |
         Avail: |■?■|■■■|■■■|■■■|
         Selct: |\▼/|~~~~~~~~~~~~ T0
         [T0] ■◉■■■┈┈◉┈┈┈┈┈┈┈┈ [◁ ▷] ┈┈┈┈◯┈┈Ex┈┈┈┈┈┈┈┤Nz UNLOADED 90.0mm
> 
```

moggieuk:
One more question.  Are you sure the Qidi Box only has a "tension" sensor in the buffer?  That's really unusual -- compression is far more useful for detecting homing to the extruder, etc

BuildLabDIY:
Went back to a clean slate for this: removed Happy Hare entirely, cloned fresh, checked out
`codex/qidi-box-support` (now at `c0d52050`), and installed with **your profile untouched** —
the only things I set by hand were the pins, which the profile can't know. Then walked through
every parameter. Answers to your questions first.

## 1. `gate_preload_homing_max` — yes please, and the number is 1500

You asked for the value, so: **1500**.

Measured pre-gate → hub on real hardware, 6 runs: **1126, 1128, 1130, 1130, 1145, 1148 mm**.
The pre-gate switch sits immediately upstream of the gear motor, so that distance is
"insert to hub", and it's the same for all four gates within ~20 mm.

With your 200 the filament stops exactly at `200.0mm` and never reaches the hub:

```
Preloading gate 0 with fixed move...
Fixed preload move complete for gate 0; filament presence was not verified
[T0] ... UNLOADED 200.0mm
```

## 2. `gate_preload_endstop: none` — you're right about the semantics, but blank works better here

Your point that `unknown` is technically correct is fair — with a fixed move Happy Hare genuinely
can't know. But since this unit *does* have a reference switch downstream (the shared hub), letting
preload home to it gives a confirmed result instead:

```
gate_preload_endstop:            (blank → inherits gate_homing_endstop = mmu_shared_exit)
gate_preload_homing_max: 1500

→ Endstop unit0:mmu_shared_exit reached after 1145.2mm
```

Gate then reports available rather than `■?■`, and the first real load starts from a known
position. Both work; blank is just more informative on this hardware. Your call which to ship
as the default.

## 3. "Are you sure the QIDI box only has a tension sensor?" — one switch, and it reads compression

There is exactly **one** switch in the buffer. Physically the box *pushes* filament in and the
switch responds to the resulting pressure — so functionally it is a **compression** sensor.

In v3 I had it declared as `sync_feedback_tension_pin` with inversion: I worked that out by trial
and error at the time, didn't know the subsystem well enough, and once it ran correctly I left it
alone. So "tension" in v3 says more about how I configured it back then than about the hardware.

The consequence matters for v4: **this buffer has no neutral state.** One switch means two
states, and the switch needs roughly 25 mm of slack fed in before it trips — more than
`buffer_maxrange`. So `adjust_filament_tension()` can never reach neutral, and in `_load_extruder`
the compression branch raises `Failed to load filament passed the extruder entrance` and aborts
an otherwise fine load.

What works here is the opposite operation — *charge* the buffer before the synced move to the
nozzle: feed with the gear motor only, homing onto the compression switch. It trips at
**22–25 mm** consistently across 6 runs, the buffer enters the synced move already loaded, and
nothing needs muting. Worth noting `toolhead_post_load_tension_adjust` then has to be 0,
otherwise the stock post-load relax immediately gives that pressure back.

Also: with `sync_to_extruder` the gear `rotation_distance` is already adjusted 13.6 → 15.11
(+11%), so the drive tops the pressure up by itself during the print — no extra reserve feed
needed.

## One more finding: the fast bowden move isn't a homing move

Unrelated to the profile, but it's the biggest difference on this machine.

`_load_bowden` / `_unload_bowden` issue a plain `move_filament(length)`. Here the toolhead sensor
triggers around 760 mm while the calibrated bowden length is 1322 mm, so the gear keeps driving
for another ~500 mm after the sensor has already fired. The follow-up `Homing to extruder` then
reports `reached after 0.0mm` — the switch was already closed, so the endstop never stops
anything and the move is bounded by distance alone.

Same clean install, same hardware, back to back:

| | stock | move made a homing move |
|---|---|---|
| load | **1446.0 mm** | **924.7 mm** |
| unload | — | 923.3 mm (`Bowden unload stopped on unit0:mmu_shared_exit after 763.3mm`) |

Repeatable within 3 mm across 8 runs.

One gotcha if you look at it: in `_unload_bowden` the local `full` flag is always False (the
caller passes an explicit `bowden_move`), so it can't be used to gate such a branch — cost me a
debugging round.

## What worked out of the box

Worth saying explicitly, since the list above is all problems: `CHOICE_GEAR_TMC_NONE`,
VirtualSelector, 4 gates, `rotation_distance 13.6` (matches our calibration exactly),
`gate_homing_max 1500`, `gate_parking_distance -80`, buffer range/maxrange, the speeds, and the
bypass all came up correctly with no intervention. Klipper started clean.

Things I had to supply that the profile simply can't know — pins, MCU, the box's heater/env
sections, the toolhead cutter option — are not complaints.

Happy to send the two code findings as focused PRs against v4 if that's useful. Hardware stays
on the bench either way.

BuildLabDIY:
Two of your earlier points I skipped over — closing them out.

**MCU and pins (your "what is the mcu used?")** — you already got this right in the branch and I
should have said so. `installer/boards/custom/Kconfig.qidi_box` has every pin correct for our
hardware: gear step/dir/enable for all four gates, the four entry sensors, `^!MCU:PB1` for the
shared exit, `^MCU:PB0` for the buffer, and `!THR:PA1` for the extruder sensor. MCU is a fixed
STM32F401xC over serial, and the `serial_exists` fallback chain found it without help. I set
those pins by hand during install before realising they were already the defaults — they matched
exactly.

**Eject clearance with `preload_endstop: none` (your point 2)** — your concern was whether
`gate_final_eject_distance` is enough to clear the gate from the parking position. It is:
`1500` with `gate_parking_distance: -80` ejects cleanly, ~10-13 s per gate, filament fully
released. No change needed there.

**The config list you asked for** — the full set, as it runs here now, beyond what the profile
already provides:

```ini
# mmu_parameters_unit0.cfg
gate_preload_endstop           :          # blank → hub (see previous comment)
gate_preload_homing_max        : 1500     # was 200
gear_homing_speed              : 120      # was 50 — preload 25 s → 10 s
gear_unload_speed              : 120      # was 80
gear_short_move_accel          : 600      # was 400
sync_gear_current              : 70       # was 100
extruder_homing_max            : 1500     # was 100
heater_max_temp                : 70       # box dryer
toolhead_post_load_tension_adjust : 0     # required with the single-switch buffer

# mmu_hardware_unit0.cfg → [mmu_unit unit0]
max_concurrent_heaters         : 1        # default is num_gates (4); the box has one heater

# toolhead geometry (Q2 with the stock toolhead cutter)
toolhead_extruder_to_nozzle    : 76
toolhead_sensor_to_nozzle      : 62
toolhead_entry_to_extruder     : 8
```

Cutter, if anyone enables `MMU_HAS_TOOLHEAD_CUTTER` for a Q2: `pin_loc_xy: 15, 0.5`,
`pin_loc_compressed_xy: -0.5, 0.5`, `retract_length: 15`, `simple_tip_forming: False`,
`pushback_length: 0`. The bed is 256×256 and the cutter pin sits at the front-left corner, so the
generic `14, 250` default drives the head to the wrong end.

One thing the profile references but doesn't generate: `[temperature_sensor box1_env]` and
`[heater_generic box1_heater]` (plus the two dryer thermistors and heater fans). Users have to
supply that section themselves or Klipper won't start. Might be worth a note in the docs, or a
starter snippet alongside the profile.

BuildLabDIY:
Two more things I got wrong in my own notes and should correct here, since both are profile-level.

**Bypass is disabled in the profile, and the Q2 does have one.**

`Kconfig.qidi_box:22` has `select UNSELECT_HAS_BYPASS`, so `PARAM_HAS_BYPASS` stays 0 and the
bypass gate never shows. On this printer the bypass is real and gets used: filament is fed
straight into the toolhead, past the box and past the hub sensors, landing directly on the
extruder gears and the toolhead sensor. It's how I print single-colour jobs without loading a
gate.

Symptom without it: insert filament into the toolhead, the sensor sees it, and the log says

```
> __MMU_SENSOR_INSERT SENSOR=default:extruder
Ignoring extruder insertion because bypass not selected
```

With `show_bypass: 1` in `[mmu_unit unit0]` the gate appears, `MMU_SELECT BYPASS=1` works, and
`bypass_autoload` picks the filament up on the toolhead sensor — heats and loads exactly like v3
did. Worth flipping in the profile, I think. (In v3 this lived as `has_bypass: 0` +
`bypass_autoload: 1`, which reads contradictory but isn't: the 0 refers to a selector-style
bypass channel *inside* the MMU, which a type-B box genuinely doesn't have.)

I said "bypass works out of the box" in an earlier message — that was wrong, I'd already set
`show_bypass` by then and didn't notice.

**Unit naming: the profile says `box1`, the unit is `unit0`.**

The profile defaults `environment_sensor` to `box1_env` and `filament_heater` to `box1_heater`,
but `UNIT_NAME`/`MCU_NAME` fall back to the Makefile default `unit0`, so a stock install ends up
with `[mcu unit0]` / `[mmu_unit unit0]` sitting next to `box1_env` / `box1_heater` in the same
config. Not broken, just inconsistent — and it confused me for a while when I was matching
sections up.

Either naming the unit `box1` in the profile, or renaming those two references to
`unit0_env`/`unit0_heater`, would make it read as one machine. `box1` is the nicer of the two
for a QIDI box, for what my opinion is worth.

BuildLabDIY:
One more, and this one is a real bug rather than a profile value: **FlowGuard status is never
reported on a buffer-only unit.**

`get_status()` has two branches. The encoder branch reports the live flag directly:

```python
# Encoder flowguard only
elif self.mmu_unit.has_encoder():
    status.update({'flowguard': {'active': self.flowguard_active, 'enabled': ...}})
```

The buffer branch instead hands back `self.flowguard_status`, which is only ever refreshed from
the controller output in `_handle_controller_status` (`mmu_sync_feedback.py:625-626`). On a unit
with a buffer and **no encoder**, that dict keeps the initialiser values from line 73 forever:

```
flowguard: {"active": false, "enabled": false, "level": 0.0, ...}
```

...while `MMU_STATUS SHOWCONFIG=1` correctly says `flowguard_enabled = 1` and the log shows
`FlowGuard monitoring activated and Autotune resumed` at print start. So monitoring is running,
but anything reading the API — dashboard, macro, `printer.mmu.flowguard.active` in a conditional —
sees it as off.

Setting the two fields in the buffer branch the same way the encoder branch does fixes it. After
that, idle reports `enabled: true, active: false` and mid-print:

```
flowguard.active : True
flowguard.enabled: True
print_state      : printing
```

I'd assumed for a while this was cosmetic and my own misreading; it isn't — the flag is simply
never populated on this hardware class, so any logic gated on it silently does nothing.

Separately and genuinely informational: `clog_detection` stays False here because that path is
encoder-based and the box has no encoder — protection runs through the sync-feedback buffer
instead, same as v3. Nothing to fix there, mentioning it only because the two fields side by side
made me chase the wrong thing first.

moggieuk:
Ok

There is a lot to unpack here.  I have a different PR (1131) that I’m about to merge into `v4` directly but I don’t consider Qidi setup done.  EDIT: This is now merged.

Here a specific open questions I need help with:

1) If the buffer is sensing **FILAMENT compression** the we need to set it up as such - it could cause all sorts of confusion if it is inverted and set as tension.  The reason is that I assume a 3-state system: compressed, neutral, tension (as experienced by the filament). Inverting and assuming tension is just wrong and I know will have unforeseen side effects.

2) Also on buffer I need to know if it is sprung?  I.e. without filament is it reading untriggered (neutral).  If no sprung so uses will automatically be disabled.

3) Toolhead — (a) does the Q2 have its own toolhead? If so (as I assume it does) what is its official name — I need to add a separate profile for this (with defaults to sent).  (b) Does it ALWAYS have a toothead cutter?

4) Happy Hare supports setting up the temperature sensor and heater and I’d actually prefer to do that because it is consistent with all other designs.  What I need for reference is the Klipper config for both of these so I can set defaults. I will remove the custom questions from menuconfig/Qidi (which was a temp hack anyway).

5) So the Qidi toothed has an extruder sensor (switch?) 8mm for the BMG(or similar) gear?  I seem to recall it using the actual gear separation (Hall effect) to detect buy maybe I’m wrong on that.

6) I’m not understanding the Bowden move explanation to gave me.  The fast part of the Bowden move is not homing — Klipper can’t cope with that (TTC errors), so there is a “move buffer” that stops it in time to do the homing.   Perhaps we can get other things set with v4 before finalizing this.

7) Bypass is now fixed — bypass is always possible with v4.  The “show bypass” is simply whether you want to visualize with a specific MMU (technically also a selection by only on type-A designs)

Since you are now using Happy Hare v4.  Here is a tip for trying configuration WITHOUT messing with your printer config...
```bash
cd ~Happy-Hare
./install.sh -z -t -i
```
The '-t' will run in test mode -- you run through menuconfig, then look in /tmp/mmu_test/.... for the results.  You can manually inspect config until it visually looks correct.

moggieuk:
> One more, and this one is a real bug rather than a profile value: FlowGuard status is never
reported on a buffer-only unit.

This has been addressed in (merged) PR1132

Also, new toolhead created (which also changes macro defaults) with PR1134 (merged)

moggieuk:
BTW Send me a DM on Discord .... might be easier to chat than here..

BuildLabDIY:
I wrote to you on Discord

moggieuk:

```sh
git remote -v
git fetch upstream
git switch <your-pr-branch>
git rebase upstream/v3
```

```sh
git add <resolved-files>
git rebase --continue
```

```sh
```

```sh
git fetch upstream
```

PR #1130, opened 2026-08-28
Clean up MMU Kconfig defaults

moggieuk:
Summary:
- reorganize Box Turtle, EMU, and KMS defaults for consistent readability
- prefer Box Turtle shared-exit gate homing while defaulting preload to per-gate exits when available
- add Box Turtle sync-current, preload parking, and NFC defaults with safe clear directions
- cover Box Turtle split-endstop and NFC-enabled configurations

Tests:
- 49 installer tests
- 35 MMU profile tests


moggieuk:
Replaced with PR1131

Issue #933, opened 2026-04-03
[Bug]:  Erratic Sensor Removal Events Marking Gates as Empty


GcodeGearHead:
### Summary

Filament showing as unavailable after gate has been checked. 

### Steps to reproduce

1) Install Latest version of HH
2) Do not have an Encoder in your setup
3) Comment out sections as per  https://github.com/moggieuk/Happy-Hare/issues/919 
2) Slice model and upload to Mainsail to print. Hit print 

### Expected behavior

All checked lanes should have filament ready for printing 

### Actual behavior

After initial printer checks, gates should be checked and then the filament will be marked as empty 

<img width="591" height="664" alt="Image" src="https://github.com/user-attachments/assets/a1b6710c-26da-4f96-9607-fd7f0bd0c870" />

I have fixed the issue with the following 

```================================================================================
HAPPY HARE BUG FIX: Spurious Sensor Removal Events Marking Gates as Empty
================================================================================

FILE TO MODIFY:
  Happy-Hare/extras/mmu/mmu.py

LOCATION:
  Line 8442-8447 (approximately) in the cmd_MMU_SENSOR_REMOVE function

--------------------------------------------------------------------------------
ORIGINAL CODE (BUGGY):
--------------------------------------------------------------------------------

                if sensor.startswith(self.SENSOR_PRE_GATE_PREFIX) and gate is not None:
                    # Ignore pre-gate runout if endless_spool_eject_gate feature is active and we want filament to be consumed to clear gate
                    if not(self.endless_spool_enabled and self.endless_spool_eject_gate > 0):
                        self._set_gate_status(gate, self.GATE_EMPTY)
                    else:
                        self.log_trace("Ignoring filament removal detected by %s because endless_spool_eject_gate is active" % raw_sensor)

--------------------------------------------------------------------------------
FIXED CODE (COPY THIS):
--------------------------------------------------------------------------------

                if sensor.startswith(self.SENSOR_PRE_GATE_PREFIX) and gate is not None:
                    # Ignore pre-gate runout if endless_spool_eject_gate feature is active and we want filament to be consumed to clear gate
                    if not(self.endless_spool_enabled and self.endless_spool_eject_gate > 0):
                        # Verify current sensor state before marking gate as empty to avoid spurious events
                        current_sensor_state = self.sensor_manager.check_gate_sensor(self.SENSOR_PRE_GATE_PREFIX, gate)
                        if current_sensor_state is False:  # Sensor currently shows no filament
[...]

PR #1136, opened 2026-08-29
Fix for [Bug]: Erratic Sensor Removal Events Marking Gates as Empty #933

kingofmonkeys:
PR for fix defined in bug #933. Credit goes to [GcodeGearHead](https://github.com/GcodeGearHead) but I have tested this fix for a few days now and it appears solid. 

The bug: When a sensor removal event is received (which can be caused by
electrical noise, vibration, or timing glitches), the original code immediately
marks the gate as empty without checking if filament is actually still present.

The fix: Before marking a gate as empty, verify the current sensor state. Only
mark the gate as empty if the sensor CURRENTLY reads no filament. If the sensor
still shows filament present, log and ignore the spurious removal event.

This matches the verification pattern already used in _validate_gate_status()
at line 8838.


moggieuk:
Hi!  I don't think "spurious" events can occur (or can you prove otherwise?)

I sounds like the pregate "entry" sensor is misfiring perhaps because it is overly sensitive.  The test you perform does give the logic a second change but also isn't foolproof against h/w that is "trigger sensitive", it just reduces the issue.

I'm not saying it's not a good idea for redundancy but I'd like to understand the root cause...

kingofmonkeys:
My first thought was the same as your, that it was a hardware issue with the switch which is why I replaced the switch and housing it was in to make sure it wasn't triggering because of clearance or bad switch (actually did several of them because it wasn't just one gate acting up). I think in my case it may be related to my switch wires being close to my stepper wires causing interference and causing the circuit to go high for a brief instant. Ive had the issue with small arduino projects when static causes false triggers even with pull down resistors. 

I spent a bunch of time watching prints to see if the ui ever showed the sensor triggering and I could never catch it. In my case it was always a gate that wasn't being used that would trigger while just having the filament stationary and when the change filament g-code was sent it would say the gate it was switching to was marked empty but the sensor clearly showed it was not. 

I will admit this wasn't my code fix so I took some of this on faith but it did fix my issue and they pointed out this logic was already in other places. 

Open to suggestions on better ways to handle it or how to test.

kingofmonkeys:
Actually looking at the hardware setup for this and this should be pulled up instead of down. Im wondering if the hardware setup is the issue. Im using the siboor ercf v2 kit and it had the switches pre-wired for the pre-gate sensors. Ill need to dig into how they have it setup and see if i need to re-wire this to prevent interference.

moggieuk:
Ah.  Not having a pull up (or down) setting WILL definitely cause the sensor to oscillate but that would be pretty obvious with repeated "MMU_SENSOR" command or refreshing in Mainsail.

Note that Happy Hare v4 DOES include a sanity check like this and reports as an assertion.   I expect v4 to be released any day now...

kingofmonkeys:
I dug into the hardware setup and these switches were NC low with internal pullup configured.  I suspect that because these are NC low that the longer wires running near the stepper wires allows some interference that for an instant makes the MCU think its been triggered.  I'm going to swap them to NC high with pulldown resistors to see if this limits the interference.

moggieuk:
Ok. Note that a microswitch based filament sensor HAS to be built to a high tolerance -- most microswitches have a movement that is large compared to the size of the filament (1.75mm) so false readings are real.  I know in the past it has caused MMU designers to change and tighten the sensor design.  

So it could be interference but could also be that the hardware switch is right on the edge of switching and a change there (even a different microswitch) may help.

That said Happy Hare v4 does include the check similar to this PR...

moggieuk:

```sh
git remote -v
git fetch upstream
git switch <your-pr-branch>
git rebase upstream/v3
```

```sh
git add <resolved-files>
git rebase --continue
```

```sh
```

```sh
git fetch upstream
```

PR #1142, opened 2026-08-31
Fix 3MS extruder sensor alias setup

moggieuk:
## Summary
- fix the no-Bowden sensor alias path updating the gate sensor list instead of the per-gate map
- alias mmu_shared_exit directly to the resolved extruder-entry sensor
- add a bootable 3MS regression profile covering every gate

## Testing
- ./venv/bin/python -m unittest test.test_mmu_profiles
- 39 tests passed


moggieuk:
Closing in favor of a clean replacement targeting main.

PR #1147, opened 2026-08-31
Map gate LEDs by config line so gates can have different LED counts

margaale:
### The problem

The gate → LED mapping is derived three separate times as `num_leds // num_gates`:

- `extras/mmu/mmu_led_manager.py` — `led_indexes()`
- `extras/mmu_led_effect.py` — the per-gate `_led_effect` sections
- `test/console.py` — the console's LED row

That forces every gate on a unit to carry the same number of LEDs.

On a per-gate-MCU machine that isn't a safe assumption. EMU gives each gate its own `[neopixel _unitN_gateI_leds]` chain, and `Kconfig.leds` generates `exit_leds`/`entry_leds` with **one line per gate**. So a lane wired with four exit pixels sitting next to lanes wired with one is an ordinary machine — and the even split quietly pointed most gates at another lane's pixels.

It doesn't always fail loudly either. `exit` of 1,1,1,1,6 is 10 LEDs over 5 gates, which divides, so the config was accepted and then handed every gate 2.

### The fix

Make `[mmu_leds]` the single owner of the mapping.

`VirtualMmuLedChain` now records how many LEDs each config line contributed. Where a per-gate segment was written with exactly one line per gate, that grouping **is** the per-gate split — which is precisely how the shipped per-gate-MCU templates already emit it. Anything else (lines that are strips rather than gates, e.g. the documented 8-gates-over-5-strips wiring in `mmu_hardware.cfg`) keeps the even spread and its divisibility check.

`mmu_led_manager` and `mmu_led_effect` then both consume that mapping instead of recomputing it, so the effect created for a gate and the LEDs the manager stops and paints for it can't drift apart.

### Compatibility

No shipped profile changes behaviour — for every one of them the two paths agree:

- `emu` (SLB): 5 lines × 4 LEDs over 5 gates → 4 each, either way.
- `emu_ebb` defaults: 5 lines × 4 over 5 gates → 4 each, either way.
- `emu_ebb`: one line of 5 over 5 gates → falls to the even spread, 1 each, as before.
- BoxTurtle / ERCF / ViViD: single-line definitions, all even.

The behaviour only differs where the line count equals the gate count *and* the lines are of unequal length — i.e. exactly the configuration that was broken.

`status` and `logo` are not indexed by gate; their existing arithmetic is left untouched so a `GATE=`-scoped `STATUS_EFFECT` behaves exactly as it does today.

### Commits

1. **Map gate LEDs by config line** — the fix itself, plus `TestUnevenLedsPerGate`.
2. **Register the uneven-LED machine as a harness profile** — moves the fixture to `test/hh/profiles.py` as `leds_uneven` and lists it in `CONSOLE_PROFILES`, so `make console --profile leds_uneven` can drive it by hand.
3. **Group the console LED row by the real per-gate counts** — `Console._swatches` had the same even-division assumption. Colours were right, but the gaps between gate groups landed in the wrong places on an uneven layout, so the console made a working fix look broken.

### What has actually been run

Being explicit about this rather than implying more than I can back up:

- ✅ **`test.test_mmu_leds` — 39/39 pass** with commit 1 applied (Python 3.12 container).
- ✅ **The 5 new tests fail without commit 1** (2 failures + 3 errors), so they're a real regression guard rather than a restatement of the new code.
- ⚠️ **Commits 2 and 3 have not been executed**, and neither has the full suite. If a maintainer would rather see green first, say so and I'll hold until it's run end to end.


moggieuk:
This is an interesting enhancement and I agree it is more flexible. 

I know it is draft but it needs to cover the Kconfig changes so that this can be specified reliable in the menuconfig setup without breaking backward compatibility. 

You should also test in the simulator with "make console" ... the leds are visible (of course the simulator logic may need minor tweaking).

margaale:
oh, ok, i was not sure about the Kconfig stuff, maybe you did not want it at all. Im already working on running the tests locally. Sorry, not a python developer here. Just a developer... :-P

margaale:
image

margaale:
image
image
@moggieuk how should i handle the global chain count?

margaale:
i was thinking in adding a flag (Different LED count per gate) that hides the deafult and enables the per gate option. What do you think?

moggieuk:
Thanks. I did a visual inspection of code and it seems to make sense (I'll still need to merge it locally and test a little before integrating because it has the potential of regressions -- particularly thinking about users that now have v4 / menuconfig setups).

I customized menuconfig/Kconfig extensively for Happy Hare and have concepts such as "computed defaults". To keep menuconfig complexity under control it would be good to hide the per-gate chain count under a simple boolean that expands a normally hidden section or "menuconfig" item that similarly hides them.

I'm a little busy today but will try an have a play tomorrow..

BTW The "menuconfig" based configuration is my recommendation to folks but, of course, complex configs can always resort to direct `cfg` file changes.

margaale:
Thanks for the feedback, i will leave the KConfig to you then. Im not happy with any of my ideas. I tried adding a menuconfig but then the chain count is alson in the box pins sections and its hardcoded with a hack for slb in another place....
image

moggieuk:
I will take a look as soon as I have a free time.  Within a week I'm sure.

margaale:
Take your time. No problem. I just reverted my kconfig changes so you can take your own approach

moggieuk:
I managed to pull locally and try.  There is one important regression to resolve before thinking about the Kconfig integration (which I don't think is hard).

Previously, fragment lines were simply concatenated into one logical chain. Gate ownership was then calculated independently as:

total logical LEDs ÷ number of gates

This was introduced so that users can spread power load over multiple pins or because of physical wiring constraints.

This PR overloads the fragment count with new meaning:
```python
if len(groups) == self.num_gates:
    counts = groups
```

So behavior becomes:
- Fragment count differs from gate count: current functionality remains.
- Fragment count equals gate count and fragments are equal-sized: behavior happens to remain equivalent.
- Fragment count equals gate count and fragments are differently sized: each fragment is reinterpreted as belonging to one gate.

Therefore, someone using one fragment per gate purely by coincidence—for distributing current across pins—will have their gate boundaries silently changed if fragment sizes differ. Whole-chain effects still work, but gate-scoped colors and animations address different LEDs.

The PR conflates two independent concepts:
- Physical fragments: electrical distribution across pins/chains.
- Logical gate groups: which logical LEDs belong to each gate.

Those need separate configuration. Fragment layout should remain electrically descriptive, while an explicit gate grouping option—such as per-gate LED counts or boundaries—should opt into uneven mapping. Existing configurations should continue using equal logical division by default.

I think the only way to resolve this is add another (optional) list parameter which is the led count per gate.  The internal default would be say 'None' that would assume even distribution as it does today.

margaale:
Ohh, good catch, i was not aware of that, sorry. Im thinking in another approach where the definition is self contained. If you dont like it I can go back and implement your suggestion.

margaale:
Here's the self-contained approach, pushed.

Unprefixed definitions are back to concatenate-and-divide-evenly, exactly as before, however the fragments happen to fall. Opting into uneven gates is done by prefixing each line with the gate that owns it:

```
exit_leds: 0: neopixel:bt_1 (1-3)
           1: neopixel:bt_1 (4)
           2: neopixel:bt_2 (1)
           3: neopixel:bt_2 (2-4)
```

Why this rather than a parallel per-gate count list:

- The total isn't stated twice, so it can't drift out of sync with the LED definition and needs no cross-validation.
- Two lines can name the same gate, which is how you describe a gate whose LEDs straddle two strips. A count list can only express that when the strips happen to be adjacent in the logical order.

Your separation stays intact — fragments remain electrically descriptive, the prefix is the logical layer on top. Prefix every line or none (half-prefixed has no obvious reading), and every gate must appear, which preserves the existing invariant that a configured segment lights every gate.

It can't collide with existing configs: a chain reference always leads with its section type, so nothing today starts a line with digits followed by a colon.

Tests pin the regression directly — three unprefixed fixtures (equal fragments; **unequal fragments at one-per-gate, 3,1,1,3 over 4 gates**; fragments unrelated to gate count) all assert 2,2,2,2. The same wiring with prefixes asserts 3,1,1,3.

Issue #1149, opened 2026-08-31
[Bug]: v4 fails to install on Creality K1 with UnicodeEncodeError.


herm:
### Summary

Installer fails to show menu

### Steps to reproduce

1) ./install.sh -k /usr/data/klipper -c /usr/data/printer_data/config -z -m /usr/data/moonraker -e
```
Skipping (git) self update
Detected Creality K1 series printer
KLIPPER_HOME=/usr/data/klipper
KLIPPER_CONFIG_HOME=/usr/data/printer_data/config
MOONRAKER_HOME=/usr/data/moonraker
No suitable Klipper Python 3 venv found; continuing with python
No '.mmu_config' found, forcing interactive menu

Using default symbol values (no '.mmu_config')
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/data/Happy-Hare/installer/lib/kconfiglib/menuconfig.py", line 3947, in <module>
    _main()
  File "/usr/data/Happy-Hare/installer/lib/kconfiglib/menuconfig.py", line 671, in _main
    menuconfig(standard_kconfig(__doc__))
  File "/usr/data/Happy-Hare/installer/lib/kconfiglib/menuconfig.py", line 741, in menuconfig
    msg, code = curses.wrapper(_menuconfig) # Happy Hare: Modified to also return exit code
  File "/usr/lib/python3.8/curses/__init__.py", line 105, in wrapper
    return func(stdscr, *args, **kwds)
  File "/usr/data/Happy-Hare/installer/lib/kconfiglib/menuconfig.py", line 842, in _menuconfig
    _draw_main()
  File "/usr/data/Happy-Hare/installer/lib/kconfiglib/menuconfig.py", line 1441, in _draw_main
    _safe_addstr_markup(_menu_win, i - _menu_scroll, 0, _node_str(node), style) # Happy Hare: Added
  File "/usr/data/Happy-Hare/installer/lib/kconfiglib/menuconfig.py", line 3889, in _safe_addstr_markup
    _safe_addstr(win, text[pos:], cur_attr())
  File "/usr/data/Happy-Hare/installer/lib/kconfiglib/menuconfig.py", line 3801, in _safe_addstr
    win.addnstr(y, x, s, maxlen, attr)
[...]

PR #1154, opened 2026-09-01
Handle ASCII-only menuconfig terminals

moggieuk:
## Summary
- recover when curses cannot encode Unicode on minimal ASCII-only systems such as Creality K1 firmware
- transliterate menu glyphs as: line to hyphen, right arrow to greater-than, and degree to caret
- replace any other character unsupported by the active curses encoding with a question mark
- add regression coverage for styled writes through an ASCII curses window

## Testing
- PYTHONPATH=installer/lib/kconfiglib venv/bin/python -m unittest test.installer.test_menuconfig
- PYTHONPATH=installer/lib/kconfiglib venv/bin/python -m unittest discover -s test/installer (79 tests)

### DIFF
--- a/installer/lib/kconfiglib/menuconfig.py
+++ b/installer/lib/kconfiglib/menuconfig.py
@@ -247,6 +247,14 @@ _SUBMENU_INDENT = 4
 # Number of steps for Page Up/Down to jump
 _PG_JUMP = 6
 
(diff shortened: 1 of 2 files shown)


moggieuk:
This should address Issue: https://github.com/moggieuk/Happy-Hare/issues/1149

PR #1171, opened 2026-09-03
Make install.sh cwd-agnostic and improve self-update handling

burkfers:
Split into two commits:

**Make install.sh cwd-agnostic**

Before: running Happy-Hare/install.sh from `~` broke the self-update: all its git calls ran in `~`, so the check died and reported the misleading "Error updating from github / old version of git", and a `-b <branch>` switch could never succeed. Same call with menuconfig also aborted with "Config … has not been saved": `install.sh` looked  at the config in `~`, while make read the one in the project dir.

Fixes: `install.sh` `cd`s into the project before running anything (not when sourced), `self_update.sh` anchors to its own location, both re-execs use the absolute script path, and a relative `KCONFIG_CONFIG` anchors to the checkout.

**self_update: use the branch's tracking remote for the update check**

Before: the update check diffed against `origin/<branch-name>`. On a local-only branch that ref doesn't exist, git diff was fatal, the error was misread as "new version found", and the pull then always failed.

Fixes: use the branch's tracking remote (@{u}, falling back to a same-named origin/ branch), verify the ref first, treat a branch with no remote counterpart as up to date. Tests now run a script copy in a throwaway checkout (the real one re-anchors to its own path and would act on your repo), cover both new cases, and isolate scratch-repo git from the runner's gitconfig so the suite stays non-interactive anywhere.

Incidentally fixes the issue that depending on the user's git configuration, the v3 checkout test could spawn an editor asking for a commit message.

:robot: Disclosure: Heavily LLM-assisted. Changes have been reviewed and tested to the best of my human abilitities.
Changes to existing code are kept minimally invasive, the only real wordy bit is the new tests.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -37,6 +37,7 @@ usage() {
     echo "${C_OFF}"
     echo "${C_INFO}(no flags for safe re-install / upgrade)${C_OFF}"
     echo "${C_INFO}[config_file]${C_OFF} is optional, if not specified the default config filename (.mmu_config) will be used."
+    echo "  Relative paths are interpreted relative to the Happy-Hare checkout directory."
     echo "  -i for interactive install (open menuconfig)"
     echo "  -u, -d for uninstall"
     echo "  -b <branch> to switch to specified feature branch (sticky)"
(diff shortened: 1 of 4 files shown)


moggieuk:
I'm not convinced the self_update.sh is correct...   but combined this is a hugely risky change that while may have some merit would need thorough manual testing -- including:
1) v3 install to v4 upgrade
2) v3 install to v3 continuation
3) Running from Happy-Hare directory
4) Running from non-HappyHare directory
3) kiauh (with Happy Hare PR) testing for install / update / remove on:
 a) v3 install
 b) v4 install
 c) no install
 
 
 A far easier and much safer "fix" is simply to disallow running with a cwd anywhere but the project root.  See [PR 1173](https://github.com/moggieuk/Happy-Hare/pull/1173)

burkfers:
I have extracted the fix to the most impactful bug (lack of isolation from the developer's git configuration making tests spawn an editor, which is annoying interactively and fatal noninteractively) in #1179. You have worked around the call-from-outside-the-project-dir problem, and fixing the self updater failing on local or fork-tracking branches does not survive a risk/benefit assessment.

PR #1175, opened 2026-09-04
Speed up Kconfig (and thus backed test suite)

moggieuk:
## Summary

- replace deterministic Kconfig shell pipelines with Python preprocessor functions
- retain shell execution only for serial and CAN hardware discovery
- reuse parsed Kconfig trees by environment during harness rendering, resetting all symbol and choice values between profiles
- add regression coverage for preprocessor equivalence and cache isolation
- refresh test documentation and per-file benchmark estimates

## Performance

A warm full-suite run on an Apple M1 Pro improved from 348.5s to 141.1s, a 59.5% reduction.

| Test file | Before | After |
|---|---:|---:|
| `test_mmu_config` | 48.4s | 19.7s |
| `test_mmu_profiles` | 45.0s | 7.7s |
| `test_mmu_selector` | 20.1s | 3.4s |
| `test_mmu_console` | 48.6s | 31.3s |

## Verification

- `make ALL=1 test` — 1,339 tests passed in 141.1s
- installer suite — 121 tests passed
- `make KCONFIG_CONFIG=/tmp/hh-kconfig-performance-check olddefconfig`
- `make lint`
- `git diff --check`


moggieuk:
Additional performance measurements for the installer path:

The Kconfig changes also materially reduce the machine-processing time in `./install.sh -i`. I compared the pre-change tree (`91c4f61e`) with this PR on the same Apple M1 Pro / Python 3.13.7 environment.

| Kconfig stage | Before | After | Improvement |
|---|---:|---:|---:|
| Fresh Kconfig parse (representative of menuconfig startup; median of 20 runs) | 8.88s | 0.71s | 92.0% |
| Installer pre-parse/pickle | 9.34s | 0.81s | 91.3% |
| Combined single-unit Kconfig processing | 18.22s | 1.52s | 91.6% |

For a typical single-unit interactive install, that is approximately **16.7 seconds saved** before accounting for the time the user spends in menuconfig.

This measures only Kconfig-related processing. Update checks, user interaction, config rendering/copying, and service restarts are unaffected, so it should not be interpreted as a 91.6% reduction in the complete wall-clock duration of an interactive installation. Multi-unit installs parse the top-level config and each unit config; their exact wall-time savings depend partly on which build prerequisites Make can overlap.

These absolute figures are from the Mac benchmark host. Raspberry Pi timings have not been measured, though avoiding the old shell-process-heavy preprocessing should also benefit lower-powered systems.

PR #1183, opened 2026-09-05
Fixes mixed environment sensor configuration and adds a debounce/arming of the blobifier bucket sensor switch

cwiegert:
## Summary

This PR improves Blobifier bucket-count reliability and diagnostics, and fixes mixed per-gate environment-sensor generation for EMU configurations.

## Blobifier bucket switch reliability

A Blobifier bucket `gcode_button` can report a release transition during Klipper startup. Previously, Happy Hare treated that startup transition as a real bucket removal and reset the persisted Blobifier blob count—even though the bucket had not actually been removed.

This PR adds a 0.5-second debounce to the bucket-presence switch and introduces an initialization guard. Bucket-release events are ignored until Blobifier has restored its saved state and completed initialization. This prevents a restart from incorrectly clearing the blob count.

Bucket insert, removal, and ignored-startup events now use `RESPOND` messages so their behavior is visible in the Klipper console and log.

### `BLOBIFIER_STATUS`

This PR adds a `BLOBIFIER_STATUS` diagnostic macro.

```gcode
BLOBIFIER_STATUS
```

It reports the current runtime Blobifier values:

- `current_blobs`
- `last_shake`
- `next_shake`

It also reports the persisted Blobifier values loaded from saved variables:

- Saved `current_blobs`
- Saved `last_shake`

If no saved Blobifier state is available, the macro reports that clearly. This makes it easier to determine whether a count issue is caused by runtime logic or by state save/restore behavior.

## Per-gate environment sensor type fix

Mixed EMU environment-sensor configurations previously generated every sensor using lane 0's sensor type.

For example, configuring:

- Lane 0 as `AHT2X`
- Lanes 1–4 as `BME280`

would generate all sensors as `AHT2X`, even though their addresses, MCUs, and I²C buses were correctly generated per lane.

The cause was that the environment sensor type was defined as one global Kconfig parameter and then rendered as one global template value.

This PR makes the sensor type a per-gate Kconfig parameter and renders the matching per-gate value in each generated `temperature_sensor` section.

Mixed `AHT2X` and `BME280` configurations now generate the correct sensor type, address, MCU, and I²C bus for each lane.

## BME280 report-time option

This PR adds a BME280-only report-time setting to `menuconfig` for shared and per-gate environment sensors.

The setting:

- Appears as `Report time (secs)` when `BME280` is selected.
- Defaults to `120` seconds.
- Generates `bme280_report_time` in the corresponding `temperature_sensor` section.
- Omits `bme280_report_time` when set to `0`.

Example generated output:

```ini
[temperature_sensor unit0_Env2]
sensor_type           : BME280
i2c_address           : 118
bme280_report_time    : 120
i2c_mcu               : unit0_gate2
i2c_bus               : i2c2_PB10_PB11
```

## Validation

- Rendered a mixed EMU configuration containing `AHT2X` and `BME280` sensors.
- Verified that each lane generates its configured sensor type and transport settings.
- Verified that BME280 sensors generate aligned `bme280_report_time` entries with the expected default.
- Ran installer configuration tests successfully.


moggieuk:
Thanks! On the `bme_report_time` part of this fix.  I think you have found a generic oversight and it would be worth fixing generically.

I.e. keeping the SINGLE "PARAM_ENVIRONMENT_SENSOR_REPORT_TIME".   It can have different ranges for different sensor types BTW with multiple "range 0 300 if XXX" constructs

Then have a promptless:
```
config PARAM_SENSOR_REPORT_TIME_PARAM
  string
  default "bme_report_time" if ...
  default "htu21d_report_time" if
  default "lm75_report_time" if ...
  default "aht10_report_time"
  ```

Then in the template so something like:
[% if PARAM_ENVIRONMENT_SENSOR_REPORT_TIME %]
[[PARAM_SENSOR_REPORT_TIME_PARAM]]     : [[PARAM_ENVIRONMENT_SENSOR_REPORT_TIME]]
[% endif %]

Then it is solved for all sensor types.  [really annoying they don't share a common parameter name]

cwiegert:
I knew you would find a better way to implement the change so it works generically.    :-)

PR #1190, opened 2026-09-06
Allow explicit per-gate LED counts

moggieuk:
## Summary

- add optional exit_led_counts and entry_led_counts settings for uneven gate layouts
- keep the existing equal distribution when either setting is empty
- centralize the gate-to-virtual-LED mapping so static updates and animations use the same indexes
- expose gate-sized array editors only when the corresponding LED segment is configured
- validate entry count, positive values, and the total against the virtual chain at runtime

## Example

    [mmu_leds unit0]
    exit_leds: neopixel:my_leds (1-10)
    exit_led_counts: 1, 1, 2, 6

The count setting is normally left empty. Existing configuration line breaks retain their current physical-chain composition meaning and do not implicitly change gate boundaries.

## Tests

- make ALL=1 test: 1,358 tests passed across 47 files
- focused menuconfig, profile, LED mapping, validation, animation, static-color, compatibility, and console grouping coverage passed after the Kconfig follow-ups

### DIFF
--- a/config/base/mmu_hardware.cfg
+++ b/config/base/mmu_hardware.cfg
@@ -791,9 +791,9 @@ color_order   : [[PARAM_COLOR_ORDER]]
 #
 # Note the use of separate lines for each part of the definition,
 #
-# By default the exit and entry virtual chains are divided equally between gates. To use
-# different counts, provide one positive count per gate; the values consume the virtual chain
-# in gate order and must total its length. For example, 10 exit LEDs split over four gates:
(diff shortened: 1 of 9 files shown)


margaale:
This looks fine. Im in the middle of a long print so i cant test it right now. Thanks for making the change though

moggieuk:
When you confirm I can merge.  The benefit of this approach is that there is zero impact to current set-up's

margaale:
image
i know that you have already merged it but i just tested it and it works like it should now

PR #1196, opened 2026-09-07
Default EMU managed fans to MCU temperature

burkfers:
Defaults the fan temperature source to mcu for the EMU, in accordance with upstream defaults.
A test is included, and I have verified on-printer.

The set points do not require tweaking from the global defaults, despite not being identical to previous v3 defaults, as verified with EMU upstream.

:robot: Disclosure: Generated with AI assistance.

### DIFF
--- a/.claude/skills/kconfig-menuconfig/SKILL.md
+++ b/.claude/skills/kconfig-menuconfig/SKILL.md
@@ -146,6 +146,8 @@ installer/
                           # fans, leds, encoder, espooler, nfc_reader, ...
                           # Convention: "# Sets/Defines parameter tokens:" header
   mmu_types/ (+ starters/)  # rsource "Kconfig.*" — one file per machine type
+                            # (each inlines Kconfig.mmu_additions; see the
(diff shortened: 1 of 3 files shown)


moggieuk:
We need to keep the mmu type default out of the specific feature Kconfig files else it will get impossibly difficult to manage.  If you add:
```
choice CHOICE_DEFAULT_FAN_TEMPERATURE_SOURCE
  default CHOICE_DEFAULT_FAN_TEMPERATURE_SOURCE_MCU if FAN_HAS_MCU_TEMPERATURE
endchoice
```
Into Kconfig.emu then it will work exactly the same, but keep out of the generic Kconfig.

Note the defaults in these feature Kconfig's are themselves "generic"

burkfers:
The default is now placed in the machine file, and I have amended the `kconfig-menuconfig` AI skill to make this requirement absolutely clear. Additionally, there's a section about the include order of Kconfig symbols which should help save research tokens on similar tasks.

PR #1204, opened 2026-09-10
Restore Kalico compatibility

burkfers:
## Requires https://github.com/KalicoCrew/kalico/pull/932 (motion_queuing downstreaming)!

With two core changes, Kalico compatibility seems restored
1. The signatures of `manual_home()` diverged and the code now branches to call the correct version
2. `mmu` and `mmu.unit` are now full packages via addition of `__init__.py`. This has no effect on Klipper but enables Kalico to load them. This used to be in place in v3 but got lost during the transition to v4.

Hardware testing I'd like to complete:
- Kalico:
    - [x] boot
    - [x] mmu park, nfc load
    - [x] toolhead load
    - [x] full print
    - [x] Bowden calibration (standard an `MANUAL=1` 
 - Klipper:
    - [x] boot
    - [x] mmu park, nfc load
    - [ ] toolhead load
    - [ ] full print

Ready for review save for the outstanding Kalico merge.


burkfers:
Superceded by #1206.
Can be closed if that one merges. Remains as alternative approach should #1206 be rejected.

PR #1206, opened 2026-09-11
Restore compatibility with Kalico `main`

burkfers:
Whew, this is a big one...

Happy Hare crashes at boot on Kalico, and on-printer testing on a real Kalico
machine then surfaced further generation gaps of the same class. This PR fixes
them, entirely within Happy Hare, **without depending on the motion queuing PR
on Kalico**. On any generation that has `klippy/extras/motion_queuing.py`,
nothing here changes behaviour: the emulator is never constructed, `do_enable`
takes the existing path, the homing branch is not taken, and the step-generator
sync is a no-op on toolheads without the API.

Five fixes, one commit each:

1. **Import.** Forks that rewrite imports into a `klippy` package (Kalico's
   `klippy/compat.py`) cannot import namespace packages: their plugin loader
   patches `spec.loader.exec_module`, and a namespace package spec has no
   loader, so plugin loading dies at startup with
   `'NoneType' object has no  attribute 'exec_module'`.
  `mmu/` and `mmu/unit/` become regular packages  with the two added
  `__init__.py` files.

2. **Boot.** `Module 'motion_queuing' not found` at `extras/mmu_stepper.py`:
   the module does not exist on older generations (Kalico `main`, old mainline).
   `load_object` now uses a None default and, where the module is absent, a
   `MotionQueuingEmulator` implementing the six methods MmuStepper calls:
   trapq alloc and finalize through the chelper ffi, plain moves step the
   drive and its manual followers then note movequeue activity on the
   toolhead, drip moves hand the steppers to
   `toolhead.drip_update_time()`. The implementation follows
   `klippy/extras/manual_stepper.py`, which Klipper ships in every generation.
   Selection is on module presence, not the fork: a generation that backports
   motion_queuing uses the real module with no follow-up.

3. **MMU_BOOTUP.**
   `'PrinterStepperEnable' object has no attribute 'set_motors_enable'` on older
   `stepper_enable` generations. `do_enable` now checks for the method and,
   where absent, mirrors its semantics (step-generation flush, no-op skip,
   `DISABLE_STALL_TIME` dwells) via the per-stepper API that exists in both
   generations.

5. **Homing.** Kalico's `manual_home` predates the `probe_pos` argument and
   reports the move target as the position, so the seven-argument positional
   call crashes on Kalico, and without the fix bowden calibration would
   measure the homing max instead of the tube length. `do_homing_move` now
   drives `HomingMove` directly with `probe_pos` on Kalico, the approach the
   verified v3 code used, and keeps the `manual_home` call on Klipper.
   `is_kalico()` in `mmu_utils` detects the fork via the `danger_options`
   object Kalico ships.

6. **Drip moves.** `Internal error in stepcompress` on real Kalico: older
   toolheads register every ExtruderStepper as a step generator, and toolhead
   step generation advances the stepper's time cursor past the manual trapq,
   so a drip move walks off the end of it. MmuStepper now keeps manual-mode
   steppers out of `toolhead.step_generators` and re-registers them on
   sync-to-extruder, as Klipper's own manual_stepper always did. Newer
   toolheads have no `step_generators` API, so this is a no-op there.

The installer now warns instead of aborting on Kalico; `-o` keeps its
remaining consumer (the MIPS layout check). A note in `test/README.md` records
that suite runs stay on `make`, which keeps AI agents from improvising their
own invocation.

The fake klippy tree can now represent a pre-motion_queuing generation:
`Session(kalico=True)` / `Session(old_klipper=True)` remove the module,
install a pure-Python chelper ffi stand-in, and give the toolhead and
stepper_enable their old surfaces. Each fix has a regression test that fails
without it (`test_mmu_motion.py`: TestKalico / TestOldKlipper, plus the
import and Kalico-homing tests). Full suite: 1411 passing.

Verified on hardware, Kalico on an EMU: boot, MMU_BOOTUP,
MMU_CHECK_GATE, MMU_TEST_MOVE, bowden calibration (manual and auto),
load, swap and unload, and a two-colour test print with a mid-print toolchange.

P.S.: I recommend a squash merge when it is time. I have deliberately exposed the making of the sausage to aid review.


burkfers:
Late addition commit: `manual_home()` now is introspected for signature, no longer branching by is_kalico. This ensures nothing breaks should Kalico downstream the changes leading to the new signature, and *may* enable support for ancient Klipper versions (not tested; out of scope)

moggieuk:
One compatibility gap found while reviewing `2c3ceee`: **G-code-axis mode needs to restore step-generator registration on Kalico.**

Kalico's older motion architecture separates two operations:

1. Registering an extra axis, so the toolhead can plan and append its movements to a trapq.
2. Registering its `generate_steps` callback, so queued movements become motor steps.

`toolhead.add_extra_axis()` does the first operation, but does not do the second. Kalico's own `manual_stepper.command_with_gcode_axis()` explicitly registers the step generator when assigning an axis, and unregisters it when removing the axis.

Modern Klipper manages step generation through `motion_queuing` instead, so it does not need this old toolhead callback registration.

The PR correctly removes standalone manual MMU steppers from Kalico's `step_generators` list to avoid advancing their step-generation time ahead of manual homing. However, `_sync_step_generator()` restores registration only in extruder mode. A stepper assigned a G-code axis still has `motion_mode == MODE_MANUAL`, and `MmuStepper.command_with_gcode_axis()` does not restore registration.

Consequently, after `MMU_STEPPER STEPPER=<name> GCODE_AXIS=A`, a subsequent `G1` move on axis A can queue movement and update the software position without generating the corresponding motor steps. I reproduced successful axis assignment with the step-generator callback absent in the PR's Kalico harness; this particular sequence has not been tested on hardware. This is a less common Kalico operating-mode issue, not a regression identified in modern Klipper or ordinary MMU load/unload operations.

### Suggested fix

Keep the legacy handling in `_sync_step_generator()`, but treat either extruder mode or an assigned G-code axis as requiring toolhead step generation:

```diff
-        if self.motion_mode == self.MODE_EXTRUDER:
+        if (self.motion_mode == self.MODE_EXTRUDER
+                or self.axis_gcode_id is not None):
```

Update its comment to clarify the distinction:

```python
        # Kalico/legacy Klipper compatibility:
        # A standalone manual stepper must not be stepped by the toolhead:
        # toolhead step generation runs ahead of the print time, and the
        # advanced time cursor breaks drip moves against the manual trapq
        # ("Internal error in stepcompress").
        # Extruder-mode steppers and manual steppers registered as G-code
        # axes must instead be stepped by the toolhead.
```

Then call the helper in `MmuStepper.command_with_gcode_axis()` in `extras/mmu_stepper.py` (line 1073 in this revision), after changing the axis state:

```diff
             # Unregister
             toolhead.remove_extra_axis(self)
             self.axis_gcode_id = None
+            self._sync_step_generator()
             return
```

```diff
         self.gaxis_limit_accel = limit_accel
         toolhead.add_extra_axis(self, self.commanded_pos)
+        self._sync_step_generator()
```

Keep the calls after `add_extra_axis()` / `remove_extra_axis()`: those operations flush pending toolhead moves, allowing them to finish under the previous registration state. On modern Klipper, the helper still returns immediately because its toolhead has no `step_generators` list.

One focused legacy regression test should cover the transition: initially manual → callback absent; assign axis A → callback present exactly once; remove axis A → callback absent; assign it again → callback present exactly once. This directly checks the missing behavior without another simulated load/unload test. The existing fake motion layer does not generate physical steps, so position assertions alone would not catch this.

For context, all 1,411 existing PR tests passed in an isolated snapshot during review. The changes above are a proposal and have not been applied or tested as a patch.

burkfers:
I have implemented the proposed fix with accompanying test.

This would never have been hit in current-day normal HH usage, but makes the implementation more universally correct and ready for new usage patterns.

moggieuk:
Pushed a follow-up cleanup to make the code added for Kalico compatibility easier to identify and potentially remove when Kalico supports the required APIs natively.

This groups compatibility imports and helpers, adds clear BEGIN/END delimiters and call-site comments, and retains the detailed step-generation explanations. The homing fallback is extracted into the compatibility section; runtime behavior is preserved.

Validation: all 1,413 tests passed before the final comment restoration, and all 80 motion/import tests passed again on the final version. The final syntax comparison confirms equivalent logic to the original PR after accounting for the refactoring. Both new package markers remain included by the Makefile.

PR #1207, opened 2026-09-11
Allow extruder without a TMC section — warn instead of raising

burkfers:
The Problem

Since v4.0.0, `MmuExtruderWrapper.__init__` raises config.error unless a [tmc<chip> extruder] section exists. Printers whose mainboard has no UART to the extruder driver (e.g. Ender 3 V3 SE + MMX) therefore cannot start klippy at all - a regression: v3.4.x had no such requirement and these machines booted.

The requirement is also stricter than the code needs: the runtime already tolerates a missing extruder TMC (tip-forming current control checks for None; the stallguard "touch" endstop only exists if the user configured it, which implies the section).

 Changes

 - MmuExtruderWrapper guards the TMC load with config.has_section() — the same pattern every other TMC load in HH uses (gear steppers, selectors) — and logs a warning in the gear-stepper style instead of raising.
 - Side benefit: a genuinely broken TMC section now surfaces its real config error instead of being swallowed by a try/except and misreported as "not found".
 - Harness: the fake TMC loader now raises for a missing section, matching real Klipper (the fake getsection() deliberately doesn't, so the bug was unreproducible before this).
 - New test TestNoExtruderTmc (mmx profile, stub stripped of [tmc2209 extruder]): fails with the reported error before the fix, passes after.

Behavior is unchanged when a TMC section is present.

I have verified "TMC-having extruder is unaffected" in hardware. I have not tested the efficacy of the fix itself for lack of hardware.

 :robot: Disclosure: This changeset was created with LLM assistance. All outputs are manually reviewed and tested to the best of my abilities.

### DIFF
--- a/extras/mmu/unit/mmu_extruder_wrapper.py
+++ b/extras/mmu/unit/mmu_extruder_wrapper.py
@@ -43,19 +43,22 @@ class MmuExtruderWrapper():
 
         self.homing_extruder_stepper = None
 
-        # Ensure corresponding TMC section is loaded so endstops can be added and to prevent error later when toolhead is created
+        # Load the extruder TMC section if present - it enables software current control
+        # (tip forming) and the stallguard touch endstop. Optional: without it the driver
+        # settings are controlled in hardware, as on printers with no UART to the extruder.
         self._extruder_tmc = None
         for chip in TMC_CHIPS:
-            try:
-                section = f"{chip} {self.name}"
+            section = f"{chip} {self.name}"
+            if config.has_section(section):
                 self._extruder_tmc = self.printer.load_object(config, section)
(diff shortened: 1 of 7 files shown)


burkfers:
Draft for lack of testing. I have asked the issue-reporting community members for assistance.

moggieuk:
Thanks.  I did review and it looks good.  At least -- no impact to existing code logic other than surfacing a previously "bad" extruder config earlier in the Klipper boot process.

Have you asked the user who pointed to this limitation to test?

burkfers:
The user who reported the issue has tested in hardware (tmc2208 using pure step/dir interface, uart not wired up on this board) and has reported success - Issue is gone.

PR #1208, opened 2026-09-13
EndlessSpool on load: refresh the target gate's status before testing the trigger

RT530:
## The gap

EndlessSpool-on-load is gated on the mapped gate reading **exactly** `GATE_EMPTY`
(`mmu_controller.py:3052`). Nothing corrects `gate_status` between bootup and a load —
`validate_gate_status()` has only two callers, `cmd_MMU_BOOTUP` and `reset_gate_map()`. So a gate that
is empty in reality but still recorded as `GATE_UNKNOWN` never enters the EndlessSpool path at all,
and the load proceeds into the empty gate.

This happens with a correct group, `endless_spool_enabled: 1` and `endless_spool_on_load: 1` all in
place. The user configured a partner gate, and the error names the gate they configured it for.

The same gap exists on the print-start path, for a different reason. `MMU_START_CHECK` runs
`MMU_CHECK_GATE TOOLS=...`, which finds the gate empty, marks it `GATE_EMPTY` and then **raises**
(`mmu_check_gate.py:191`, when `is_in_print()`), pausing the print before `MMU_START_LOAD_INITIAL_TOOL`
is ever reached. So the load-path fix alone never gets its turn there. `MMU_CHECK_GATE` has no
EndlessSpool logic of its own, so a configured group cannot help at the one moment users most expect
it to.

## Before / after

Gates 0 and 1 in one EndlessSpool group, T0 → gate 0, gate 0 physically empty but marked
`GATE_UNKNOWN`, gate 1 available:

| | Before | After |
|---|---|---|
| `MMU_CHANGE_TOOL TOOL=0` | loads gate 0, fails | remaps T0 → gate 1 and loads it |
| print start (`MMU_START_CHECK`) | pauses: `Required Tool T0 on gate 0 marked EMPTY` | remaps T0 → gate 1, verifies it, print continues |
| `gate_status[0]` | stays `-1` | corrected to `0` from its own sensor |
| gate 0's material / temp / spool id | destroyed by the failed load's `set_gate_status` | preserved |

## The change

Two commits, one per path.

### 1. The load path — refresh the target gate's status before the trigger

`_select_and_load_tool` refreshes **just the gate it is about to load**, from that gate's own sensors,
immediately before the trigger.

- Scoped to `[gate]` — loading one tool never rewrites another gate's status.
- Skipped for the bypass (`tool < 0`), where `gate` comes from `gate_selected`.
- A no-op on machines with no gate sensors: `check_gate_sensor()` returns `None`, neither correction
  branch fires, status is untouched. Covered by a test on a sensorless profile.
- Type-B machines with exit sensors are unaffected — `gear_detected is True` short-circuits to
  `GATE_AVAILABLE` before the entry sensor is consulted.
- No persistence added. `validate_gate_status` deliberately rebinds the list so `get_status()`
  webhooks see the change, without writing to `variables.cfg`; that is left as-is.

### Why `clear_attributes=False`

`validate_gate_status()` gains a `clear_attributes` parameter, **default `True` so both existing
callers behave exactly as before**. The new call passes `False`.

A sensor reporting a lane empty is not the same claim as *"I no longer know what filament belongs to
this lane"*. Nothing was ejected, so the gate's recorded material, colour, temperature and spool id
remain the best record of what belongs there — and the load is about to read that temperature to heat
the extruder. Clearing it at that moment is exactly how a load ends up at the 200 °C fallback instead
of the spool's temperature.

`reset_gate()`, which **is** a genuine ejection, still clears. Upstream already agrees with this
distinction elsewhere: `reset_gate_map()` carries the comment that metadata is *deliberately*
preserved on a gate whose validated status is EMPTY.

### 2. The print-start path — substitute a gate during the pre-print check

When `MMU_CHECK_GATE` finds a *referenced tool's* gate empty, it now consults the tool's EndlessSpool
group the same way the runout path does and remaps to a gate that still has filament, instead of
raising and pausing the print.

- The replacement is appended to the work list so it is **itself checked** before the print proceeds.
  Substituting a gate nobody verified would only move the failure later into the print.
- Appending to the list under iteration is deliberate and terminates: each failed check marks that
  gate `GATE_EMPTY`, and `get_next_endless_spool_gate()` only returns gates that are not `GATE_EMPTY`,
  so a group is exhausted in at most `num_gates` steps and the original error is raised.
- **Unchanged when there is no alternative:** whole group empty, tool alone in its group, or
  `endless_spool_on_load` unset — it still raises and the print still pauses. A gate check outside a
  print still only logs, as before.

## Tests

`test/test_mmu_endless_spool.py`, six new tests:

- the remap away from an UNKNOWN-but-empty gate, and that the gate ends up `GATE_EMPTY`
- **gate metadata survives the correction** — guards the flag above
- only the target gate is revalidated
- a gate that turns out to be **full** is loaded, not remapped away from — `GATE_UNKNOWN` means
  "nobody has checked", not "empty"
- a machine with no gate sensors is unaffected and raises nothing

plus four for the print-start path: that the check remaps rather than pausing, that the substitute
gate is itself verified, that a fully empty group still stops the print, and that a tool alone in its
group is never remapped outside it.

Six of these fail without the corresponding change. Full suite: **1411 tests, green**.

An `@unittest.expectedFailure` records a remaining asymmetry this PR does **not** fix: the trigger
demands `== GATE_EMPTY` while `get_next_endless_spool_gate()` accepts `!= GATE_EMPTY`, so an unchecked
gate cannot trigger a switch but *can* be chosen as the gate to switch to. It will self-heal when
someone fixes that on its own branch.

## Verified on hardware

Tested on a 12-gate ERCF (entry sensors only, no exit or toolhead sensor, `LinearServoSelector`),
gates 0 and 1 in one EndlessSpool group, T0 mapped to gate 0, gate 0 physically empty. Same command
and same starting state in each run; only the code differs:

| Code | `gate_status[0]` | Result |
|---|---|---|
| unpatched | `GATE_EMPTY` | remaps to gate 1 — the path works when status happens to be correct |
| unpatched | `GATE_UNKNOWN` | **no remap.** Goes straight to `Loading gate 0` and fails on the empty gate |
| this PR | `GATE_UNKNOWN` | **remaps to gate 1** |

With the change, from `mmu.log`:

```
Loading tool T0...
Gate 0 is empty! Checking for alternative gates for T0 in EndlessSpool Group B (checked gates: 1)
Remapping T0 to gate 1
Tool T0 enabled on gate 1
```

`ttg_map[0]` persisted as `1`, and gate 0 kept `material=PLA` / `temperature=220` across the
UNKNOWN→EMPTY correction — the `clear_attributes=False` path above. On the same run a *different*
gate hit the existing `set_gate_status(GATE_EMPTY)` path and had its material and temperature reset to
defaults, which is the behaviour this PR deliberately leaves alone.

## Notes

- Based on `upstream/main`, so this branch carries this change and nothing else.
- Found while writing the tests, **not fixed here**: `MMU_GATE_MAP`'s `TEMP` defaults to
  `default_extruder_temp` rather than the gate's current value (`mmu_gate_map.py:288`), unlike every
  sibling attribute which defaults to `None` = leave alone. So any call omitting `TEMP` silently
  resets that gate's temperature. Separate one-line fix, separate branch.

https://claude.ai/code/session_01J3rj2raMwJyGf8CtM9GGfQ


moggieuk:
I found a P1 issue in the new `MMU_CHECK_GATE` EndlessSpool fallback: it treats **“filament was found, but could not be parked” as “the gate is empty.”** Those need different recovery actions.

### Why this happens

For each gate, the check loads filament far enough to confirm it exists, then unloads it back to its parked position so another gate can be selected safely. Both operations are covered by the same `except MmuError` handler.

Previously, either failure stopped the check during a print. The new fallback at `extras/mmu/commands/mmu_check_gate.py:196` can substitute another gate for either failure, including an `_unload_gate()` failure after successful detection.

### Concrete example

T0 maps to gate 0; gates 0 and 1 are in the same EndlessSpool group and both contain filament. EndlessSpool on load is enabled.

1. Print start runs `MMU_CHECK_GATE TOOLS=0`.
2. `_load_gate()` succeeds on gate 0, physically advancing and detecting filament.
3. `_unload_gate()` fails to park that filament.
4. The shared handler marks gate 0 `GATE_EMPTY` and sets the software filament position to `FILAMENT_POS_UNLOADED`. **Updating that software state does not retract the physical filament.**
5. The new fallback remaps T0 to gate 1 and appends gate 1 to the check list.
6. The command selects and loads gate 1 despite unresolved filament in gate 0.

On a moving-selector machine, this could attempt to move the selector while filament still crosses it. On a machine with separate lane drives, it could feed another lane into an occupied shared path. The exact consequence depends on the hardware, but safe parking of the previous filament has not been established.

### Reproduction

On PR head `87d263f77a51a1dfca14db3a8c06b54ba4f6fb73`, I used the existing EndlessSpool harness with filament in all four gates, enabled `endless_spool_on_load` and `test_force_in_print`, and injected an `MmuError` from `_unload_gate()` for gate 0 while allowing the other gates to use the original method. Running `MMU_CHECK_GATE TOOLS=0` produced:

```text
unload attempts: [0, 1]
ttg_map: [1, 1, 2, 3]
errors: []
gate 0 filament tip: 0.0 (left unparked)
```

This is a software harness reproduction, not a hardware damage test. The existing EndlessSpool suite also passed: 29 tests, including one expected failure.

### Suggested fix

Separate an empty-gate result from a parking or motion failure, and permit substitution only for the empty-gate result.

- **Minimum fix:** Move the parking operation outside the error handler that performs substitution. A parking failure must stop the check without remapping or attempting another gate.
- **More robust fix:** Also distinguish confirmed absence of filament from other `_load_gate()` failures. A general `MmuError` should not automatically authorize selecting another lane.

Conceptually (the specific exception below is illustrative):

```python
try:
    load_gate()
except GateEmptyError:
    mark_gate_empty()
    substitute_or_report_empty()
else:
    mark_gate_available()
    unload_gate()  # Failure propagates and stops the check.
```

Please add a regression test where gate 0 loads successfully but parking fails. Assert that the error is reported, T0 remains mapped to gate 0, and gate 1 is never checked or loaded. Retain the existing test proving that a genuinely empty gate successfully substitutes gate 1.

RT530:
You're right, and thank you — this was a real hazard and the reproduction made it easy to confirm. Fixed in 735c7f54.

The conflation was exactly as you described: `_unload_gate()`'s failure was re-raised as `MmuError` and caught by the same handler that means "this gate is empty", so the substitution I added acted on a gate whose filament had been successfully detected and then left unparked.

**What changed**

A flag records which operation failed. Only a `_load_gate()` failure marks the gate empty and may substitute; selection and parking failures propagate, so the check stops, nothing is remapped, and no other gate is selected or loaded.

I also took your "more robust" point about the gate status. A parking failure no longer marks the gate `GATE_EMPTY` — `_load_gate()` had just proved it holds filament, so recording it as empty is untrue and would additionally let the *next* load remap away from a perfectly good gate. That is a behaviour change relative to the pre-existing code, which marked it empty on either failure; flagging it explicitly in case you would rather I preserve the old marking and restrict this commit to the substitution path alone.

```python
try:
    _ = mmu._load_gate(allow_retry=False)
except MmuError:
    gate_is_empty = True
    raise
...
except MmuError as ee:
    if not gate_is_empty:
        raise   # found but not parked: stop, do not remap, do not touch another gate
```

**Tests** — four added, in `TestGateCheckParkingFailureIsNotAnEmptyGate`, injecting a parking failure on gate 0 while every gate keeps its filament:

- the failure is reported rather than swallowed
- `ttg_map[0]` stays `0`
- gate 1 is never selected, and no gate is loaded past the extruder
- gate 0 is not marked `GATE_EMPTY`

Without the fix all four fail, matching your reproduction exactly: unload attempts `[0, 1]`, `ttg_map` `[1, 1, 2, 3]`, `errors: []`, gate 0 empty. The genuinely-empty-gate substitution test is unchanged and still passes.

Full suite: 1415 tests, green.

moggieuk:
Sorry -- the test breakage is mine.  Fixing now

moggieuk:
Please exclude the impersonal AI review but I have verified the finding..

**[P2] Preserve buffered filament status during the new pre-load sensor refresh.**

The new call to `validate_gate_status()` in `_select_and_load_tool()` runs before every tool load. When a gate's exit sensor is triggered, that helper unconditionally changes its status to `GATE_AVAILABLE` (1), even if it was `GATE_AVAILABLE_FROM_BUFFER` (2).

This affects configurations where parked filament still covers the exit sensor. In the harness, using forward parking, unloading left the gate at status 2 with its exit sensor triggered; the next load entered `load_sequence()` with status 1. On machines configured with a filament buffer, this discards the information used to select `gear_from_filament_buffer_speed` and `gear_from_filament_buffer_accel`, so loading uses the ordinary spool-loading settings instead.

**Suggested fix:** When the exit sensor confirms filament is present, preserve an existing buffered status—for example, replace the unconditional assignment in `validate_gate_status()` with:

```python
v_gate_status[gate] = max(status, GATE_AVAILABLE)
```

This matches the existing availability update in `_home_to_gate()`. Add a regression test with a buffered gate and a triggered exit sensor, asserting that the pre-load refresh preserves `GATE_AVAILABLE_FROM_BUFFER`.

The previously reported P1 parking-failure issue is fixed in the updated revision. The reviewed test run passed 174 tests with one expected failure.

RT530:
Confirmed and fixed in 7b3ed1c5 — thank you, and no need to apologise about the test breakage.

You're right about the mechanism, and the aggravating factor is mine: `validate_gate_status()` has always assigned `GATE_AVAILABLE` unconditionally, but while it only ran at bootup and on an explicit map reset the downgrade was rare enough to go unnoticed. The pre-load refresh in this branch runs it before *every* load, so a gate parked with filament still covering its exit sensor loses `GATE_AVAILABLE_FROM_BUFFER` every time — and with it the buffer speed and accel for the load.

Applied your suggestion verbatim, since it is also what `_home_to_gate()` and `MMU_CHECK_GATE` already do:

```python
if gear_detected is True:
    v_gate_status[gate] = max(status, GATE_AVAILABLE)
```

**Tests** — `TestBufferedStatusSurvivesTheRefresh`, with the gate parked past its exit sensor and marked buffered:

- the refresh preserves `GATE_AVAILABLE_FROM_BUFFER` — fails without the change (`1 != 2`)
- an `UNKNOWN` gate is still promoted to `GATE_AVAILABLE`, so `max()` cannot freeze a stale lower status

Full suite: 1417 tests, green.

On the earlier point I raised: a parking failure still does not mark the gate `GATE_EMPTY` in this revision, since `_load_gate()` has just proved filament is present. Happy to restore the old marking if you would rather keep that change out of this PR.

moggieuk:
Hi!  I've merged this PR into PR1216.  The reason is that is highlighted a lot more assumptions that weren't correct in MMU_CHECK_GATES and so I used to opportunity to fix those as well.

Given that PR1216 is really large I've marked it for the v4.1 release.  This release won't be long and will also carry Kalico support and an new LED themes mechanism.

moggieuk:
Closing this in favour of #1216, but I want to be precise about what that means, because
most of this PR is in there.

Your diagnosis was right, and it was the useful kind of right: you found a gap, reproduced
it on hardware, and the write-up was clear enough that I could go looking for the same
assumption elsewhere in `MMU_CHECK_GATE` — which is where the rest of #1216 came from. It
grew past your PR because of what yours exposed, not because there was anything wrong with
it.

What carried over, and whose it is:

| commit in #1216 | |
|---|---|
| `EndlessSpool on load: refresh the target gate's status before the trigger` | **yours**, unchanged — same fix at the same site |
| `Allow gate status changes to retain filament metadata` | **yours** — the `clear_attributes` distinction, and your reasoning for it |
| `Preserve buffered availability when validating gate status` | the `max(status, GATE_AVAILABLE)` fix — my report, your implementation and tests |
| `Substitute an EndlessSpool gate during MMU_CHECK_GATE` | your idea and your termination argument; I reworked the implementation (below) |

All six of your test classes came across unmodified, and I've added `Co-Authored-By` trailers
to those commits so this is recorded in the history rather than only here.

**Where #1216 deliberately does less than this PR, so you're not surprised by it later.**
Your version let any `_load_gate()` failure authorise a print-start remap. After the P1 you
fixed, that narrowed to load failures only — but a failed pickup still isn't the same
evidence as a switch saying the lane is clear, and I didn't want a remap resting on it. So
#1216 substitutes only where a per-gate entry switch positively reads empty.

For your 12-gate ERCF that's a straight improvement: the entry switch reports empty, the
remap happens, and no gear motion is attempted at all — where this PR spun into the empty
lane first. But on machines with no per-gate entry sensors (Tradrack, 3D Chameleon,
PicoMMU/MMX, encoder-homing ERCF) the print-start substitution your PR gave them isn't
there any more. They still get the on-load fix. That's a trade I made on safety grounds and
I'd rather name it than have you find it yourself.

#1216 is marked for v4.1, which also carries Kalico support and the new LED themes
mechanism, so it won't be long.

Thanks for this — both the fix and the two rounds of review response were a pleasure to
work with.

PR #1212, opened 2026-09-15
Let a proportional sync-feedback sensor report neutral

RT530:
### The problem

On a machine with a proportional (type-P) sync-feedback buffer, `get_sync_feedback_string()` can never
return `neutral`.

It sign-tests whatever `_get_sensor_state()` hands back:

```python
return 'compressed' if state > 0 else 'tension' if state < 0 else 'neutral'
```

For a proportional sensor that value is a float straight off the ADC, so `neutral` requires the
reading to be exactly `0.0` — which an analog sensor essentially never produces. The result is that a
type-P machine reports `tension` or `compressed` and nothing else, no matter where the buffer actually
sits. On my printer a buffer resting at `+0.084` reported `compressed`.

Switch-based (type-D) machines are unaffected: they already get a discrete `{-1, 0, 1}`, so their
neutral works.

### The fix

Ask `_get_sensor_state()` for the discrete state instead:

```python
state = self._get_sensor_state(use_virtual_threshold=True)
```

That reads the virtual tension/compression sensors, which the proportional sensor already derives by
threshold, so the reported string now agrees with what Happy Hare itself believes about the buffer
rather than applying a second, stricter rule to the same reading. Those sensors assert at
`±analog_sensor_threshold` and release one hysteresis interval inward, so the state cannot chatter at
the boundary. No new configuration.

The same call is already used for exactly this purpose in `check_filament_in_mmu()`.

### Why not the commented-out `ctrl.polarity()` block

There is a deadband version sitting commented out two lines above, and `git log` says it was live
before the v4 refactor. Restoring it as-is would crash: `self.ctrl` is `None` outside an active sync,
and `detail=True` reaches that line from both `MMU_STATUS` and `MMU_SYNC_FEEDBACK`, so it would
`AttributeError` on a machine that has not synced yet. `_get_sensor_state()` has no `self.ctrl`
dependency. Worth noting the two paths would land on the same threshold anyway, since
`flowguard_extreme_threshold` is assigned from `analog_sensor_threshold`. I have left the commented
block alone — removing it seemed like your call rather than mine.

### Scope

Reporting only. The controller consumes the raw float and its own thresholds, so rotation-distance
correction, FlowGuard and tangle prevention are untouched. The visible changes are the `MMU_STATUS` /
`MMU_SYNC_FEEDBACK` text and the buffer indicator in the UI, which has always had a neutral state it
could never reach.

One knock-on worth flagging: with neutral now reachable, the filament display's existing branch at
`mmu_controller.py:1053` renders the modelled bias (`[ -0.3 ]`) instead of `N` for a proportional
buffer. That branch was previously unreachable on a type-P machine.

### Tests

Adds `test/test_mmu_sync_feedback.py` — `get_sync_feedback_string()` had no coverage. 11 tests:
the mid-range regression on `emu`, both extremes, the hysteresis band, a parity check that the string
never disagrees with the derived sensors across the range, and the switch-based path on `kms` to show
it is unchanged.

With the one-line change reverted, 7 of the 11 fail. The 4 `kms` switch tests pass either way, which
is the check that type-D is untouched. Full suite green.

### Verified on hardware

ERCF, type-P buffer in EKF mode:

| buffer value | reported |
|---|---|
| `+0.995` (compression end) | `compressed` — unchanged, extreme still reported |
| `+0.084` (mid-range) | `neutral` — previously reported `compressed` |


moggieuk:
You have identified a hole -- actually some unfinished rework I think when I introduced the concept of "virtual filament sensors" tied to the proportional sensor.  I'm not that your approach is best so I created a variation for comparison.  Superficially they have similar results the different is in not creating another "slightly different" interpretation. Let me know what you think.


I created PR#1214 as an alternative fix I'd like you to review.  This means the status string will always match the virtual sensor states.  I.e. the small hysteresis in value will be incorporated.  The default thresholds should therefore be:

Buffer reading / movement | Tension virtual sensor | Compression virtual sensor | sync_feedback_state
-- | -- | -- | --
Near center, e.g. +0.084 | False | False | neutral
From center to +0.85 | False | False | neutral
Reaches +0.90 or above | False | True | compressed
Returns to +0.88 after compression | False | True | compressed
Falls to +0.864 or below | False | False | neutral
Reaches −0.90 or below | True | False | tension
Returns to −0.88 after tension | True | False | tension
Rises to −0.864 or above | False | False | neutral


So the difference in the PR's would be:
With all sensors enabled, **PR #1212 and PR #1214 produce the same status string and hysteresis**. The difference is how disabled virtual switches are handled.

### PR #1212

`_get_sensor_state(use_virtual_threshold=True)` reads the virtual tension/compression switches through `check_sensor()`, which ignores disabled switches.

Consequently, disabling the compression switch can make a buffer at `+0.95` report `neutral`, even though its enabled analog sensor still detects compression.

### PR #1214

Reads the analog sensor’s existing hysteresis classification directly, independently of derived-switch enablement. This is the same classification that drives the virtual switches; no new thresholds or hysteresis calculations are introduced.

| Situation | PR #1212 | PR #1214 |
|---|---|---|
| All sensors enabled | Same physical state | Same physical state |
| `+0.95`, compression switch disabled, analog enabled | `neutral` | `compressed` |
| Analog sensor disabled | Falls back to enabled discrete switches | Falls back to enabled discrete switches |

**Neither PR changes EKF operation.** Continuous sensor readings, controller thresholds, and control calculations remain unchanged.

PR #1214 also removes the obsolete controller-polarity comments and adds regression coverage for disabled sensors, controller-mode independence, and unchanged continuous controller input.

moggieuk:
Hi RT530,   thank you very much for this PR.  As mentioned about I've merged PR1214 instead which includes all your work as well as fixing some other corner cases.

Please let me know if the PR1214 has missed anything - I don't think so but happy to hear back if not.

PR #1214, opened 2026-09-15
Match proportional buffer status to virtual sensor hysteresis

moggieuk:
## Summary

Proportional buffers currently report tension or compression for every nonzero reading, so even a near-center value such as +0.084 appears compressed. Report the analog buffer's existing virtual-sensor hysteresis state instead: neutral means neither physical extreme is asserted, consistently across EKF and two-level modes.

Expose that classification through a read-only accessor so disabling a derived switch cannot hide an actual buffer extreme. Keep the existing discrete fallback when the analog sensor is disabled, and remove the obsolete controller-polarity comments. Continuous EKF readings, sensor event payloads, and controller thresholds are unchanged; status reads do not call controller polarity or mutate its hysteresis.

Related: #1212. This addresses the same reporting issue while preserving physical classification when derived switches are disabled.

## Validation

56 tests passed:

```sh
venv/bin/python -m unittest -v test.test_mmu_sync_feedback
venv/bin/python -m unittest test.test_mmu_sensor_enable test.test_mmu_adc_compat test.test_mmu_tangle_prevention test.test_mmu_filament_display
```

The nine new tests cover mid-range neutral reporting, hysteresis in both directions, direct extreme transitions, disabled sensors, controller-mode independence, public status, unchanged continuous controller input, and existing switch/event reporting behavior. `git diff --check` passed. No hardware validation was performed.


moggieuk:
Measured the before/after of this change on the `emu` profile with sync-feedback active, since the reporting delta isn't spelled out in the description. Both `get_sync_feedback_string(detail=True)` and the `MMU_STATUS` buffer glyph:

| analog value | before | after |
|---|---|---|
| +0.084 | compressed / `[ C ]` | neutral / `[ 0.1 ]` |
| +0.500 | compressed / `[ C ]` | neutral / `[ 0.5 ]` |
| +0.950 | compressed / `[ ▷C◁ ]` | compressed / `[ ▷C◁ ]` |
| -0.300 | tension / `[ T ]` | neutral / `[-0.3 ]` |

Two things this surfaces that are worth an explicit decision:

1. **At -0.300 the reported direction disappears, not just its magnitude.** `[ T ]` becomes `[-0.3 ]` — the word "tension" goes away while the number still reads negative. That's the phrasing a support report would arrive in.
2. **The buffer glyph now shows a number over most of the range.** The numeric branch in `mmu_controller.buffer_segment` (the `has_sensor(SENSOR_PROPORTIONAL) and sf_value is not None` path) previously needed the value to be *exactly* 0.0, so it was near-unreachable; it's now the common case for type-P. Arguably the improvement that branch was written for, but it isn't mentioned in the description and no test covers the composition of `get_sync_feedback_string()` with `buffer_segment()` (`filament_display_review.py` drives `sync_feedback_state=` explicitly, which is right for a display-rendering review).

Otherwise this verifies clean: full suite green on a merge onto current `main` (1444 tests, 55 files), the new tests genuinely fail on the pre-PR tree, no control path or shipped macro/config/doc reads `sync_feedback_state`, and `SENSOR_PROPORTIONAL` only ever maps to `buffer.proportional_sensor` so the unguarded `get_virtual_sensor_state()` in a polled `get_status()` can't `AttributeError`. Worth noting the change also removes a status read that mutated `_twolevel_hys_state` via `ctrl.polarity()`.

PR #1217, opened 2026-09-15
LED themes: per-unit effect files, menuconfig selection, and an EMU theme

burkfers:
Moves the per-unit LED effect defaults out of `mmu_hardware.cfg` into
swappable theme files, adds a menuconfig selection for them, and ships a
vendor theme for EMU units.

## Commit 1 — the selection (`installer/Kconfig.leds`)

A "LED theme" choice in the LED config menu, stored as `PARAM_LED_THEME`.
Shipped themes are `mmu_leds` and `emu_leds`; `custom` uses a machine-local
file the installer creates once; any other value is taken as the name of a
user-maintained theme file. This commit lands the selection alone so it can
be seen and reacted to before the mechanism behind it ships.

## Commit 2 — the per-unit theme file

The `effect_*` assignments of `[mmu_leds <unit>]` move from
`mmu_hardware.cfg` into `mmu/led_theme/<theme>_<unit>.cfg`, installed from
`config/led_theme/<theme>.cfg` and pulled in by the hardware file with a
Klipper include whose path is relative to that file. Klipper's configparser
merges same-named sections per option, and the include sits where the effect
block used to live, so a theme's value wins any duplicated option.

- Only the selected theme is built and installed; when the selection
  changes, the install step prunes shipped theme files the unit no longer
  selects.
- `custom` is a machine-local file seeded exactly once, from the machine's
  default theme with the user's current values reapplied, installed
  no-clobber (like `mmu_vars.cfg`), and never written again: no build reads
  it back as an upgrade input, so the user owns it and upgrades and theme
  switches leave it untouched.
- A 4.00 refresh moves edited effect values from an installed hardware file
  into the new theme file (the builder re-applies live values regardless of
  which live file they came from) and settles byte-identical on the second
  pass.

For now the only stock theme (`mmu_leds`) is byte-identical to what the
hardware file used to carry, so no machine's effective config changes. The
include line is named from `PARAM_LED_THEME`, so new themes can be added
without touching the hardware template again.

## Commit 3 — the EMU theme

EMU units default to `emu_leds`, which matches the vendor's effect set: gate
availability shows dim white on the eject buttons (full white for the
selected gate), but loading/unloading speeds are swapped to match stock
instead of v3 upstream style (gate slow, extruder fast).
The vendor-specific `[mmu_led_effect ...]` definitions live in the theme file,
named per unit (a section name is global, so a multi-unit install would
otherwise define the same names twice). The `emu_leds` theme also
works on any other unit.

## Testing

- New tests cover the Makefile's per-unit theme selection (custom and
  selection-less values must still build the machine default, or the
  hardware include dangles), the custom seed and its leave-alone guarantee
  on rebuild, the install set, the parse exclusion, the EMU theme's
  Makefile cases, and the prune step.
- The 4.00 refresh fixtures model the effect migration, including user
  edits in the hardware file surviving into the theme file.
- EMU theme: multi-unit boot test with the vendor effect set and a clean
  boot of the EMU profile.
- Full suite green; the stock theme is identical to the block the
  hardware file used to carry.


burkfers:
@igiannakas Excited for your thoughts

moggieuk:
After review I'm going to merge these "high risk" changes into the `development` branch so we can experiment together. Then likely a 4.1 update.

burkfers:
Open review by EMU upstream promised for the coming days.
Marked ready in case you wish to override.

moggieuk:
Initial testing, getting a few errors:
1) unescaped # in the LED-theme helper caused Make to truncate the expression.
Fixed with this Makefile change:
-led_theme_of = $(strip $(subst ",,$(shell sed -n 's/^CONFIG_PARAM_LED_THEME=//p' $(1) 2>/dev/null | sed 's/#.*//')))
+# Escape the hash for make, which processes comments even inside shell quotes.
+led_theme_of = $(strip $(subst ",,$(shell sed -n 's/^CONFIG_PARAM_LED_THEME=//p' $(1) 2>/dev/null | sed 's/\#.*//')))

2) When running on multi-mmu setup:
no parsed Kconfig and no '/tmp/mmu_test/.mmu_config_unit1'; LED theme default assumed 'mmu_leds' (Could not open '/tmp/mmu_test/.mmu_config_unit1' (ENOENT: No such file or directory). Check that the $srctree environment variable (set to '/Users/paulmorgan/github/Happy-Hare3/installer') is set correctly.)

One thing that is strange is that I get both themes for a given unit id instead of just the one I selected.

Still playing..

burkfers:
> One thing that is strange is that I get both themes for a given unit id instead of just the one I selected.

I now render all themes at all times, to avoid theme deletion pitfalls when switching. You always get all available themes in the themes dir, one set per unit, and only the selected one is imported. The custom theme is special and is only first seeded when you select it.

I'll look at the issues you found asap.

moggieuk:
@burkfers I've added an alternative implementation as a single commit, 2d33a850e146038237716820cafcbf6ec205013f, directly on top of your original PR commit so it is easy to review or remove.

The selected theme is rendered into each unit's hardware config, avoiding separate installed theme files and simplifying the Makefile. Vendor effects can specify `unit:` so they are instantiated only for that unit; existing unscoped effects retain their all-unit behavior.

Refresh (including non-interactive updates) preserves the installed preset and edits. Merge preserves existing effect mappings and definitions while filling missing settings from the selected preset. Replace regenerates the preset defaults. Existing theme includes are migrated during Refresh/Merge.

Validation: all 1,898 tests across 67 files passed on this PR branch, including installer update-mode and multi-unit LED coverage. This is harness validation; it has not been tested on physical hardware.

To remove the entire alternative and restore the original approach without rewriting the PR history, run from a clean checkout of your fork:

```bash
git switch led_themes
git pull --ff-only
git revert 2d33a850e146038237716820cafcbf6ec205013f
git push
```

That reverts just this added commit and leaves your original PR commit intact.

moggieuk:
I realized more simplifications as I went to bed.  Will explain tomorrow..

moggieuk:
Pushed ee568a26: cleanup of dead code. The separate `mmu/led_theme/<theme>_<unit>.cfg` + `[include]` layout never shipped, so nothing installed can carry it. This removes the include-following migration in `prepare_led_theme`, the `led_theme` input filters in `build_config_file`, the `mmu/led_theme` glob in `hh_configs_to_parse`, and the two tests that only exercised that migration. No behavior change for the render-into-hardware design; `test_build`, `test_mmu_profiles`, `test_mmu_config` and `test_mmu_console` pass.

PR #1219, opened 2026-09-16
Fix calibrated bowden lengths resetting to -1 on restart

burkfers:
## Observed error

Calibrated bowden lengths reset to `-1` on every restart. After `MMU_CALIBRATE_BOWDEN`,
`mmu_vars.cfg` contains the correct values under `mmu_<unit>_bowden_lengths`, but after
a power cycle the same key holds `[-1, -1, ...]`. The value of `mmu_<unit>_bowden_home`
is `mmu_gate` and is stable across restarts (it is never rewritten). The machine has no
encoder fitted.

## Mechanism

Persistence of the per-unit bowden calibration rests on two variables: the length list
`mmu_<unit>_bowden_lengths` and the reference endstop `mmu_<unit>_bowden_home`, i.e. the
gate homing endstop the lengths were measured against. `MmuCalibrator._handle_ready`
loads both and accepts the lengths only if the home is a member of the current
`GATE_ENDSTOPS` set (`mmu_shared_exit`, `encoder`, `mmu_exit`, `extruder`); on rejection
it substitutes `[UNCALIBRATED] * num_gates`, and the unconditional write-back at the end
of the ready handler persists whatever it loaded.

Two defects interact:

1. **The reference endstop was never recorded.** `update_bowden_length()` persisted the
   lengths but not the home. The only writer of `VARS_MMU_BOWDEN_HOME` was
   `adjust_bowden_lengths_on_homing_change()`, which runs only after a *successful* load
   (it is gated on the `CALIBRATED_BOWDENS` flag). A stale home value therefore made the
   load fail, which made the repair path unreachable: the file was re-written with `-1`
   on every boot while the home stayed `mmu_gate`, a self-perpetuating loop that
   re-calibration could not break, because calibration does not touch the home.

2. **The stale value is unrepairable by the check.** `mmu_gate` is the v2/v3 name of the
   per-gate gate endstop; v4 renamed it to `mmu_exit` and removed it from
   `GATE_ENDSTOPS`. A v3 to v4 migration of `mmu_vars.cfg` that carried the value into the
   v4 namespaced key leaves a home that no v4 code path can write or accept, so the
   validity check rejects the load indefinitely.

A second, independent manifestation of defect 1: with `current_home` absent,
`adjust_bowden_lengths_on_homing_change()` treated the first boot after calibration as a
reference-endstop change and applied a spurious `±gate_endstop_to_encoder` adjustment.
With encoder homing, a bowden shorter than the offset becomes negative, and the next
boot's `x < 0 → UNCALIBRATED` normalization writes `-1` — the same observed symptom by a
different route.

## Changes

- `update_bowden_length()` records `VARS_MMU_BOWDEN_HOME = gate_homing_endstop` when a
  real calibration is saved (reset path unchanged). The baseline is now established at
  the moment calibration happens, which also covers the `MMU_TEST_CONFIG
  calibrated_bowden_length=` and autotune save paths.
- `adjust_bowden_lengths_on_homing_change()` treats a missing reference as "no
  baseline": it records the current endstop and returns without adjusting. A genuine
  endstop change after calibration still adjusts the lengths as before.
- `_handle_ready()` no longer rejects the whole load on an unrecognised stored home. It
  logs a warning naming the variable, the stored value, the valid set, and the assumed
  endstop, keeps the lengths, and lets the reference be re-stamped. Affected files heal
  on the next boot without manual editing.
- The negative → `UNCALIBRATED` coercion of stored lengths now logs a warning instead of
  happening silently.

## Tests

`test/test_mmu_bowden_persistence.py` (new): multi-boot tests where each "restart" is a
fresh `Session` pointed at the previous session's `mmu_vars.cfg`.

- stale v3-era home (`mmu_gate`) keeps the lengths and re-stamps the reference
- calibration persists `mmu_<unit>_bowden_home`
- restart with non-zero `gate_endstop_to_encoder` performs no adjustment
- encoder homing with a bowden shorter than the offset does not reset to `-1`
- guards: plain restart is a no-op; a live endstop change after calibration still
  adjusts (123.4 → 76.4 → 123.4 round trip with a 47.0 offset)

Verified against the pre-fix code: the four bug tests fail (lengths read back as `-1`,
`170.4`, `-17.0`→`-1`, or the home key absent); the two guards pass. Regression suites
(`test_mmu_save_variables`, `test_mmu_calibration_events`, `test_mmu_local_gate`,
`test_mmu_config`, `test_mmu_console`, `test_mmu_encoder`, `test_mmu_selector`,
`test_mmu_sensor_enable`) are green.

## On-printer verification

Efficacy of the fix verified by the Discord user experiencing the issue.


moggieuk:
P1: Stale-reference recovery still corrupts lengths. [Line 83](https://github.com/moggieuk/Happy-Hare/pull/1219/files#diff-placeholder) sets only the local bowden_home to None; adjust_bowden_lengths_on_homing_change() rereads the unchanged stored value. With mmu_gate and a 47 mm offset, I reproduced 123.4 → 170.4 mm. With encoder homing, 30 → −17 → −1 mm over two boots. Clear or normalize the stored reference before adjustment, and test stale references with a nonzero offset.

P2: The new negative-length warning includes normal UNCALIBRATED = -1 entries. This reports expected, partially calibrated configurations as bad data on every restart. Exclude the sentinel from the warning.

burkfers:
I've addressed both points.

Additionally, I've gated the stale-reference recovery on the unit actually having bowden lengths to protect: the "not a valid gate homing endstop" warning and the re-stamp now only fire when require_bowden_move is set and a length list was loaded, so a stale reference on a unit not performing bowden homing no longer produces a spurious warning on every boot.

PR #1220, opened 2026-09-17
Release the sync-feedback buffer spring at end of print

RT530:
A sprung sync-feedback buffer is left wherever the last extrusion put it, and on a machine that doesn't unload at the end of a print it stays that way until the next one starts — holding its spring loaded for however long the printer sits idle.

The print-end teardown already deactivates sync feedback, restores the rotation distance and unsyncs the gear, but it moves no filament: `on_print_end` calls `reset_sync_gear_to_extruder()` without `force_grip=True`, deliberately. So "sync feedback released" and "spring released" are not the same thing, and only the first one happens today.

### The change

`adjust_filament_tension()` gains a `target`, defaulting to `0.0`. Every existing caller keeps neutral and is unchanged; the arrival test just becomes `state - target` instead of `state`, and the initial move already carried the right sign.

`MMU_SYNC_FEEDBACK RELEASE=1` aims it at the buffer's own resting position, and `MMU_END` calls that on the branch where the tool is *not* unloaded. Where it is unloaded the filament leaves the buffer and the spring relaxes on its own, so there is nothing to do — and skipping it there also keeps `check_filament_in_mmu()`'s "buffer is not in resting position thus filament must be present" test meaningful during that unload.

### On the config key

The resting position comes from `buffer_spring_state`, which already exists and already describes exactly this. **That is a new use of it** — its documented purpose is "aids filament detection logic", and until now it was read-only, feeding `supports_active_probing()` and the detection test above. I'd rather say that plainly than have it look like the key was always meant as a motion target.

The upside is there's no new configuration and the feature is off by default: `none` is the shipped default and means no reliable resting position is known, so `RELEASE=1` is a no-op. It's also a no-op for switch buffers, which can only home to neutral — they decline rather than guess at a rail.

A target at a rail uses a one-sided arrival test. In the sensor's normal range this agrees with the symmetric band; it differs only for a reading past the far side of the target, which is what a drifted `analog_max_*` calibration produces. It is a robustness guard, not load-bearing.

### Testing

Five new tests in `test/test_mmu_sync_feedback.py`, all verified to fail without the change: the target and direction come from config rather than being hardcoded (tension retracts, compression feeds), both no-op cases, the rail arrival, and a regression test that `ADJUST_TENSION=1` still targets neutral. Full suite green.

Worth knowing for anyone extending these: the harness only models buffer travel dynamically for two-switch tension-sprung buffers, so a proportional buffer's analog reading doesn't respond to gear moves in tests. These script the readings and assert on the moves commanded.

### Verified on hardware

ERCF with a proportional buffer (`buffer_range: 6`, `buffer_maxrange: 12`), `buffer_spring_state: tension`, filament loaded, printer idle:

| Command | Move | Final state |
|---|---|---|
| `MMU_SYNC_FEEDBACK ADJUST_TENSION=1` | +7.20 mm | −0.108 |
| `MMU_END` | −5.95 mm | −1.000 (raw `0.9925` vs `analog_max_tension` `0.9925`) |

Run with a stop after each command so the buffer could be inspected by hand before the next was sent. At neutral the spring was confirmed compressed; after `MMU_END`, released. So the sensor's `-1.0` and the spring's actual rest position are the same point, which was the one assumption no amount of log reading could have settled.


config/macros/mmu_software.cfg
```diff
@@ -294,6 +294,11 @@ gcode:
         {% if unload or unload_tool %}
             MMU_LOG MSG="Unloading filament on print end"
             MMU_UNLOAD RESTORE=0
+        {% else %}
+            # Filament stays loaded, so a sprung buffer stays laden and holds its spring
+            # under load until the next print. Park it at its resting position instead.
+            # No-op unless 'buffer_spring_state' says where this buffer rests.
+            MMU_SYNC_FEEDBACK RELEASE=1
```
moggieuk: Don't forget that this can give an error if the MMU has no sync-feedback buffer. E.g.
```
MMU_SYNC_FEEDBACK RELEASE=1
No sync-feedback buffer on unit!
```

moggieuk:
@RT530, this PR has been merged into `development` after retargeting it from `main`.

Your original commit is preserved. I pushed maintainer follow-up commit 74219c6f1f688fb62dd16911ce50cf711e2428da to your `feat/sync-feedback-release` branch before merging. It adds:

- A guard before unit inference for release-only requests. An unknown gate on a multi-unit machine could otherwise raise an error and interrupt `MMU_END` before its cleanup. Disabled MMUs, bypass, and unloaded filament now skip release quietly.
- Macro conditions requiring a valid gate, loaded filament, and a sync-feedback buffer before calling `MMU_SYNC_FEEDBACK RELEASE=1`. Runtime checks also cover a print-end extension changing the state after the macro is rendered.
- Quiet skips for unsupported release conditions, regression tests, and shorter non-test comments. The switch-target comment now accurately says non-neutral targets are not implemented.

The PR's proportional-buffer release behavior and existing configuration are retained. No macro-vars option was added, and the original `Kconfig.sync_feedback_buffer` help text is unchanged.

Validation: all 1,456 tests passed on the updated PR branch; all 1,461 tests and lint passed on the combined result with `development`. The required GitHub `test` and `ruff` checks also passed before merging. No additional hardware testing was performed.

PR #1228, opened 2026-09-19
cut_tip: drop the exit safe-margin move

BuildLabDIY:
### Observation

`_CUT_TIP_DO_CUT_MOTION` ends with a move that backs the toolhead away by the
safe margin:

```jinja
{% if vars.gantry_servo_enabled|default(false)|lower != 'true' %}
   G1 X{pin_park_x_loc + safe_margin_x} F{travel_speed * 60} # back up safe margin
{% endif %}
```

`safe_margin_xy` is documented as an *approach* parameter - "Safety margin used
when approaching the cutter pin", and in `mmu_macro_vars.cfg`: "we make a safer
but longer move if we are closer to the pin than this specified margin. Usually
setting these to the size of the toolhead (plus a small margin) should be good
enough". It describes the toolhead footprint, so that coming at the pin from the
side does not clip it.

On the way out there is nothing to clear. The cut loop already ends with

```jinja
G1 X{pin_park_x_loc} Y{pin_park_y_loc} # Evacuate to start
```

so the toolhead is parked at `pin_loc + pin_park_dist` before this runs, and
`_MMU_POST_FORM_TIP` moves it on from there. The extra hop is a detour between
two moves that already leave the pin behind.

### Change

Drop the exit move, and the two `{% set %}` lines that become unused with it
(`safe_margin_x`/`safe_margin_y` and `travel_speed` in this macro only - the
approach block keeps its own copies).

### Testing

* Hardware, QIDI Box on a Q2: cut, park at `pin_park`, then straight to
  `post_form_tip_position`. Log shows the cut and the following park one second
  apart with nothing in between, across four toolchanges in a day.
* `python3 -m unittest discover -s test/installer` - failure list identical to
  `development` before the change (environment-only failures here: no jinja2 /
  menuconfig).
* Jinja block balance 24/24 before and after.

I have been running without this move for a while; the approach guard is
untouched. If it is deliberately there for a case I have missed - a taller
gantry, or `cut_iterations > 1` on some machine - say so and I will close this.


moggieuk:
To be honest, that wasn't an addition I made to the macro.  I understand the approach safety so that the pin is not grazed by the toolhead.   Can you help me understand why this same precaution isn't required on exit?  I.e. couldn't the toolhead return to position also cause the toolhead to clip the pin?

This isn't a feature I use on my printers so I'm trying to imagine in my head whether it is safe to remove..

BuildLabDIY:
Good question - I went and ran both paths on the machine to answer it properly.

The asymmetry is not approach-vs-exit, it is **two-axis vs one-axis**.

With `pin_loc_xy: 15, 0.5`, `pin_park_dist: 5`, `safe_margin_xy: 30, 30`, starting from a print position at X40 Y100, the approach expands to:

```gcode
G1 X50 Y30.5 F9000   ; 1st safety move: pull away in Y first
G1 X50 Y0.5  F9000   ; 2nd safety move: drop onto the cut line, still 30mm clear in X
G1 X20 Y0.5  F9000   ; final: approach the pin along the cut line
G1 X-0.5 Y0.5 F480   ; the cut
G4 P50
G1 X20 Y0.5  F9000   ; evacuate to start (inside the loop)
```

That guard matters because a direct diagonal from X40 Y100 would bring the toolhead onto the pin *sideways*. It avoids that by leaving along Y first, then coming in along the cut line. On my machine the depressor is a lever on the toolhead rather than a fixed post, so approaching off the cut line would load it from the wrong direction - the guard is doing real work here.

The move this PR removes is this one:

```gcode
G1 X45 F9000         ; back up safe margin - X only, no Y
```

A single X hop from X20 to X45. It has no Y component, so it is not the manoeuvre above. If the move that follows is unsafe, this hop does not make it safe - `RESTORE_GCODE_STATE ... MOVE=1` still travels in a straight line to its target, just starting 25mm further out.

I ran both sequences back to back on the printer, identical except for that one line. Approach, cut and evacuate are the same in both; the only visible difference is the toolhead darting out to X45 and then turning around, before heading where it was going anyway.

So: **if the exit does need protection, it needs the mirror of the approach** -

```gcode
G1 X50 Y30.5 F9000   ; leave along Y before travelling anywhere
```

- not a lone X backup. Dropping the X hop does not drop a guard, because it was not acting as one.

One thing I should be upfront about: I run `restore_position: False`, so after the cut `_MMU_POST_FORM_TIP` takes the toolhead to my bin position - away from the pin in both X and Y. I have **not** tested the wipetower path with `restore_position: True`, which is the case you are asking about.

If you would rather keep a guard there, I am happy to close this and open one that makes the exit a proper mirror of the approach instead. That covers the wipetower case properly and still drops the detour for parking-logic setups.

---

On a separate note - while getting the readers working I noticed Happy Hare can read tags but not write them. Right now binding a spool to a tag means either Spoolman holding the UID->spool mapping, or writing the tag from a phone or a dedicated writer first.

The rc522 does support MIFARE Write, so filament data could be written to the tag from the printer itself during normal operation - no second device involved. Before I spend time on it: is that something you would want in Happy Hare, and do you think it is useful beyond my box? Happy to drop the idea if it is out of scope.

BuildLabDIY:
One more option, thinking about it further: since this touches a shared macro, the safest form might be to make it opt-in rather than change behaviour for everyone.

Something like `BOOL_CUT_TIP_SKIP_EXIT_MARGIN` in `Kconfig.cut_tip`, defaulting to `n`, with the exit move wrapped in it - and `select`ed only by profiles that park elsewhere after the cut. Nothing changes for existing setups, including the wipetower case you are asking about, and parking-logic setups can drop the detour.

Let me know which you prefer - removing it outright, making it a mirror of the approach, or the opt-in flag - and I will rework this PR accordingly.

ningpj:
The inboard move is to provide clearance when the depressor is close to blobifier so it doesnt clip.

BuildLabDIY:
Thanks - that is exactly the piece I was missing. I could not work out what the move was clearing, because on my machine there is nothing next to the cutter.

So it is not redundant, it is conditional: it matters when a Blobifier sits close to the depressor, and does nothing otherwise. Removing it outright was wrong.

Rather than close this, would you take it as a guarded move? `blobifier.cfg:839` already tests for the module at runtime, with the same intent:

```jinja
{% if printer["gcode_macro _BLOBIFIER_VARS"] is defined %} # bypass if configuration isnt defined
```

Wrapping the exit move in the same test keeps the clearance for every Blobifier setup and drops the move for everyone else. No new config option and no template changes - and since `mmu_cut_tip.cfg` is not templated by the installer, a runtime test is the only way to make it conditional.

I checked the negative side on my machine: `_BLOBIFIER_VARS` is not defined here, even though the `BLOBIFIER*` macros themselves are installed - the vars block is only emitted under `[% if MMU_HAS_BLOBIFIER %]`, so the test discriminates correctly rather than just detecting the file. I cannot verify the positive side, as I have no Blobifier to test against, but it comes from the same flag that enables the rest of it.

Happy to push that to this PR if the approach looks right to you.

moggieuk:
So rather than making the gocde dependent on BLOBIFIER (which creates additional dependencies) why not just change the default set by Kconfig to be dependent on blobifier.

I.e. something like this (untested):
```
  config VAR_CUT_TIP_SAFE_MARGIN_XY
    string "Safe margin (X,Y)            "
    default "" if !MMU_HAS_BLOBIFIER
    default "30, 30"
    help
      Safety margin used when approaching the cutter pin.
      Primarily used to avoid collision with Blobifier
```
@ningpj  would love your opinion too.

BuildLabDIY:
> So rather than making the gocde dependent on BLOBIFIER (which creates additional dependencies) why not just change the default set by Kconfig to be dependent on blobifier.
>
> ```
>   config VAR_CUT_TIP_SAFE_MARGIN_XY
>     string "Safe margin (X,Y)            "
>     default "" if !MMU_HAS_BLOBIFIER
>     default "30, 30"
> ```

I tried it on my machine before replying, since you marked it untested.

An empty default does not survive config parsing - Klipper reads `variable_*` as a Python literal, so it fails before any Jinja runs:

```
Option 'variable_safe_margin_xy' in section 'gcode_macro _MMU_CUT_TIP_VARS'
is not a valid literal: invalid syntax (<unknown>, line 0)
```

`"0, 0"` parses fine, but it changes the wrong thing. `safe_margin_xy` is read twice - line 204 for the approach and line 255 for the exit - and the approach uses it as the *trigger* for the two-step manoeuvre:

```jinja
{% if printer.gcode_move.gcode_position.x < pin_park_x_loc + safe_margin_x %}
```

With the margin at zero that test is effectively never true, so the approach guard goes away as well. On my toolhead the depressor is a lever rather than a fixed post, and coming at it off the cut line loads it sideways - so that guard matters here even though the Blobifier clearance does not. And the exit still emits `G1 X20`: a move that goes nowhere, rather than no move at all.

So the two uses need separating either way. The runtime test only touches the exit line and leaves the approach alone, which is why I reached for it - but if you would rather keep it in Kconfig, a separate `VAR_CUT_TIP_EXIT_MARGIN_X` defaulting to `0` when there is no Blobifier does the same job without a cross-dependency in the macro. Either works for me.

moggieuk:
Ah, my Kconfig change is sound but the macro doesn't expect `variable_safe_margin_xy` to be empty. "0, 0" might be better but looks like the macro would also need a slight tweak.  A simple if "safe_margin_x != 0" would be sufficient.

I'm not looking at the macro code right now, but I would have though 0,0 should work fine and mean "no safe clearance needed" ?

moggieuk:
For now I've created PR1244 to document the parameter better and going to close this for now.  I think the symmetry of the operation should be retained.   Happy to consider additional options but I want to consider existing setups as well...

PR #1229, opened 2026-09-19
Allow spool rotation when the gear motor turns the spool

BuildLabDIY:
### Observation

Spool rotation during drying is gated on having an eSpooler, with VVD special
cased by vendor name:

```python
if rotate and not (mmu_unit.has_espooler() or mmu_unit.mmu_vendor == VENDOR_VVD):
    self.mmu.log_warning("Rotation requested but eSpooler not fitted - ignoring")
```

and the same comparison again in `mmu_environment_manager.py`.

On a type-B unit without an eSpooler the gate's own gear motor turns the spool.
That holds for VVD and for the QIDI Box alike, so the test is really about the
design rather than about the vendor.

Related: `heater_rotate_interval` is emitted only under `MMU_HAS_ESPOOLER`, so
on these units the key never reaches `mmu_parameters.cfg` at all - even though
the firmware reads it and defaults to 5.

### Change

Add a `gear_rotates_spool` attribute to the unit profile, alongside
`filament_always_gripped` and the others, overridable from config the same way.
Set it for VVD and QIDI and use it in place of the two vendor comparisons.
Mirror it in the installer as `MMU_GEAR_ROTATES_SPOOL` so the interval key is
emitted for these designs too.

### Testing

* Hardware, QIDI Box on a Q2: before, `MMU_HEATER DRY=1 ROTATE=1 GATES=0` logged
  "Rotation requested but eSpooler not fitted - ignoring"; after, the drying
  cycle rotates all four spools.
* `python3 -m unittest discover -s test/installer` - failure list identical to
  `development` before the change.

I went with a profile attribute rather than a third vendor comparison, since
that list would otherwise grow with every type-B design that drives its spools
this way. Happy to rename it or move it elsewhere if you prefer.


moggieuk:
Nice.  This is the correct thing to do.  I special cased VVD because it was the first and only MMU that worked like this.

Now I think about it, the EMU design is also similar with a powered spool winder..

BuildLabDIY:
Thanks - glad it fits.

On EMU: I checked the profile and it does not declare `MMU_HAS_ESPOOLER`, so rotation is currently blocked there the same way it was for me. If its winder is driven by the gate motor rather than a separate eSpooler, then adding `select MMU_GEAR_ROTATES_SPOOL` to `Kconfig.emu` is all it needs - the same one-liner QIDI and VVD got.

I do not have an EMU to verify that, so I have not touched its profile. Happy to send a follow-up if you confirm that is how it drives the spool.

PR #1235, opened 2026-09-20
Introduce stepper driver selection on the stepper blobifier (2209, 2240, 5160)

igiannakas:
For completeness sake and to avoid manual configuration, introducing 2240, 5160 as options in the stepper blobifier.

PR raised against development branch to allow testing before upstreaming to main.


installer/Kconfig.purging
```diff
@@ -121,12 +121,61 @@ menu "Purging"
...
+
+      # SPI drivers are addressed by cs_pin, UART drivers by uart_pin
+      config BOOL_BLOBIFIER_TMC_SPI
+        bool
+        default y if CHOICE_BLOBIFIER_TMC5160 || CHOICE_BLOBIFIER_TMC2240
+
       config PIN_BLOBIFIER_STEPPER_UART
-        prompt "Stepper TMC UART pin       "
+        string
+        prompt "Stepper TMC uart pin       " if PARAM_BLOBIFIER_TMC != "" && !BOOL_BLOBIFIER_TMC_SPI
```
burkfers: lower case "uart" is inconsistent with other mentions

installer/Kconfig.warnings
```diff
@@ -164,6 +169,7 @@ if SHOW_PER_UNIT_WARNINGS || SHOW_SHARED_WARNINGS
     comment "Toolhead sensor is enabled but no MCU pin defined"                                if SW2
     comment "Extruder sensor is enabled but no MCU pin defined"                                if SW3
     comment "Blobifier bucket switch is enabled but no valid MCU pin is defined"               if SW4
+    comment "Blobifier tray stepper TMC is selected but no uart/cs pin is defined"             if SW5
```
burkfers: UART / CS should be upper case here as well

config/base/mmu.cfg
```diff
@@ -658,37 +656,28 @@ homing_speed             : 20		# Unused in the macros, here for completeness
 velocity                 : 150
 accel                    : 1000
 
-[tmc2209 manual_stepper stepper_blobifier]
+[% if PARAM_BLOBIFIER_TMC != "" %]
+[[ "[" ~ PARAM_BLOBIFIER_TMC ~ " manual_stepper stepper_blobifier]" ]]
+[% if BOOL_BLOBIFIER_TMC_SPI %]
+cs_pin                   : [[PIN_BLOBIFIER_STEPPER_CS]]
+[% if PIN_BLOBIFIER_STEPPER_SPI_SCLK != "" %]
```
burkfers:
```suggestion
[% if (PIN_BLOBIFIER_STEPPER_SPI_SCLK != ""
  and PIN_BLOBIFIER_STEPPER_SPI_MOSI != ""
  and PIN_BLOBIFIER_STEPPER_SPI_MISO != "") %]
```
Guarding only on SCLK means a filling out SCLK and leaving both or either of MOSI/MISO blank leads to an invalid unparseable config being generated.

igiannakas:
Review comments implemented

burkfers (approved):

moggieuk:
I pushed an update that did a few things:
1) Retained the separation of config token typing and the actual prompt.  Safer because the type is always defined with default in case tokens are not used perfectly in a suitable `if` block in the jinja template.
2) Clarified TMC2240 is SPI-only.
3) Expanded regression coverage.
4) Fixed silent fallback from incomplete software-SPI pins.
5) Added hardware SPI bus selection.

I'll pause on merging until you have validated I haven't broken anything you did -- I don't think but worth another check.

igiannakas:
I’ll pull into my dev and test today. Should be ok I think

burkfers:
> Clarified TMC2240 is SPI-only.

This is not actually true. It's a tiny micro soldering job on most step sticks, but there is a jumper to change to UART.

moggieuk:
I wondered about TMC2240 ... I completed the code in place rather than verify.  I wonder if there should be separate TMC2240 / uart or how complete this needs to be for the blobifer user case

burkfers:
I rate it very low priority if the current shape can't cover TMC2240 over uart. Stepper Blobifier is a niche within a niche, 2240 is a niche within that, and 2240s modified to speak UART will be a niche within that.

PR #1240, opened 2026-09-21
Board authoritative i2c defaults

burkfers:
Proof of concept for board-authoritative i2c choices and defaults

The feature files now know nothing about the offered I2C busses and chipdefs, they just define an "I2C bus for this features goes here" and a custom escape hatch with free text field for the chipdef. Any offered bus choice is inserted by the board Kconfig.


moggieuk:
This approach is perfect and makes it a point and click rather than editing a text field.    By that I mean you COULD just set the default PARAM string default conditionally instead of creating the choice+custom setup.   I'm not complaining just saying there would be a simpler way to code which might be useful if the number of choices is/becomes too large.  If there are only a few set choices then your approach is preferable...

burkfers:
> If there are only a few set choices then your approach is preferable...

Currently we only have two sets of choices regarding i2c busses (env sensor and NFC reader). More applications are worth a think but out of scope here.

Each board file usually only adds one choice - The majority of boards only expose one i2c bus. So the extent in each board file will be a couple lines.

moggieuk:
All sounds good and complete then...

